PR #2, opened 2015-01-01
Fix compatibility issues cause by libmproxy update from 0.10 to 0.11

vtoubiana:
This pull request only fixes the update of libmproxy.


englehardt:
Changed indent to 4 spaces, removed some comments/print statements and extra code, slightly refactored. I will do a bit more testing before pulling into master but you can find the changes here: bbcae2761582ee8a07993b7d22c77cbc5919cf1e

Thanks for the contribution!

PR #39, opened 2015-12-07
Added file to sequence commands in one batch for one top url site visit

englehardt:
Added file to sequence commands in one batch for one top url site visit

Added support for command sequence in task manager

Replacing top_url with visit id for http_request, http_response, flash_cookies and profile_cookies tables

Changed task manager to execute a command sequence

Task manager updated to rectify Reading single entry from the table

Changed iteration for commands of command sequence

Add visit id in the command sequence instead of browser manager

Added visit id to proxy, command executor and browser manager

Added file to sequence commands in one batch for one top url site visit

Added support for command sequence in task manager

Replacing top_url with visit id for http_request, http_response, flash_cookies and profile_cookies tables

Changed task manager to execute a command sequence

Task manager updated to rectify Reading single entry from the table

Add visit id in the command sequence instead of browser manager

Added visit id to proxy, command executor and browser manager

Added visit id in extension

Changes task manager to close after completing the command sequence

Fixed small changes


englehardt:
Updated in #74

PR #74, opened 2016-05-04
Adding up-to-date CommandSequences and adding tests

dreisman:
Adds CommandSequence.py to automation/, providing a new API for creating complex command sequences

Ex:
Instead of manager.get(url)...
cs = CommandSequence(url)
cs.get(arg1, arg2...)
...set up other things to do in same page visit...
manager.execute_command_sequence(cs)

manager.get() and manager.browse() now wrap command sequences if you still want a simpler interface.

-Data in DB now keyed by 'visit_id,' which maps to a single executed CommandSequence on a particular site_url.

-Added tests


automation/CommandSequence.py
```diff
@@ -0,0 +1,58 @@
+from Errors import CommandExecutionError
+
+class CommandSequence:
+    """A command sequence should be associated with one top url site visit"""
```
englehardt: This comment should describe the command sequence class. How it's useful and perhaps include sample code on how to use it. I think this is particularly important as this is a big change in the infrastructure's interface and the class is user-facing.

automation/CommandSequence.py
```diff
@@ -0,0 +1,58 @@
...
+        """ dumps from the profile path to a given file (absolute path) """
+        if not self.contains_get_or_browse:
+            raise CommandExecutionError("No get or browse request preceding "
+                                        "the dump storage vectors command", self)
+        command = ('DUMP_PROFILE_COOKIES',)
+        self.commands_with_timeout.append((command, timeout))
+
+    def dump_profile(self, dump_folder, close_webdriver=False, compress=True, timeout=120):
+        self.total_timeout += timeout
+        """ dumps from the profile path to a given file (absolute path) """
```
englehardt: Move docstrings directly below function declaration.

automation/Commands/browser_commands.py
```diff
@@ -141,16 +141,13 @@ def extract_links(webdriver, browser_params, manager_params):
 
     sock.close()
 
-def browse_website(url, num_links, sleep, webdriver, proxy_queue,
+def browse_website(url, num_links, sleep, visit_id, webdriver, proxy_queue,
                    browser_params, manager_params, extension_socket):
     """
     calls get_website before visiting <num_links> present on the page
-    NOTE: top_url will NOT be properly labeled for requests to subpages
```
englehardt: This comment should not have been removed, as it's still true. The wording needs to be updated to show explain that the top_url in the site visits table won't match.

automation/TaskManager.py
```diff
@@ -111,6 +111,23 @@ def __init__(self, manager_params, browser_params, process_watchdog=False):
...
+                self.db.commit()
+                last_visit_id = cur.fetchone()[0]
+                if last_visit_id is None:
+                    last_visit_id = 0
+                query_successful = True
+            except OperationalError:
+                time.sleep(0.1)
+                pass
+        self.next_visit_id = last_visit_id + 1
+
```
englehardt: The only reason to wrap in the `while not query successful` loop is to avoid `OperationalError`s that arise from multiple writers attempting to write to the database at the same time. That error shouldn't be thrown on a read. Also, there is no need to commit after a SELECT.

Can be reduced to:
```
# read the last used site visit id
cur = self.db.cursor()
cur.execute("SELECT MAX(visit_id) from site_visits")
last_visit_id = cur.fetchone()[0]
if last_visit_id is None:
    last_visit_id = 0
self.next_visit_id = last_visit_id + 1
```

test/test_simple_commands.py
```diff
@@ -0,0 +1,173 @@
...
+    def get_config(self, data_dir):
+        manager_params, browser_params = TaskManager.load_default_params(self.NUM_BROWSERS)
+        manager_params['data_directory'] = data_dir
+        manager_params['log_directory'] = data_dir
+        manager_params['db'] = os.path.join(manager_params['data_directory'],
+                                            manager_params['database_name'])
+        browser_params[0]['headless'] = True
+        return manager_params, browser_params
+
+    def test_get_siteVisitsTableCorrect(self, tmpdir):
```
englehardt: Please change functions names to all lowercase separated by underscores.

automation/schema.sql
```diff
@@ -3,6 +3,10 @@
  * IF NOT EXISTS, otherwise there will be errors
  */
 
+
+/* SCHEMA VERSION */
+/* PRAGMA schema.user_version = 2; */
```
englehardt: Is this not working?

test/test_simple_commands.py
```diff
@@ -0,0 +1,173 @@
...
+        assert qry_res[0][0] == visit_ids[url_b]
+
+        # Page simple_a.html has a link to simple_c.html. This request should
+        # be keyed to the site visit for simple_a.html
+        qry_res = utilities.query_db(manager_params['db'],
+                                     "SELECT visit_id FROM http_responses"
+                                     " WHERE url = ?", (url_c,))
+        assert len(qry_res) == 1
+        assert qry_res[0][0] == visit_ids[url_a]
+
```
englehardt: Could you add a test to check that visit_id properly propagates to the extension? That seems to be untested.

automation/Extension/firefox/data/create_content_policy_table.sql
```diff
@@ -6,5 +6,5 @@ CREATE TABLE IF NOT EXISTS content_policy(
 	request_origin TEXT,
 	mime_type_guess TEXT,
 	page_id INTEGER,
-        top_url TEXT
+        visit_id INTEGER,
```
englehardt: Let's fix the indentation on the extension sql files while we're making changes here.

automation/schema.sql
dreisman: It wasn't, I couldn't figure out why but I didn't dig very hard. I'll remove it for now and take another stab before we want to finalize this

automation/CommandSequence.py
dreisman: Done.

automation/CommandSequence.py
dreisman: Done.

automation/Commands/browser_commands.py
dreisman: Done.

automation/TaskManager.py
dreisman: Done.

test/test_simple_commands.py
dreisman: Done.

test/test_simple_commands.py
dreisman: Done.

automation/Extension/firefox/data/create_content_policy_table.sql
dreisman: Done.

englehardt:
Great work to both you and Shivam! This feature has been on the TODO for quite some time, some I'm happy to see it completed.

PR #90, opened 2016-09-12
Detecting AudioContext 

albertterradas:
Code to be available to detect audio fingerprinting.

Advice: The openwpm/automation/Extension/firefox/@openwpm-0.0.1.xpi compressed directory needs to be regenerated.

### DIFF
--- a/automation/Extension/firefox/data/content.js
+++ b/automation/Extension/firefox/data/content.js
@@ -440,6 +440,15 @@ function getPageScript() {
 
     // Access to webRTC
     instrumentObject(window.RTCPeerConnection.prototype,"RTCPeerConnection", true);
+    
+    //Access to AudioContent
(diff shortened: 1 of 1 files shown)


englehardt:
Thanks a bunch for the pull request! I already had Audio API instrumentation and tests written in another repo. I've just merged into the public repo in 4a9f18a86bd848665f61cbe5805de407f5984f54, so I'm going to close this.

I decided not to add `window.webkitOfflineAudioContext` instrumentation as Firefox and Chrome don't support this prefix and we'd only see the initial access with OpenWPM (since the call will return `undefined`). Does that make sense to you?

albertterradas:
It makes perfect sense to me! (I thought I've previously removed it from the script) Thank you for adding it!

PR #104, opened 2016-11-14
Add support for custom, inline commands

englehardt:
The `CommandSequence::run_custom_function` method allows a user to specify browser commands "inline", within their crawl script. It makes it possible to write tests which depend on user/webdriver interaction and is much quicker than adding a new CommandSequence command for simple, one-off tasks.


englehardt:
Documentation of this has been added here: https://github.com/citp/OpenWPM/wiki/Available-Commands#run_custom_function

PR #105, opened 2016-11-16
Remove webdriver HTML attribute from documentElement and navigator

### LINKED ISSUE #91: Block webdriver html attribute from getting set
Selenium adds an HTML attribute "webdriver = true" to the <html> tag. This should be removed before any scripts on the page are executed.

(also, measuring how many sites/scripts check for this attribute would be informative)

gunesacar:
Fixes #91.

The fix assumes that OpenWPM's `content.js` runs _after_  Selenium code linked below.

- https://github.com/SeleniumHQ/selenium/blob/b82512999938d41f6765ce8017284dcabe437d4c/javascript/firefox-driver/extension/content/server.js#L49
- https://github.com/SeleniumHQ/selenium/blob/b82512999938d41f6765ce8017284dcabe437d4c/javascript/firefox-driver/extension/content/dommessenger.js#L98

Manually tested on https://output.jsbin.com/webawamabu.

The console logs should read:
```
- Successfully started all instrumentation
- navigator.webdriver:  undefined
- webdriver in navigator:  false
- documentElement attribute webdriver:  null

```

### DIFF
--- a/automation/Extension/firefox/data/content.js
+++ b/automation/Extension/firefox/data/content.js
@@ -343,6 +343,13 @@ function getPageScript() {
       };
     }
 
+    function removeWebdriverAttributes(){
+      if ("webdriver" in navigator){
+        delete window.navigator["webdriver"];
+      }
+      document.documentElement.removeAttribute("webdriver");
+    }
+
     // Log properties of prototypes and objects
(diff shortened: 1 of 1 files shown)


englehardt:
Looks great, thanks Gunes!

PR #106, opened 2016-11-17
Revert "Remove webdriver HTML attribute from documentElement and navigator"

englehardt:
Reverts citp/OpenWPM#105

### DIFF
--- a/automation/Extension/firefox/data/content.js
+++ b/automation/Extension/firefox/data/content.js
@@ -343,13 +343,6 @@ function getPageScript() {
       };
     }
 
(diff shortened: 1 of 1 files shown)


englehardt:
I've decided to revert this merge for a couple reasons:

(1) We should really have an automated test for this. If it does break in the future for some reason it's unlikely to be noticed by an user of the platform

(2) In order to activate, the user must set `browser_params['extension']['enabled'] = True` and `browser_params['extension']['jsInstrument'] = True`. We shouldn't tie the protection to the use of instrumentation.

PR #108, opened 2016-11-23
Remove Selenium `webdriver` attributes from DOM

englehardt:
See commit comments for more information.

@gunesacar Let me know what you think! Thanks for the work on #105.


test/test_extension.py
```diff
@@ -20,6 +20,36 @@ def get_config(self, data_dir):
...
+        * https://github.com/SeleniumHQ/selenium/blob/b82512999938d41f6765ce8017284dcabe437d4c/javascript/firefox-driver/extension/content/dommessenger.js#L98
+        """
+
+        from ..automation import CommandSequence
+        def check_for_webdriver(**kwargs):
+            """ Check if webdriver self-identification attributes in the DOM"""
+            driver = kwargs['driver']
+
+            # Check if document element has `webdriver` attribute
+            assert 'true' != driver.execute_script(
```
gunesacar: @englehardt these asserts fail when you have `disable_webdriver_self_id=false`, right?
Just wanted to make sure since they are called within a function that is passed around.

automation/Extension/firefox/data/remove_webdriver_attributes.js
```diff
@@ -0,0 +1,54 @@
...
+        value: function(obj, prop, descriptor) {
+          if (obj == window.navigator && prop == 'webdriver') {
+            console.log("Preventing definition of webdriver property on navigator.");
+
+            // Return Object.defineProperty to original state
+            Object.defineProperty(Object, 'defineProperty', {
+              value: originalDefineProperty
+            });
+            delete originalDefineProperty;
+            return;
```
gunesacar: Could we use `return undefined;` instead of `return;`?
The former would be more readable.

automation/Extension/firefox/data/remove_webdriver_attributes.js
```diff
@@ -0,0 +1,54 @@
...
+      Object.defineProperty(Object, 'defineProperty', {
+        value: function(obj, prop, descriptor) {
+          if (obj == window.navigator && prop == 'webdriver') {
+            console.log("Preventing definition of webdriver property on navigator.");
+
+            // Return Object.defineProperty to original state
+            Object.defineProperty(Object, 'defineProperty', {
+              value: originalDefineProperty
+            });
+            delete originalDefineProperty;
```
gunesacar: I couldn't understand why we have to delete `originalDefineProperty`. A comment would be great.

Also, what happens when `Object.defineProperty` is called more than once with the arguments `navigator` and `"webdriver"`? Wouldn't it set `Object.defineProperty` to `undefined` since `originalDefineProperty` is deleted in the first call?

gunesacar:
This looks way better than my naive approach, it has the resilience we need to cope with various race conditions.

My high-level concern is the overwriting of `object.defineProperty`. This function is so fundamental that, there's no room for even a tiny glitch.

Please see my other, more specific comments next to the code.

automation/Extension/firefox/data/remove_webdriver_attributes.js
englehardt: My intention was to prevent that variable from staying around the global namespace. However I tested it and I think that's handled anyway with the way we inject scripts. I've removed the delete statement.

englehardt:
Thanks for the detailed review. Just pushed a couple commits to address your comments.

I tried to avoid messing with `Object.defineProperty` but my other approaches didn't work:

1. I tired using `Object.watch()`. It will detect property assignments but won't detect property creation with `Object.defineProperty` so it's not an option.
2. I tried to set `window.navigator = new Proxy(window.navigator, {...})` and proxy property creation there, but it doesn't seem to be possible to overwrite `window.navigator`.
3. I tried creating my own `webdriver` attribute on the `navigator` with a custom `get` and `set` property that would destroy the attribute the first time it changes, but subsequent `Object.defineProperty` calls don't make use of the current getters and setters.

Extending from (3) I think it might be possible to define our own property and have some special handling on deletion that prevents the subsequent re-creation of the property...but my current approach seems better.

gunesacar:
Thanks for addressing all the comments, @englehardt.
Looks good to me.

RonnieBlade:
Hi! Is it possible to remove `webdriver` attributes from Firefox controlled by Selenium using C#?

PR #110, opened 2016-12-02
Store function name and eval/Function details for the JS calls (#49).

### LINKED ISSUE #49: Some javascript URLs have "line XX > function" after them
Will need to update [this regex](https://github.com/citp/OpenWPM/blob/master/automation/Extension/firefox/data/content.js#L104) to fix it. Should be able to check for eval or function.

gunesacar:
Remove regex based call script URL parsing.
Remove redundant array access and parenthesis.
Add test for the call stack parsing.

Fixes #49.


englehardt:
This looks great, thanks Gunes.

The stack parsing is much clearer and I'm glad we now have tests for this.

PR #120, opened 2016-12-13
Fixed bug in TaskManager.browse() wrapper, where it was accidentally …

dreisman:
…calling get()


englehardt:
Thanks Dillon! Do you mind to also add a test for this? We already have browse command tests, just not for this shortcut. It should be easy to hook into those.

PR #121, opened 2016-12-14
Store timestamp for javascript instrumentation.

gunesacar:
By timestamping Javascript calls, we could correlate network events to Javascript execution.


test/test_extension.py
```diff
@@ -114,3 +115,13 @@ def test_js_call_stack(self, tmpdir):
...
+    def test_js_time_stamp(self, tmpdir):
+        # Check that timestamp is recorded correctly for the javascript table
+        MAX_TIMEDELTA = 30  # max time diff in seconds
+        db = self.visit('/simple_a.html', str(tmpdir))
+        utc_now = datetime.utcnow()  # OpenWPM stores timestamp in UTC time
+        rows = utilities.get_javascript_entries(db, all_columns=True)
+        for row in rows:
+            js_time = datetime.strptime(row[14], "%Y-%m-%dT%H:%M:%S.%fZ")
+            # compare UTC now and the timestamp recorded at the visit
+            assert (utc_now - js_time).seconds < MAX_TIMEDELTA
```
englehardt: This test will pass if the command fails (since rows will have length 0). Would you mind to add the following after the for loop:

`assert not utilities.any_command_failed(db)`

See utilities function here: https://github.com/citp/OpenWPM/blob/fe066152d63642421ebf1ad85c984b9f17b28997/test/utilities.py#L98-L104

gunesacar: Thanks for pointing out, should be addresses in the updated PR.

gunesacar:
I used a different test page (canvas) in the updated PR.
This is to extend the test coverage by testing both logCall and logValue instrumentation, which are timestamped separately.

englehardt:
> I used a different test page (canvas) in the updated PR.
This is to extend the test coverage by testing both logCall and logValue instrumentation, which are timestamped separately.

Good catch! Looks good to me.

PR #122, opened 2016-12-15
Don't use system Firefox binary in the manual tests, improve terminal log colors

gunesacar:
Switch to test directory before starting the web server.

Log table name for the database queries to terminal.
Update terminal colors for the manual_test to be more readable, will look like this:

terminal

Currently it looks like this:
terminal_current


englehardt:
Thanks, this is much better. Glad to have this in the public repo. Would you also mind to re-generate the xpi since we don't do that automatically here?

gunesacar:
> Would you also mind to re-generate the xpi since we don't do that automatically here?

Sure, done.

PR #124, opened 2016-12-18
Minor test refactoring: DRY get_config, improve runtime environment tests

gunesacar:
Move boilerplate `get_config` code to the parent test class
This is to remove the frequently repeating code from the tests.

Read the python dependencies to be tested from requirements.txt

Add tests to make sure firefox-bin directory and the Firefox
binary within exist.

Fix some PEP8 warnings in the test code.

Remove unused imports.


test/openwpmtest.py
```diff
@@ -16,6 +19,20 @@ def visit(self, page_url, data_dir, sleep_after=0):
         manager.close()
         return manager_params['db']
 
+    def get_test_config(self, data_dir="",
+                        num_browsers=NUM_BROWSERS):
+        """Load and return the default test parameters."""
+        if not data_dir:
+            data_dir = mkdtemp()
```
englehardt: One benefit of using pytest's tmdir is the predictable location (i.e. `/tmp/pytest-of-<user>/pytest-<user>/`, which always points to the latest test's tmp folder) and the versioned directories for the last `n` runs. I've relied on this pattern when tweaking tests to inspect log files and databases of the recent tests.

Would you mind to include the generated directory in stdout to make it easy to navigate to?

PR #125, opened 2016-12-21
Store POST request data as JSON in the database.

gunesacar:
Incorporated some code from: https://github.com/redline13/selenium-jmeter

Handle various content encodings for the POST data (encType).
This includes `text/plain`, `application/x-www-form-urlencoded` and
`multipart/form-data`.

Store "request headers from upload stream" in the database.
We add additional headers from POST body to existin headers column.

Discard OCSP (POST) requests to prevent noise.

Add tests for form submissions with different encodings by JavaScript.
Add tests for making POST requests with AJAX (jQuery) and FormData.

Handle exception when looking up the `Content-Type` response header.

Add shared/utils.js file. We can add common test functions to this file.


automation/Extension/firefox/lib/http-post-parser.js
```diff
@@ -0,0 +1,250 @@
...
+
+
+HttpPostParser.prototype.parseSinglePart = function(part) {
+  /*
+   * Parse a single part of a multipart request body
+   * e.g., one part is as follows:
+     Content-Disposition: form-data; name="username"
+
+     name surname+
+  */
```
englehardt: According to: https://stackoverflow.com/questions/16548517/what-is-request-headers-from-upload-stream%60 there may be additional headers within parts, such as `Content-Type: `. From what I can tell your code doesn't handle this?

automation/Extension/firefox/lib/http-post-parser.js
```diff
@@ -0,0 +1,250 @@
...
+      post_vars.push(post_vars.pop() + x);
+    }
+  }
+  return post_vars.join("&");
+}
+
+HttpPostParser.prototype.parseEncodedFormData = function(formData, encodingType){
+  var obj = {};
+
+  if (formData.indexOf("=") == -1)  // not key=value form
```
englehardt: What if we have non key-value data that contains `=`?

automation/Extension/firefox/lib/http-post-parser.js
```diff
@@ -0,0 +1,250 @@
...
+  this.body = 0;
+  if (this.seekablestream instanceof components.interfaces.nsIMIMEInputStream) {
+    this.seekablestream.QueryInterface(components.interfaces.nsIMIMEInputStream);
+    this.hasheaders = true;
+    this.body = -1;
+  } else if (this.seekablestream instanceof components.interfaces.nsIStringInputStream) {
+    this.seekablestream.QueryInterface(components.interfaces.nsIStringInputStream);
+    this.hasheaders = true;
+    this.body = -1;
+  }
```
englehardt: The upload stream can be a `nsIMIMEInputStream`, and `nsIStringInputStream`, or some third option (i.e. the default, which is what happens when line 18 and line 22 of `http-post-parser.js` are false)? The first two of which have which have headers but the default option which won't? This is confusing, what is the default stream interface type and why doesn’t it have headers? Are this third options headers included in the main HTTP headers?

automation/Extension/firefox/lib/http-post-parser.js
```diff
@@ -0,0 +1,250 @@
...
+      // match can return null...
+      if (keyValue) {
+        this.postHeaders[keyValue[1]] = keyValue[2];
+      } else {
+        this.postLines.push(line);
+      }
+      line = this.readLine();
+    }
+    this.body = this.tell();
+  }
```
englehardt: It was very confusing to me what headers you are parsing here. I originally thought the stream you were accessing contained the entire request and thus you needed to parse the HTTP request headers as well as the post body content. It wasn’t clear to me why some request streams might not have headers and why there was any uncertainty in extracting them. It wasn’t until I walked through the `parsePostRequest` method that I found the link to the stack overflow explaining post headers (line 134 of `http-post-parser.js`). Can we add a comment to this method that explains how these are distinct from request headers and why there is ambiguity in parsing them. A simple example request might be helpful to understand HTTP request headers, post-body headers, and headers within multipart post-bodies.

test/test_http_instrumentation.py
```diff
@@ -114,3 +114,78 @@ def test_javascript_saving(self, tmpdir):
...
+                                       WHERE method = 'POST'")
+
+    def get_post_request_body_from_db(self, db):
+        """Return the body of the first POST request in crawl db."""
+        posts = self.get_post_requests_from_db(db)
+        return posts[0][19]  # Column 19 is post body
+
+    def test_record_post_data_x_www_form_urlencoded(self):
+        enc_type = "application/x-www-form-urlencoded"
+        db = self.visit("/post_request.html?enc_type=" + enc_type)
```
englehardt: Shouldn't this be `db = self.visit("/post_request.html?encoding_type=" + enc_type)`

test/test_http_instrumentation.py
```diff
@@ -114,3 +114,78 @@ def test_javascript_saving(self, tmpdir):
...
+        post_body = self.get_post_request_body_from_db(db)
+        assert post_body == self.post_data_multiline
+
+    def test_record_post_data_text_plain(self):
+        enc_type = "text/plain"
+        db = self.visit('/post_request.html?enc_type=' + enc_type)
```
englehardt: Shouldn't this be `db = self.visit('/post_request.html?encoding_type=' + enc_type)`

automation/Extension/firefox/data/create_http_requests_table.sql
```diff
@@ -17,5 +17,6 @@ CREATE TABLE IF NOT EXISTS http_requests(
   loading_href TEXT,
   req_call_stack TEXT,
   content_policy_type INTEGER NOT NULL,
-  time_stamp TEXT NOT NULL
+  time_stamp TEXT NOT NULL,
+  post_body TEXT
```
gunesacar: Sure, done.

automation/Extension/firefox/lib/http-post-parser.js
gunesacar: See my new comments in the updated code.

These two stream types may contain headers which is verified by checking the stream types and upstream headers encountered in our `TestPOSTInstrument` tests.

During the verification I found that `nsIMultiplexInputStream` (used when the AJAX payload is `FormData`) may also contain headers and added a check for it.

MDN lists tons of other stream types, although I think not all of them can be used in the POST context:
https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XPCOM/Guide/Streams

We don't really know if these other stream types may contain headers or not (not documented).

We added a debug log for unknown stream types, which we can investigate later.

automation/Extension/firefox/lib/http-post-parser.js
```diff
@@ -0,0 +1,250 @@
...
+  return this.seekablestream.tell();
+};
+
+HttpPostParser.prototype.readLine = function() {
+  var line = "";
+  var size = this.stream.available();
+  for (var i = 0; i < size; i++) {
+    var c = this.stream.read(1);
+    if (c == '\r') {
+    } else if (c == '\n') {
```
gunesacar: Added `continue` to make it explicit.

automation/Extension/firefox/lib/http-post-parser.js
```diff
@@ -0,0 +1,250 @@
...
+      break;
+    } else {
+      line += c;
+    }
+  }
+  return line;
+};
+
+// visitor can be null, function has side-effect of setting body
```
gunesacar: Stray comment, removed.

automation/Extension/firefox/lib/http-post-parser.js
```diff
@@ -0,0 +1,250 @@
...
+HttpPostParser.prototype.headers = function() {
```
gunesacar: Renamed to `extractUploadStreamHeaders`, hope that's fine.

automation/Extension/firefox/lib/http-post-parser.js
gunesacar: Added comments and an example request with upload stream headers.

automation/Extension/firefox/lib/http-post-parser.js
```diff
@@ -0,0 +1,250 @@
...
+    });
+    return JSON.stringify(obj);
+  }catch (e) {
+    console.log("Exception: Cannot parse POST data:", e, encodingType, "formData:", formData);
+    return formData;  // return the original body if we can't decode
+  }
+}
+
+
+HttpPostParser.prototype.parsePostRequest = function(encodingType){
```
gunesacar: Added the comment

test/test_http_instrumentation.py
gunesacar: Yes, fixed

automation/Extension/firefox/lib/http-post-parser.js
```diff
@@ -0,0 +1,250 @@
...
+  var postString = "";
+  try {
+    // This is to avoid 'NS_BASE_STREAM_CLOSED' exception that may occurs
+    // See bug #188328.
+    for (var i = 0; i < size; i++) {
+      var c = this.stream.read(1);
+      c ? postString += c : postString += '\0';
+    }
+  } catch (ex) {
+    console.log("Error parsing the POST request", ex);
```
gunesacar: Changed to `loggingDB.logError`

automation/Extension/firefox/lib/http-post-parser.js
gunesacar: Good catch!
I started to filter out all `Content-*` headers to prevent this.

gunesacar: I added `test_record_file_upload` to make sure we don't parse inline headers as content in a file upload scenario.

automation/Extension/firefox/lib/http-post-parser.js
gunesacar: Good catch!

I added a failing test case (`test_record_post_data_ajax_no_key_value_base64_encoded`), fixed it and then [added comments about the issue](https://github.com/gunesacar/OpenWPM/commit/84f1a79122dad0d65f2cb50e65caa9fd63054758#diff-455c9b9d783e44486d3b4a7cb6fe0ec1R149).

Still, there's a risk of "over-parsing". We've some naive checks to infer key=value data. We may work on improving them.

The problem is that sender can craft POST data in any format, including ambiguous forms where it's impossible to tell non-key value data that contains `=` form from genuine form-data.

I checked how Firefox devtools interface handles ambiguous cases, and to my surprise it does over-parse as well:

New test mentioned above sends a base64 string using jQuery's `post` with default arguments:
`$.post("bogus_submit.html", "dGVzdEBleGFtcGxlLmNvbSArIG5hbWUgc3VybmFtZQ==");`

Checking the Params panel for the POST request, `dGVzdEBleGFtcGxlLmNvbSArIG5hbWUgc3VybmFtZQ== ` is parsed as `key:dGVzdEBleGFtcGxlLmNvbSArIG5hbWUgc3VybmFtZQ` and `value: ""` ('=' is perhaps lost due to split or decoding)

base64_post

gunesacar:
Thanks for the very detailed review. Helped a lot to improve the code.

I did my best to address all your comments, but the following issue may need some attention: https://github.com/citp/OpenWPM/pull/125#discussion_r93650086

englehardt:
Awesome work on the updates, the updated comments are super clear. It's also very cool to see that we can capture uploads with this.

Regarding your concern of over-parsing, I think this is totally fine. Given what you've told me about your manual checks the benefits of having pre-parsed request bodies far outweighs the random improperly parsed object. Even then, the "data loss" is really minimal since we simply end up with a string that's split into key,value when it shouldn't be. Plus, if we save the pre-parsed post bodies instead, downstream users of the data will likely face the same problem anyway.

Thanks again!

PR #126, opened 2016-12-21
Extension logging

dreisman:
Adding logging to Firefox extension, along with some basic log messages.

Future work, which I'd like to save for a different PR:
-Adding logging to the content script: I'd rather not couple this enhancement to that one, though they are similar.

It isn't productive to write tests that test specific log messages - if you can think of a more robust way to automate the check that logging works let me know.

### DIFF
--- a/automation/Extension/firefox/lib/loggingdb.js
+++ b/automation/Extension/firefox/lib/loggingdb.js
@@ -93,6 +93,58 @@ exports.logInfo = function(msg) {
     logAggregator.send(['EXT', JSON.stringify(log_json)]);
 };
 
+exports.logDebug = function(msg) {
+    // Always log to browser console
+    console.log(msg);
+
+    if (debugging) {
+        return;
+    }
+
+    // Log level DEBUG == 10
+    var log_json = makeLogJSON(10, msg);
+    logAggregator.send(['EXT', JSON.stringify(log_json)]);
+};
+
+exports.logWarn = function(msg) {
+    // Always log to browser console
+    console.warn(msg);
+
+    if (debugging) {
+        return;
+    }
+
+    // Log level WARN == 30
+    var log_json = makeLogJSON(30, msg);
+    logAggregator.send(['EXT', JSON.stringify(log_json)]);
+};
+
+exports.logError = function(msg) {
+    // Always log to browser console
+    console.error(msg);
+
+    if (debugging) {
+        return;
+    }
+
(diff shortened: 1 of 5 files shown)


dreisman:
Still checking failing tests...

dreisman:
Tests should be all passing now.

englehardt:
This is awesome, I did a quick scan of the code and things look good. Would you mind to further debug the test failures? Once you're able to do that I'll do a full review and merge.

Thanks Dillon!

dreisman:
Zoinks, it was a breaking typo that didn't get committed. As long as travis passes here I'm good for now.

automation/Extension/firefox/lib/loggingdb.js
```diff
@@ -7,18 +7,27 @@ var visitID = null;
 var debugging = false;
 var sqliteAggregator = null;
 var ldbAggregator = null;
+var logAggregator = null;
 var listeningSocket = null;
 
-exports.open = function(sqliteAddress, ldbAddress, crawlID) {
-    if (sqliteAddress == null && ldbAddress == null && crawlID == '') {
+exports.open = function(sqliteAddress, ldbAddress, logAddress, curr_crawlID) {
```
englehardt: The `curr_crawlID` never changes for the lifetime of a browser (and thus an extension instance), can we revert this name change back to `crawlID`?

automation/Extension/firefox/lib/loggingdb.js
```diff
@@ -51,6 +61,88 @@ exports.close = function() {
...
+
+exports.logInfo = function(msg) {
+    // Always log to browser console
+    console.log(msg);
+
+    if (debugging) {
+        return;
+    }
+
+    // Log level INFO == 20
```
englehardt: Can you include a link to the log level string --> integer mapping here?

dreisman: done.

automation/Extension/firefox/lib/loggingdb.js
dreisman: There was a javascript variable scope bug that didn't crop up before because crawlID was never actually used here. Basically, the naming collision between the function scoped "crawlID" and the global "crawlID" meant that the global crawlID was never getting set.

englehardt: Makes perfect sense, thanks for pointing that out!

PR #127, opened 2016-12-21
Add instrumentation for document.cookie

gunesacar:
Log cookie read and writes, capture call stack.

OpenWPM already monitors cookie change events in `cookie-instrument.js`,
but it doesn't capture calls to read cookies.

Remove leading and trailing whitespace from Javascript call stack.


englehardt:
Looks great, thanks Gunes!

PR #129, opened 2016-12-28
Don't complain for visitID being null in debugging mode.

### LINKED ISSUE #128: Don't complain about null visitID when debugging is true
[`createInsert`](https://github.com/citp/OpenWPM/blob/master/automation/Extension/firefox/lib/loggingdb.js#L199-L202) complains about visitID being null even if the `debugging` is true (i.e. when the [query will not be executed](https://github.com/citp/OpenWPM/blob/master/automation/Extension/firefox/lib/loggingdb.js#L151-L159)).

This creates a flood of errors when using `manual_test.py`to test new features.

gunesacar:
When `debugging` is true, we don't expect to have a valid `visitID`
and we don't execute SQL statements.

Fixes #128.


englehardt:
Good catch! Thanks!

PR #131, opened 2016-12-31
Refactor utilities, add a database utility module

gunesacar:
Add a new utility module for DB related
functions. This is to be used by test and data analysis code.

Use shutil's rmtree, remove file_utils.py.
Move domain_utils.py to automation/utilities.

Use domain_utils' `get_psl`, remove repeated PSL related code.


automation/Commands/profile_commands.py
```diff
@@ -62,7 +61,8 @@ def save_flash_files(logger, browser_params, dump_location, clear=False):
 
         if clear:
             logger.debug("BROWSER %i: CLEARING %s during flash file archive" % (browser_params['crawl_id'], location))
-            rmsubtree(location)
+            shutil.rmtree(location)
```
englehardt: I think this change might break things, but I'm not completely sure. If I remember correctly, the reason for adding `rmsubtree` is that `shutil.rmtree` deletes `location` (the parent directory, `#SharedObjects`), so instead we spider within the directory and delete all files and subdirectories.

Did you confirm that flash will recreate the `#SharedObjects` directory after deleting it?

automation/Commands/utils/webdriver_extensions.py
```diff
@@ -9,7 +9,7 @@
 import random
 import time
 
-import domain_utils as du
+from ....automation.utilities import domain_utils as du
```
englehardt: Nit: You should just be able to do `from ...utilities import domain_utils as du` since this file and the utilities file are both under `automation`.

automation/Commands/profile_commands.py
gunesacar: I totally missed that. I rolled back the `file_utils.py` related changes.

automation/Commands/utils/webdriver_extensions.py
gunesacar: Thanks, fixed.

PR #132, opened 2016-12-31
Distribute Travis tests to four concurrent jobs.

gunesacar:
This is to reduce the test wall time.
With this change Travis tests will take 7-8 mins (was 20+mins).

Measure and print the slowest ten tests.


englehardt:
This is great! I didn't know we could run concurrent tests. :)

PR #133, opened 2017-01-06
Fix and expand instrumentObject javascript instrumentation and add tests

englehardt:
The current `instrumentObject` javascript call instrumentation assumes that all properties being instrumented are accessor properties, and thus fails to work correctly for data properties. This pull request:

* Fixes `instrumentObject` for both accessor and data properties
* Refactors `instrumentObject` to better handle function instrumentation. It was previously possible to miss certain instrumentation when functions were overwritten by values or vice-versa. Similarly, when a function is overwritten by a new function, the new function is now instrumented on-the-fly.
* Simplifies the API -- it's no longer necessary to specify when the object being instrumented is a prototype of the object to be monitored.
* Adds optional "recursive" instrumentation that will instrument any nested objects within the object in question
* Adds tests for `instrumentObject` along with a "testing" mode for OpenWPM, which makes it possible to only use certain configuration options when set.

There are still some limitations
1. Unlike before, every `get` is recorded, even for function calls and nested objects. Though this is more correct, it adds noise to the logs. I've decided to keep this functionality so if any object were overwritten, we would still see gets and sets of the new object (even if the new object isn't instrumented -- see below).
    1. `console.log(test.obj1.obj2.prop1);` will log a `get` for `obj1`, `obj2`, and `prop1` when `test` is instrumented recursively.
    2. `test.methodCall(argument1);` will log a `get` for `test.methodCall` as well as a `call` for the function.
2. When instrumenting recursively, objects overwritten with new objects do not re-instrument the new object.


englehardt:
@gunesacar if you're able to give this a sanity check I'd appreciate it!

automation/Extension/firefox/data/content.js
```diff
@@ -351,106 +354,180 @@ function getPageScript() {
...
+    function isObject(object, propertyName) {
+      try {
+        var property = object[propertyName];
+      } catch(error) {
+        return false;
+      }
+      if (property === null) { // null is type "object"
+        return false;
+      }
+      return typeof property == 'object';
```
gunesacar: Not sure if it matters here but maybe `===` instead of `==`?

automation/Extension/firefox/data/content.js
```diff
@@ -351,106 +354,180 @@ function getPageScript() {
...
-          return function(){
-            if (!originalGetter){
-              console.log("originalGetter is undefined!")
-              logValue(objectName + '.' + property, "", "get(failed)", callContext, logSettings);
+          return function() {
+            var origProperty;
+            var callContext = getOriginatingScriptContext(!!logSettings.logCallStack);
+
+            // get original value
+            if (!originalGetter && !('value' in propDesc)) {
```
gunesacar: Any reason to use negations in this if-else clause?
I though could be written more "positively", by something like `if (originalValue) else if (originalGetter) else { error } ...`

automation/Extension/firefox/data/content.js
```diff
@@ -351,106 +354,180 @@ function getPageScript() {
...
+              console.error("Property descriptor for",
+                            objectName + '.' + propertyName,
+                            "doesn't have getter or value?");
+              logValue(objectName + '.' + propertyName, "",
+                  "get(failed)", callContext, logSettings);
               return;
+            } else if (!originalGetter) { // if data property
```
gunesacar: See above comment

automation/Extension/firefox/data/content.js
```diff
@@ -351,106 +354,180 @@ function getPageScript() {
...
-        if (prototype) {
-          instrumentPrototypeProperty(object, objectName, properties[i], logSettings);
-        } else {
-          instrumentObjectProperty(object, objectName, properties[i], logSettings);
-        }
-      }
-    }
+        if (!!logSettings.recursive && properties[i] != '__proto__' &&
+            isObject(object, properties[i]) &&
+            (!('depth' in logSettings) || logSettings.depth > 0)) {
```
gunesacar: This beefy if condition was a bit difficult for me to follow, esp. the last line :)

I understand that this is the stopping condition for the recursion, so maybe include a short comment to explain what is being checked here.

gunesacar:
I'm not very confident reviewing this prototype magic, but FWIF looks pretty good to me :)
Glad that you wrote extensive tests, fixed PEP8 violations and moved out the expected conditions.

I've a few comments inline.

automation/Extension/firefox/data/content.js
englehardt: fixed

automation/Extension/firefox/data/content.js
englehardt: Totally agree. The original if-else used a negated condition (which made sense) so I just expanded. But it's much clearer as you suggest.

automation/Extension/firefox/data/content.js
englehardt: fixed

automation/Extension/firefox/data/content.js
englehardt: Agreed, I've added a comment that walks through it in a readable way :)

Issue #101, opened 2016-11-10
Support Firefox's Tracking Protection


englehardt:
Tracking Protection was added in Firefox 42, so we should be able to turn on support for it.

The pref is currently disabled: https://github.com/citp/OpenWPM/blob/master/automation/DeployBrowsers/configure_firefox.py#L30-L35

However, enabling the pref still won't work correctly as we remove the update URL to prevent stray HTTP requests to mozilla's servers during a crawl (see: https://github.com/citp/OpenWPM/blob/master/automation/DeployBrowsers/configure_firefox.py#L137). It should be pretty easy to figure out the API to query and manually download the list at the start of a crawl.

englehardt:
The two update urls are:
* `https://tracking-protection.cdn.mozilla.net/mozstd-track-digest256/1471874828`
* `https://tracking-protection.cdn.mozilla.net/mozstd-trackwhite-digest256/1478554625`

So we can just make these requests manually with the `requests` library and update the timestamp at the end of the url. In order to have Firefox consume these lists we probably need to set `browser.safebrowsing.provider.mozilla.lists` to `mozstd-track-digest256,mozstd-trackwhite-digest256`, but we'll need to be sure that auto-updates are not enabled.

vringar:
The basic way to reenable tracking protection is to have a `browser_params["tp_cookies"].lower() == "etp"` check in here
https://github.com/mozilla/OpenWPM/blob/65337e0c19dc857f63bbca1b3aa3773a2c49fd42/automation/DeployBrowsers/configure_firefox.py#L17-L23
and set `cookie_behaviour` to 5 if that's the case.
The only other modification required is to wrap this code block in an if to check the same thing.
https://github.com/mozilla/OpenWPM/blob/65337e0c19dc857f63bbca1b3aa3773a2c49fd42/automation/DeployBrowsers/configure_firefox.py#L98-L115

wesley-tan:
[...]

Issue #93, opened 2016-09-24
Investigate Selenium 3 and geckodriver compatibility


gunesacar:
Mozilla is already phasing out support for Selenium 2. It may be a good idea to start working towards Selenium 3 and geckodriver compatibility.

Here's a WIP branch, feel free to take it from there:
https://github.com/gunesacar/OpenWPM/tree/selenium3
https://travis-ci.org/gunesacar/OpenWPM/builds/162494675

I listed the remaining problems in the commit message and the code comments.
Most pressing ones are:
- No self-signed certificate support (https://bugzil.la/1103196)
- `send_keys` command doesn't work (hence the `tab_restart_browser`)

englehardt:
Self-signed certificates aren't an issue anymore as we'll be removing the proxy in future releases.

englehardt:
Selenium only officially supports FF 47 and below (though 45 is recommended as that's the latest ESR). Their new issue template includes the following:

```
Any issue logged here for Firefox 48 or later will be closed as a duplicate of
#2559. Our recommendation is to switch to GeckoDriver, or to continue testing
on Firefox 45 until GeckoDriver is a viable option for you.
```

Combing through some Selenium issues it seems like `FirefoxDriver` will continue to work with FF 48 and 49 as long as we use unbranded builds that don't require add-on signing (see: https://github.com/SeleniumHQ/selenium/issues/2942), so switching to unbranded builds could be a stop-gap until we can get geckodriver supported.

alexristich:
Hi! Just checking to see when this might be planned. I followed up on the workaround noted in https://github.com/SeleniumHQ/selenium/issues/1181, but didn't receive a response. Looking back at this again and hoping something might not be too far off. Also, suggestions on how to implement the workaround are also most welcome!

englehardt:
We likely won't make any headway on this until March / April when the next Firefox ESR is released. Aside from the engineering work required for OpenWPM, we are blocked on `send_keys` support, which seems to currently be supported in `GeckoDriver` but is not yet available in Selenium (See: https://github.com/SeleniumHQ/selenium/issues/2611).

englehardt:
Added in #152

Issue #92, opened 2016-09-24
OpenWPM can't be installed with Firefox 48+ release version


gunesacar:
Firefox 48+ won't allow unsigned extensions to be installed, with no override through `xpinstall.signatures.required` preference.

https://wiki.mozilla.org/Add-ons/Extension_Signing#Timeline

An alternative is to use the developer edition for the crawls, which does not require extensions to be signed.

englehardt:
We can also use [unbranded builds](https://wiki.mozilla.org/Add-ons/Extension_Signing#Unbranded_Builds) as long as we confirm that the only change branding.

englehardt:
A second issue with 48+ is e10s compatibility. Our add-on will require a few updates, see:

https://developer.mozilla.org/en-US/Add-ons/SDK/Guides/Multiprocess_Firefox_and_the_SDK

englehardt:
#152 moved us to FF52. Extension signing and multiprocess requirements don't seem to be an issue yet. The browser just provides a warning that our extension shouldn't be trusted since it isn't signed. e10s is disabled since our extension isn't compatible, but that's fine.

e10s compatibility will be a concern for the next ESR, FF 59. We will need to update to that version by June 26th, 2018, when support is ended for FF 52 ESR ([timeline here](https://www.mozilla.org/en-US/firefox/organizations/faq/)]. Since FF57 drops support for add-on-sdk-based extensions, we will likely need to make a number of changes to continue to use our instrumentation. Let's track those in #148.

Issue #88, opened 2016-09-09
Investigate changes necessary to move to Python 3


englehardt:
Python 2.7 is being [slowly removed](https://wiki.ubuntu.com/XenialXerus/ReleaseNotes#Updated_Packages) from the default installs of Ubuntu, we should investigate what changes are necessary to port to Python 3. I'm guessing some of our dependencies will be 2.7+ only.

gunesacar:
I ran [2to3](https://docs.python.org/3/library/2to3.html) on the current master. Issues related to `print, unicode, (x)range`are the overwhelming majority, which can be easily fixed.

But I can't be sure about the other problems, such as requirements without python 3 support (e.g. pyamf, libmproxy).

You can find the patch generated by 2to3 [here](https://gist.github.com/gunesacar/ca279f993a80af311559b680ebb61c98), logs are given below:

```
RefactoringTool: Files that need to be modified:
RefactoringTool: ./demo.py
RefactoringTool: ./automation/BrowserManager.py
RefactoringTool: ./automation/CommandSequence.py
RefactoringTool: ./automation/Errors.py
RefactoringTool: ./automation/MPLogger.py
RefactoringTool: ./automation/SocketInterface.py
RefactoringTool: ./automation/TaskManager.py
RefactoringTool: ./automation/platform_utils.py
RefactoringTool: ./automation/Commands/browser_commands.py
RefactoringTool: ./automation/Commands/command_executor.py
RefactoringTool: ./automation/Commands/profile_commands.py
RefactoringTool: ./automation/Commands/utils/XPathUtil.py
RefactoringTool: ./automation/Commands/utils/domain_utils.py
RefactoringTool: ./automation/Commands/utils/file_utils.py
RefactoringTool: ./automation/Commands/utils/firefox_profile.py
RefactoringTool: ./automation/Commands/utils/lso.py
RefactoringTool: ./automation/Commands/utils/webdriver_extensions.py
RefactoringTool: ./automation/DataAggregator/DataAggregator.py
RefactoringTool: ./automation/DataAggregator/LevelDBAggregator.py
RefactoringTool: ./automation/DeployBrowsers/configure_firefox.py
RefactoringTool: ./automation/DeployBrowsers/deploy_browser.py
RefactoringTool: ./automation/DeployBrowsers/deploy_chrome.py
RefactoringTool: ./automation/DeployBrowsers/deploy_firefox.py
RefactoringTool: ./automation/PostProcessing/Cookie.py
[...]

PR #143, opened 2017-03-17
Support Python 3.4+, Selenium 3.3 and Firefox 52esr

zackw:
This PR addresses issues #88, #92, #93, and maybe #101.

Selenium 2.x does not support Python 3, so it's necessary to bump the Selenium and Firefox versions, and that was actually the hardest part of the change.  MITMProxy has been completely ripped out, but I kept the proxy_queue around in case it is desirable to add some other HTTP proxy in the future (it'll just always be None and do nothing).  The only other dependency that hadn't been updated was PyAMF, which I forked and pruned down to just enough for what OpenWPM needs.

This passes the internal test suite, but it probably needs quite a bit more testing; I expect there is lots of stuff the internal test suite does not cover.  I am happy to continue working with you on this, but I need to get back to the project that this is in aid of, so I may be slow, and I'm not going to do any more testing myself, beyond what my own project involves.  Y'all are welcome to add patches on top of this, however is convenient for you.  I may add patches on top of this if I turn up more problems in my own testing.  (More bytes/str/unicode mixup bugs are practically certain to exist.)

TravisCI report for my branch: https://travis-ci.org/zackw/OpenWPM

Some notes on specific concerns from the issues and/or bits of the patch that may not seem obvious:

 * Yes, Python 2 is still supported.  I am *tempted* to start a further branch that only supports 3.6+ and uses asyncio instead of threads, as this would make it much easier to integrate with my own code, but that's probably at least another man-month of coding, and from a project management perspective I'm sure you don't want to take that plunge all at once. :wink:
 * The OpenWPM extension *seems* to load and run just fine, despite the concerns about extension signatures and e10s in #92. (I did notice something about "extension @openwpm incompatible with e10s" in firefox startup spew.)
 * I had to work around some bugs in GeckoDriver and Selenium.  The one that worries me the most is that it copies the profile, [geckodriver issue #299](https://github.com/mozilla/geckodriver/issues/299) implies that it's *supposed* to delete its copy when it exits, and profile archiving relies on it not doing that. My long-term recommendation would actually be to work directly with the [Marionette client library](https://marionette-client.readthedocs.io/en/latest/), which would eliminate *both* Selenium and GeckoDriver from the equation, but I expect that would be a lot of work and might break existing OpenWPM crawler scripts.  The Marionette API *resembles* the Selenium/Python API but is not actually a subset or superset, I don't think.
 * Selenium 3 does not expose itself to page JS, so toggling disable_webdriver_self_id no longer has any effect.  There are a number of vestiges of that kept around until we're sure we don't need to go back to Selenium 2.
 * Selenium 3 needs both geckodriver and firefox to be in $PATH, so what we do now is make sure the top-level firefox-bin directory is at the head of $PATH if it exists, and `install.sh` copies geckodriver in there.  If it doesn't exist, we go ahead and try to run system-installed firefox/geckodriver instead.
 * I un-disabled Tracking Protection but I did not actually test it.  With the removal of MITMProxy, perhaps "behind the scenes" network traffic is less of a problem now?
 * I replaced Adblock Plus with uBlock Origin and Ghostery with Disconnect, because that seems to be what privacy-conscious users are doing these days.  These are also not being tested.  I can't spare the time to develop tests for them right now.
 * There are a number of places where I rejiggered the code just to make it easier to read.  I apologize if this makes the _diff_ harder to read.


englehardt:
Thanks so much for taking this on -- this has been a looming TODO I haven't been able to get started on. It may be a little while before I am able to review and further test your changes, but I just wanted to jump in now to say I appreciate the contribution.

If you run into any issues or have any questions as you continue work on your project, please feel free to reach out.

cgoldberg:
> Selenium 2.x does not support Python 3

FWIW, Selenium 2.x has supported Python 3 for several years.

zackw:
@cgoldberg 
> FWIW, Selenium 2.x has supported Python 3 for several years.

The version that OpenWPM had pinned when I started this project didn't. :-)

englehardt:
@zackw I've been testing this over the past month or so and added a bunch of patches on top of your changes. Thanks again for the contributions! I'm going to close this in favor of #152.

PR #149, opened 2017-08-29
Installer improvements

eriktews:
Improved the installer a bit so that it's more suitable for scripting.


englehardt:
Thanks for this!

PR #164, opened 2017-11-12
Use extension based http request and responses for building cookie tables

gunesacar:
Fall back to proxy based tables if extension based http_requests
or http_responses does not exist.

### DIFF
--- a/automation/utilities/build_cookie_table.py
+++ b/automation/utilities/build_cookie_table.py
@@ -167,9 +167,16 @@ def build_http_cookie_table(database, verbose=False):
     # Parse http request cookies
     commit = 0
     last_commit = 0
-    cur1.execute("SELECT id, crawl_id, headers, time_stamp FROM http_requests_proxy \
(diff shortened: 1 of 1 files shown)


englehardt:
Thanks!

The proxy-based instrumentation was removed entirely in https://github.com/citp/OpenWPM/pull/152, and we have no plans to add it back in. Would you mind to remove the fallback?

gunesacar:
I thought people might still want to use this code to build cookie tables for the past crawl data.
But, maybe they should checkout a past release in order to do that, so that the code stays clean?

Let me know and I'll remove the fallback.

englehardt:
> I thought people might still want to use this code to build cookie tables for the past crawl data.
But, maybe they should checkout a past release in order to do that, so that the code stays clean?

My preference is cleaner code, as we've made other breaking changes related to this in the past. For example, the proxy tables used to be labeled `http_requests` and `http_responses` before having the `_proxy` suffix added.

gunesacar:
It makes sense. I removed the fall back, but kept the explicit check for empty query results, just in case.

PR #165, opened 2017-11-23
Change sqlite timeout to 60 seconds to avoid database locking problems.

eriktews:
Sometimes a query fails with "database is locked". By changing the timeout from the default value of 5 seconds to 60 seconds, we can maybe avoid that problem. The problem usually looks like this in the logs:

```
DataAggregator       - ERROR    - Unsupported query
<class 'sqlite3.OperationalError'>
database is locked
```

### DIFF
--- a/automation/DataAggregator/DataAggregator.py
+++ b/automation/DataAggregator/DataAggregator.py
@@ -24,7 +24,7 @@ def DataAggregator(manager_params, status_queue, commit_batch_size=1000):
 
     # sets up DB connection
     db_path = manager_params['database_name']
-    db = sqlite3.connect(db_path, check_same_thread=False)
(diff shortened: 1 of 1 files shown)


eriktews:
One python 3.6 test failed with:
```
3.6 is not installed; attempting download

Downloading archive: https://s3.amazonaws.com/travis-python-archives/binaries/ubuntu/14.04/x86_64/python-3.6.tar.bz2

The command "curl -s -o python-3.6.tar.bz2 ${archive_url}" failed and exited with 56 during .
```

I'm pretty sure this is not caused by this patch, so I suggest to re-run the tests.

englehardt:
> I'm pretty sure this is not caused by this patch, so I suggest to re-run the tests.

I agree. I'll let that test re-run and merge when it passes.

Thanks!

PR #171, opened 2018-02-22
Install curl and wget in the installer.

eriktews:
Some editions of ubuntu 16.04 such as the default docker images do not come with curl and wget. So just try to install them in install.sh. When they are already installed, they will maybe be updated but it is not supposed to cause any problems.


englehardt:
Perfect, thanks for catching this!

PR #176, opened 2018-03-15
Provide JavaScript 1st/3rd party information

eriktews:
By adding the fields document_url, top_document_url, and referrer_url to the javascript table, it is now possible to determine whether a JavaScript was executed in a 1st or 3rd party context. The easiest way to do that is by checking whether top_document_url is NULL, which indicates that this script is a third party JavaScript. Other evaluations are possible too.


motin:
Good news is that all three properties added in this PR is now implemented in the master branch. 
document_url and top_document_url is available here: https://github.com/citp/OpenWPM/blob/46e2dd32f31f799de3ad903fe2c49447f1ae454c/automation/Extension/firefox/lib/javascript-instrument.js#L32-L33

referrer is tracked in the http instrumentation: https://github.com/citp/OpenWPM/blob/46e2dd32f31f799de3ad903fe2c49447f1ae454c/automation/Extension/firefox/lib/http-instrument.js#L135

@englehardt, is there any merit in also storing the referrer_url property in the javascript instrumentation?

englehardt:
> Also, it may be more relevant to instrument the accessing of document.referrer property directly instead of including it as metadata to all instrumented javascript calls.

I agree, instrumenting directly is the right approach. The Javascript tables are already the largest, so I think it's best to avoid adding new columns to all records.

Closing this as the rest of the information is already available. Thanks for the contribution @eriktews and the input @motin.

PR #177, opened 2018-03-15
Add original_url and is_replace to http_requests

eriktews:
This patch adds the fields original_url and is_replace to the http_requests table. is_replace checks for Ci.nsIHttpChannel.LOAD_REPLACE and original_url stores the httpChannel.originalURI.spec value, which might be useful for redirects.


englehardt:
Given that `LOAD_REPLACE` is [slated for removal](https://bugzilla.mozilla.org/show_bug.cgi?id=1319110) and `originalURI` [returns the first URL in a redirect chain](https://bugzilla.mozilla.org/show_bug.cgi?id=1319110#c1) I don't think we should add this. We currently have redirect info in the `http_redirects` table.

Thank you for the PR!

PR #178, opened 2018-03-15
Bugfix for isThirdPartyWindow

eriktews:
When a new page is loaded, then by definition, the http request does not belong to a third party window. This corner case is currently not handled correctly by OpenWPM. To fix this, this commit sets 'isThirdPartyWindow' only to true, when 'isFullPageLoad' is false.


englehardt:
I think the naming in OpenWPM is super misleading. This case was added to allow us to replicate the tracking protection checks. See [the platform code](https://searchfox.org/mozilla-central/rev/d4b9e50875ad7e5d20f2fee6a53418315f6dfcc0/netwerk/base/nsChannelClassifier.cpp#355-375). From that code,

```
    // We check isThirdPartyWindow to expand
    // the list of domains that are considered first party (e.g., if
    // facebook.com includes an iframe from fatratgames.com, all subsources
    // included in that iframe are considered third-party with
    // isThirdPartyChannel, even if they are not third-party w.r.t.
    // facebook.com), and isThirdPartyChannel to prevent top-level navigations
    // from being detected as third-party.
```

`isThirdPartyWindow` is the result of a call to `isThirdPartyURI` which checks if the `channelURI` for the resource being loaded is third-party to the`topWindowURI`. Tracking Protection then exempts all loads which are not third-party to the top-level window or not third-party to their current window (e.g., the iframe they are loaded in). This is every more confusing because `ThirdPartyUtil` [has a method](https://searchfox.org/mozilla-central/rev/d4b9e50875ad7e5d20f2fee6a53418315f6dfcc0/netwerk/base/mozIThirdPartyUtil.idl#40-78) called `isThirdPartyWindow`, which check something along the lines of what you're suggesting in this pull request.

Really, we should change the naming in OpenWPM from `isThirdPartyWindow` to `isThirdPartyToTopWindow`.

englehardt:
Thanks for taking the time to make a bugfix PR! Closing given the discussion above. Added #185 to clarify usage in OpenWPM.

PR #181, opened 2018-04-17
Fix of browser sync/index in _distribute_command

mamamot:
Faulty comparison (first with integers, and only after that with '\*' and '\*\*') made the code crash when execute_command_sequence was called with the special '\*' and '\*\*' values. This made the demo.py disfunctional. The order was changed to make the special values work again.


englehardt:
Thank you!

PR #195, opened 2018-06-16
Fix crash due to unsuccessful closing of windows

crazyhorst:
Fixes citp/OpenWPM#166

This should fix the referenced issue.
`tab_restart_browser` will now immediately create a new window and change to it (unless the current window is `about:blank`) and let `close_other_windows` close all other windows.
`close_other_windows` will now loop through the list of window handles, as long as there's more than one in it, to kill popups caused by closing windows. This can lead to non-termination, e.g. when a window keeps reopening itself. However, this might be acceptable, as the TaskManager will kill the browser when the timeout period has elapsed.

### DIFF
--- a/automation/Commands/browser_commands.py
+++ b/automation/Commands/browser_commands.py
@@ -76,36 +76,20 @@ def close_other_windows(webdriver):
         for window in windows:
             if window != main_handle:
                 webdriver.switch_to_window(window)
+                webdriver.execute_script("window.open = null")
                 webdriver.close()
         webdriver.switch_to_window(main_handle)
 
 
 def tab_restart_browser(webdriver):
     """
-    kills the current tab and creates a new one to stop traffic
+    navigates to about:blank and kills all other windows to stop traffic
     """
-    # note: this technically uses windows, not tabs, due to problems with
-    # chrome-targeted keyboard commands in Selenium 3 (intermittent
-    # nonsense WebDriverExceptions are thrown). windows can be reliably
-    # created, although we do have to detour into JS to do it.
+    webdriver.get("about:blank")
     close_other_windows(webdriver)
 
-    if webdriver.current_url.lower() == 'about:blank':
-        return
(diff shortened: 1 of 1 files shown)


crazyhorst:
Unfortunately, this doesn't work for every website.

From an IPython session:

```
In [9]: import time

In [10]: from selenium.webdriver import Firefox

In [11]: import time

In [12]: def testcase():
    ...:     ff = Firefox()
    ...:     ff.get('https://www.afilmywap.in/')
    ...:     ff.execute_script("window.open('')")
    ...:     time.sleep(3)
    ...:     for window in ff.window_handles:
    ...:         ff.switch_to_window(window)
    ...:         print(ff.current_url)
    ...:         

In [13]: testcase()
https://www.afilmywap.in/
http://gdgzz.jyj88fqu.dxtuf.today/?sov=2657846316&hid=bffbhdrdrndrdr&v=337764G317&noaudio=1&noalert=1&noexit=1&noaudio=1&noalert=1&noexit=1&cntrl=10101&pid=12783&redid=44742&gsid=444&campaign_id=148&p_id=12783&id=XNSX.fa5zXD0PnmK10OYVLBm60XKKhvzJs098PMYbtBQfsk0-r44742-t444&impid=cca632b6-7224-11e8-a543-aa1f778d2780
```

Surprisingly enough, `window.open('')` is not guaranteed to open an _empty_ window. Maybe there's some kind of event listener that hijacks it?

crazyhorst:
I amended the commit.
`close_other_windows` works as before, except that `window.open` is set to `null` before closing the window. This should prevent windows from opening new windows by using events like `beforeunload`.
`tab_restart_window` will open `about:blank` in the active window instead of creating a new one using `window.open`. Is creating a new window essential? `window.open` seems problematic as websites can override it.

crazyhorst:
Ok, tests are failing. Giving up for now.

PR #198, opened 2018-06-28
[BUGFIX] Move timestamp to the content script part

eriktews:
This patch moves the generation of the timestamp for the message from the page script part to the content script part. The page script code is execute with all the other JavaScript code of the website, and when a website decides to change the Date prototype to something else, then the previous variant might generate incorrect timestamps. By generating the timestamp in the content script part, the timestamp is not affected by any JavaScript code running in the website.


eriktews:
I have seen that the CI checks have failed, but I'm currently not sure whether my patch is causing this or not. Can you re-run the pipelien?

englehardt:
Thanks for this. This is a good catch! Did you observe sites doing this in practice?

I just re-triggered the jobs.

eriktews:
Yes, for example https://www.motorkari.cz/ seems to do that. Just go there and type the following in your developer console:
```
(new Date()).toISOString()
```
You will see a T at the end instead of a Z. For me, this causes problems when I try to import OpenWPM results in Postgresql which does not like this for a timestamp.

eriktews:
And I still have no idea where the failed build comes from? Maybe you can try rebuilding the XPI

eriktews:
Hi, there was a typo in the patch, it should be fixed now.

englehardt:
Perfect! Thank you!

Issue #145, opened 2017-05-06
Exception in sock.accept() causes TaskManager to hang


itdelatrisu:
Context: Running a script that repeatedly creates a new TaskManager, loads a site, and then closes the TaskManager.  Pseudocode:
```
for url in sites:
	manager = TaskManager.TaskManager(manager_params, browser_params)
	command_sequence = CommandSequence.CommandSequence(url)
	command_sequence.get(sleep=1, timeout=120)
	manager.execute_command_sequence(command_sequence, index='**')
	manager.close()
```
Received this error infrequently (every ~12 hours), and then the program just hangs without exiting:
```
Exception in thread Thread-2905:
Traceback (most recent call last):
  File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/threading.py", line 754, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/.../OpenWPM/automation/SocketInterface.py", line 35, in _accept
    (client, address) = self.sock.accept()
  File "/usr/lib/python2.7/socket.py", line 206, in accept
    sock, addr = self._sock.accept()
  File "/usr/lib/python2.7/socket.py", line 174, in _dummy
    raise error(EBADF, 'Bad file descriptor')
error: [Errno 9] Bad file descriptor
```
Relevant line in `SocketInterface.py`:
```
(client, address) = self.sock.accept()
```
And when I tried catching the error, I couldn't even kill the program:
```
try:
	(client, address) = self.sock.accept()
except:
	print('sock error')  # prints successfully
	os._exit()  # program doesn't exit
```

eriktews:
I can confirm that this still seems to happen. Do you have any idea what is happening here?

Unfortunately this is hard to reproduce.

eriktews:
Hi, I can also confirm, this is happening with python3.

eriktews:
@itdelatrisu Could you try to run the code from merge request #201 and see whether this still happens. When so, please post the exception here, it should now contain more information which socketserver exactly crashes and might help to debug the issue.

eriktews:
I have one more idea, what about this line:

https://github.com/citp/OpenWPM/blob/master/automation/DataAggregator/DataAggregator.py#L27

When the connect fails with a timeout, then the 'socket' is never send here:

https://github.com/citp/OpenWPM/blob/master/automation/DataAggregator/DataAggregator.py#L35

Might this cause this behaviour?

[...]

PR #201, opened 2018-07-11
Give the thread for socketserver a name.

eriktews:
By giving the thread that accepts the socket connections in socketserver a name, the code should be easier to debug so that in case an exception happens, it is clear in which thread it happened.

This will not fix issue #145 but it will help at least to debug it.


englehardt:
Good idea. I hope this helps tracking down #145. Thanks for this!

PR #202, opened 2018-07-11
Implement some kind of popup protection

eriktews:
When 'no-popups' in the browser-configuration is set to true, websites cannot open a new tab or window by for example setting a link target to '_blank' or a similar method. This is useful when you wanna make sure that all actions happen only inside a single tab and window.

This can also be archived using the current codebase and setting additional preferences, this patch makes it just more convenient for the user.

### DIFF
--- a/README.md
+++ b/README.md
@@ -279,6 +279,8 @@ left out of this section.
   * **NOT SUPPORTED.** See [#101](https://github.com/citp/OpenWPM/issues/101).
   * Set to `True` to enable Firefox's built-in
     [Tracking Protection](https://developer.mozilla.org/en-US/Firefox/Privacy/Tracking_Protection).
+* `no-popups`
+  * Set to `True` to disable popups (or new windows/tabs) by for example clicking on a link that would open in a new window or tab.
 
 Browser Profile Support
 -----------------------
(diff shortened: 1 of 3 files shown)


automation/DeployBrowsers/configure_firefox.py
```diff
@@ -86,6 +86,11 @@ def privacy(browser_params, fp, fo, root_dir, browser_profile_path):
         shutil.copy(os.path.join(root_dir, 'firefox_extensions',
                                  'ublock_origin', 'storage.js'), ublock_dir)
 
+    if browser_params['no-popups']:
+        # Disable popups
+        fo.set_preference("browser.link.open_newwindow.restriction", 0)
```
englehardt: Mind to give a brief description of these two prefs? Even just a copy/paste from http://kb.mozillazine.org/Browser.link.open_newwindow and http://kb.mozillazine.org/Browser.link.open_newwindow.restriction?

README.md
```diff
@@ -279,6 +279,8 @@ left out of this section.
   * **NOT SUPPORTED.** See [#101](https://github.com/citp/OpenWPM/issues/101).
   * Set to `True` to enable Firefox's built-in
     [Tracking Protection](https://developer.mozilla.org/en-US/Firefox/Privacy/Tracking_Protection).
+* `no-popups`
```
englehardt: Is my understanding correct that these two prefs would completely prevent a site from being able to create new tabs/windows? If so, let's call this `force-single-tab` and update the description to "Set to `True` to prevent pages from opening new windows/tabs, such as those that would result from a call to `window.open`."

eriktews: At least I hope that these two lines cause this behaviour, but trying to circumvent this restriction seems to be a famous research topic on the dark side of the web.

englehardt:
Closing due to inactivity.

PR #207, opened 2018-07-20
[WIP] Fix Issue 205: add pinned pipenv to replace pip and virtualenv

stephendonner:
Not (yet) ready for review on this, although I welcome comments/feedback, of course!

Caveat: Ideally, I'd be ```git -rm```ing ```requirements.txt``` **and** ```git add```ing the resulting ```Pipfile``` from ```pip install -r requirements.txt``` as a swap-in replacement, but I need to be careful with at least two issues:

1. Make sure the current [WIP] PR passes in Travis CI, with multiple Python versions
2. Ditto for once we remove ```requirements.txt``` and add the ```Pipfile```


stephendonner:
Having issues getting both ```plyvel``` and its required, version-compatible ```LevelDB``` packages installed and happily configured, using ```pipenv``` - even with the new Dockerized setup.

Help appreciated!

stephendonner:
Deferring work here, for now; might come back to it once I've caught up with quite a few other things.

PR #210, opened 2018-07-23
[WIP] Run Travis tests using pytest-xdist for concurrency

stephendonner:
Got a ways to go, here; we'll have to smartly deal with having enough unique webserver ports, dynamically, if we're going to go the pytest-xdist route, due to the current hardcode in https://github.com/citp/OpenWPM/blob/7be2dec15807a46f86876eea480696b1584ab9ec/test/utilities.py#L13

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,16 +8,7 @@ python:
  - "3.5"
  - "3.6"
 env:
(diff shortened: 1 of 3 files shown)


stephendonner:
I'm going to close this from my side, but if anybody wants to take it, etc., feel free!

PR #213, opened 2018-07-30
Refactor data handling to use a generic format internally.

englehardt:
The platform currently uses SQL throughout all of the instrumentation. This makes it impossible to provide the option of an alternate data format. This commit moves all SQL logic into a SQLiteAggregator and uses simple dictionaries for passing data through the rest of the platform. The BaseAggregator and BaseListener classes can be used to support alternate output formats.

The internal format is still structured, as follows:

(`instrument_name`, `dict` that maps field name to data)

Thus, when the data is save to a SQL database we map `instrument_name` --> table, `field_name` --> column, and `data` to the values saved. Future aggregators can handle this differently.

I have special cased `create_table` as an instrument name to allow tables to be generated on the fly. Future aggregators can drop these records.

This PR also removes the `extract_links` command, since it produces a table which is not keyed with the rest of the instrumentation and thus was not worth updating.


turban1988:
Is there any alternative to the ``extract_links`` command?

englehardt:
@turban1988 nothing that's built into the platform

PR #219, opened 2018-09-17
Mac developer support

motin:
Work in progress. Haven't succeeded in getting the tests to run yet, neither locally nor via Docker for Mac.


motin:
This works now, ready for merge as soon as the Travis tests pass.

automation/utilities/platform_utils.py
```diff
@@ -30,22 +30,42 @@ def parse_http_stack_trace_str(trace_str):
...
+    return it. Else, throw a RuntimeError.
     """
-    root_dir = os.path.dirname(__file__)  # directory of this file
-    ffbin = os.path.abspath(root_dir + "/../../firefox-bin")
-    if os.path.isdir(ffbin):
-        curpath = os.environ["PATH"]
-        if ffbin not in curpath:
-            os.environ["PATH"] = ffbin + os.pathsep + curpath
+    if "FIREFOX_BINARY" in os.environ:
+        firefox_binary_path = os.environ["FIREFOX_BINARY"]
```
englehardt: Let's refactor this slightly to make debugging easier since this has been a common mistake on first install. In this branch let's add

```python
If not os.path.isfile(firefox_binary_path):
    raise RuntimeError(
        "No file found at the path specified in environment variable `FIREFOX_BINARY`. 
        "Current `FIREFOX_BINARY`: %s" % firefox_binary_path)
return firefox_binary_path
```

automation/utilities/platform_utils.py
```diff
@@ -30,22 +30,42 @@ def parse_http_stack_trace_str(trace_str):
...
+    else:
```
englehardt: remove `else` since we'll return early in as mentioned in the comment above

automation/utilities/platform_utils.py
```diff
@@ -30,22 +30,42 @@ def parse_http_stack_trace_str(trace_str):
...
+        root_dir = os.path.dirname(__file__)  # directory of this file
+        firefox_binary_path = os.path.abspath(root_dir +
+                                              "/../../firefox-bin/firefox-bin")
+    if os.path.isfile(firefox_binary_path):
+        return firefox_binary_path
+    else:
+        raise RuntimeError(
```
englehardt: Change this error meassage to: 

```python
            "The `firefox-bin/firefox-bin` binary is not found in the root "
            "of the OpenWPM directory (did you run the install script "
            "(`install.sh`)?. Alternatively, you can specify a binary location "
            "using the OS environment variable FIREFOX_BINARY.")
```

install-mac.sh
```diff
@@ -0,0 +1,49 @@
...
+hdiutil attach -nobrowse -mountpoint /Volumes/firefox-tmp firefox.dmg
+cp -r /Volumes/firefox-tmp/Firefox.app .
+hdiutil detach /Volumes/firefox-tmp
+rm firefox.dmg
+
+# Selenium 3.3+ requires a 'geckodriver' helper executable, which is not yet
+# packaged. `geckodriver` 0.16.0+ is not compatible with Firefox 52. See:
+# https://github.com/mozilla/geckodriver/issues/743
+# npm geckodriver 1.5.x = geckodriver 0.15.0
+npm install geckodriver@1.5.0
```
englehardt: Why not use the mac geckodriver from a Github release like we do for linux (i.e., https://github.com/mozilla/geckodriver/releases/download/v0.15.0/geckodriver-v0.15.0-macos.tar.gz)?

We'll be able to drop the `jpm` build tools with the ESR 60 upgrade and depending on the work required for #221 we can drop nodejs as a dependency entirely.

motin: It gives linux, mac and windows users the same installation procedure. Also, if the geckodriver releases location changes in the future, the install command will still be the same in openwpm (granted that the npm package is still maintained).

Note: I [started off](https://github.com/citp/OpenWPM/pull/219/files#diff-691148d686ec41672cee309128557845R30) by using https://www.npmjs.com/package/get-firefox to fetch the latest firefox esr as well, so using npm to install geckodriver as well was the logic choice. Later, I reverted away from get-firefox for this particular PR since get-firefox can't download older versions of firefox esr on demand (https://github.com/freaktechnik/get-firefox/issues/61). In the future, we may want to use it for downloading the latest esr version easily across linux, mac and windows.

automation/utilities/platform_utils.py
motin: Done

automation/utilities/platform_utils.py
motin: Done

automation/utilities/platform_utils.py
motin: Done

englehardt: I don't think you added this change.

PR #220, opened 2018-09-29
Add in_frame property to javascript instrumentation

motin:
This change seems to be the only meaningful difference between the main master branch of OpenWPM and [the code that was used for the Overscripted crawl](https://github.com/RuizhiYou/OpenWPM/tree/RunThis).

Adding this PR to gather feedback and discuss whether or not it ought to be included in the main branch (if so, of course it should have accompanying tests).

### DIFF
--- a/automation/Extension/firefox/data/content.js
+++ b/automation/Extension/firefox/data/content.js
@@ -660,7 +660,11 @@ function insertScript(text, data) {
 
 function emitMsg(type, msg) {
   msg.timeStamp = new Date().toISOString();
-  msg.inFrame = window.self !== window.top;
+  try {
+    msg.inFrame = window.self !== window.top;
+  } catch (err) {
+    msg.inFrame = true; // we land here if the document into which it's embedded has a different origin
+  }
   self.port.emit(type, msg);
 }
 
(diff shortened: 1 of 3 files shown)


motin:
@englehardt: There is already an is_frame_load property within the http instrumentation. Is this sufficient for detecting whether or not the script was executed within a frame?

motin:
Just realized that this particular implementation is not very relevant after upgrading to web extensions (https://github.com/citp/OpenWPM/issues/148), since by then we will be able to read the sender.frameId property for messages sent from content scripts. frameId === 0 means top-level and frameId <> 0 means that the content script was loaded in a frame. That is a much more reliable check which is not affected by same-origin policies.

englehardt:
This can be done by checking if the `document_url` != `top_level_url`. Of course there's a chance that a page embeds an iframe with the same URL, so it's not perfect but it's sufficient for the old instrumentation.

I'd prefer to do from outside the page script, so let's use the `frameId` in the new instrumentation (if you haven't already).

@motin does that make sense? If so, let's close this.

motin:
Closing this in favor of future PRs that addresses https://github.com/mozilla/openwpm-webext-instrumentation/issues/3 in the manner that we have discussed here.

PR #225, opened 2018-10-24
Align SQL and parquet schemas

motin:
I noticed these inconsistencies when creating [the client side typescript interfaces corresponding to the OpenWPM schemata](https://github.com/mozilla/openwpm-webext-instrumentation/blob/master/src/types/schema.d.ts).

Note: Adding table-name comments to the SQL schema allows for `diff -y automation/schema.sql automation/DataAggregator/parquet_schema.py` to present a side-by-side comparison of the schemas.


englehardt:
I cherry-picked this into #226 since I also made some schema changes there. Thanks!

PR #226, opened 2018-10-25
Full Parquet support and performance improvements

englehardt:
This PR addresses several remaining tasks for parquet writing support:
1. Batch Parquet writes so the database isn't completely fragmented.
2. Support for saving files and page content to S3
3. Refactor the local data aggregator to handle both sqlite and leveldb writes.


englehardt:
We can cut down on content saving time by saving all content hashes that have been sent to S3 locally in memory. This is especially important when crawls are not running on EC2 machines. Will file a follow-up issue for this.

englehardt:
I filed #230 and #231 for follow-up work.

Issue #235, opened 2018-11-15
Re-enable CI now that we've transferred orgs


englehardt:
We will no longer be running CI tests on citp/OpenWPM.

PR #244, opened 2018-11-27
Fix Travis badge

stephendonner:
@englehardt r?


englehardt:
Nice! Looks like this is all that's needed for #235.

Issue #258, opened 2019-02-08
Function get_psl shoudl use logger instead of print


turban1988:
Hi,
I noticed that the function` get_psl()` uses `print` statements instead of a logger. 

https://github.com/mozilla/OpenWPM/blob/b3ead7e38892095950806e8bcbb2e1129c27ca96/automation/utilities/domain_utils.py#L17-L28

If desired, I can try to fix this over the weekend.

englehardt:
This would be a nice change. In general we have an odd setup of having a separate logging process that serlializes writes from the other python processes as well as from the Firefox extension. That might lead to some complications in fixing this.

PR #260, opened 2019-02-10
Fix to #258

turban1988:
I passed the logger to the function. I do not know if this is elegant, in the opwnWPM context, but it works.


englehardt:
Closing as this PR relies on the old logger structure that will be replaced in #434.

PR #262, opened 2019-02-20
Upgrade selenium to a newer version

eriktews:
The currently used git snapshot version fails to build:

https://travis-ci.org/mozilla/OpenWPM/jobs/485662449

But at least the latest 3.141.* release of selenium seems to build fine.

I'm not exactly sure whether this fixes all problems related to selenium, but it seems to be an improvement at least.


englehardt:

PR #263, opened 2019-02-20
Set daemon to false to fix the socketio server

eriktews:
Sets the daemon flag of the BrowserManager process to false. Daemon processes in python cannot have childs and the BrowserManager tries to fork a new child process when starting the socketio server.


englehardt:

PR #264, opened 2019-02-20
Fix a few syntax bugs

eriktews:
A few strings have been line wrapped, which broke the string formatting using the percent sign. In general, "a" + "%s" % "b" works in python, but "%s" + "b" % "a" will not work since the % binds first to "b", before the + can be applied. This commit fixes this by wrapping all line wrapped strings with brackets.


englehardt:

PR #265, opened 2019-02-20
Develop all fixes

eriktews:
That's a combined pull request for #262 #263 and #264 and I hope it runs at least partially fine through travis CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,7 @@ git:
 before_install:
   - "export DISPLAY=:99.0"
 install:
+  - echo "y" | ./install.sh
   - echo "y" | ./install-dev.sh
   - pip install -r requirements.txt
 before_script:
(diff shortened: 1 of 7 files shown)


eriktews:
So the tests still fail, but at least the build runs properly and the tests actually start without breaking immediately.

englehardt:
Thanks @eriktews. We decided not to continue with the upgrade to socketIO and instead to rely on an experiment API to continue to use the old socket. This saves the work of replacing the old socket interface or maintaining two separate interfaces, and forces us to do the upfront work of supporting web extension experiments. This will give us more flexibility to run measurements that wouldn't otherwise be possible in a standard web extension.

#267 is the new starting point, and I see that you've already made a PR to that branch. I'll keep this open for now, but will likely close this in favor of a new branch building on #267.

englehardt:

PR #270, opened 2019-02-27
Also try to test the Dockerfile

eriktews:
Also try to test the Dockerfile in travis-ci

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,3 +30,17 @@ before_script:
 script:
   - cd test
   - py.test -s -v --durations=10 $TESTS
+
+jobs:
+  include:
+    language:
+    python:
+    env:
+      - TESTS="Docker"
+    services:
+      - docker
+    before_install:
+    before_script:
+    install:
+    script:
+      - docker build -f Dockerfile -t openwpm .


eriktews:
I think this MR is now ready to be merged. It also tests the Dockerfile on Travis-CI, but the test currently fails since the Dockerfile (or the extension) is still broken. I hope that gets fixed soon.

eriktews:
I opened a separate bug for the problem with the extension:

https://github.com/mozilla/openwpm-firefox-webext/issues/12

As soon as this is fixed, the additional test should run fine but other tests might still fail due to other problems.

.travis.yml
```diff
@@ -30,3 +30,17 @@ before_script:
 script:
   - cd test
   - py.test -s -v --durations=10 $TESTS
+
+jobs:
+  include:
+    language:
```
englehardt: Are these blank fields required?

englehardt:
@eriktews mind to rebase this on the current master that includes your Dockerfile fixes?

.travis.yml
eriktews: Yes, as far as I can see they are required or at least useful since they unset what has been set previously for the none docker build tasks. Here is a test with those lines removed, which shows that Travis-CI goes through the (useless) setup steps which are not required for the Docker test:

https://travis-ci.org/eriktews/OpenWPM/jobs/533198065

eriktews:
Since c175dd8ea1ca620fbdac3e6a9265f0a6be0c5025, the test for the Dockerfile runs fine, so that this should be fine to merge right now.

eriktews:
> @eriktews mind to rebase this on the current master that includes your Dockerfile fixes?

Rebase done

PR #271, opened 2019-02-28
Fix one more problem with checking json results

eriktews:
This fixes one more problem with checking json results


englehardt:
Thanks again @eriktews for helping to get these tests passing.

Issue #289, opened 2019-04-24
content.js "could not establish connection" errors in the console


englehardt:
See #288 for an example. `JavaScript error: moz-extension://6cbcf7fe-c654-4558-bfa1-bcc2f8a7769c/content.js, line 983: Error: Could not establish connection. Receiving end does not exist.`

Possible cause? https://stackoverflow.com/questions/44673632/firefox-webextension-error-could-not-establish-connection-receiving-end-does-n

nhnt11:
This happens when the JS instrument is not enabled - we are injecting the instrumentation anyway but nothing is listening for the messages in the privileged scope.

The solution is to either only inject our instrumentation if the JS instrument is enabled in the config, or to silently ignore this error by wrapping the block in a try/catch. The problem with the former is that it will likely cause an async delay while injecting our script which will cripple its utility. The problem with the latter is that it will suppress legitimate errors.

englehardt:
We should figure out how to do the former. The `content.js` script mucks with a lot of the JS APIs, so there may be crawls where we want it disabled (especially if we're not collecting JS data for those crawls).

tooflop101:
I solved the problem # 288 by installing the repository 0.8 by downloading it from the link. Errors occurred when installing via the console of an unknown version through the command git clone.Version 0.8 is stable.

motin:
This was fixed in #375 along with #374

Issue #242, opened 2018-11-19
Ensure WebExtensions inject content script instrumentation before any page scripts run


englehardt:
The instrumentation should block page script execution until all of the instrumentation injected by the `js_instrument` is set up. We should determine whether this is actually an issue or if we are only experience this during the initial page load as described in #224. See also: gbaptista/luminous#55 and EFForg/privacybadger#1865.

Note that even if this isn't an issue right now, we may experience it once we fix #68.

englehardt:
In #291 we enabled "testing mode" for all content script injections. When we fix this, we should reverse that and rely on the `testing` parameter once again.

nhnt11:
@jallmann Thanks for that investigation! I'm curious - could you share your test extension somewhere? Did you try putting a small delay in the content script and then checking if page scripts are affected by that delay?

jallmann:
This is the extension I built for testing: 
https://github.com/jallmann/webExtTest

The repo includes a test page and script with some console prints to illustrate the order of events. 
The extension is currently configured to register content scripts 5 seconds after it loads. If the test page is loaded wihtin that interval, the page script runs without the content script being injected at all. 

>  Did you try putting a small delay in the content script and then checking if page scripts are affected by that delay?

I not sure what you mean. Some artificial, blocking delay in the content script? I haven't tried that, but I would expect it to delay page script execution as well.

nhnt11:
> This is the extension I built for testing:
> https://github.com/jallmann/webExtTest

Thanks!

> The repo includes a test page and script with some console prints to illustrate the order of events.
> The extension is currently configured to register content scripts 5 seconds after it loads. If the test page is loaded wihtin that interval, the page script runs without the content script being injected at all.

That sounds right. 👍

> > Did you try putting a small delay in the content script and then checking if page scripts are affected by that delay?
> 
> I not sure what you mean. Some artificial, blocking delay in the content script? I haven't tried that, but I would expect it to delay page script execution as well.

Yup, that's what I meant.

jallmann:
Yes, blocking the content script for some time delays execution of the page script as well, just as expected.

nhnt11:
I think we can close this!

Issue #68, opened 2016-04-13
Javascript instrumentation should be configurable per-API


englehardt:
In the current implementation, javascript instrumentation is configured at the bottom of [content.js](https://github.com/citp/OpenWPM/blob/master/automation/Extension/firefox/data/content.js#L390-L442). As the number of APIs grows and the number of logging options increases, this will not scale and will lead a much unnecessary instrumentation.

Instead, we should consider having `content.js` parse an external configuration file. One format we've considered is the following:

```
'<API Name>: {
    logging: <all, none>,
    excluded_properties: [],
    included_properties: [],
    logging_flags: [logCallBack, logFunctionsAsStrings, ...]
},
'CanvasRenderingContext2D.prototype': {
...
},
...
```

englehardt:
Tracking in https://github.com/mozilla/openwpm-webext-instrumentation/issues/8

vringar:
https://github.com/mozilla/openwpm-webext-instrumentation/issues/8 was migrated to #353 which is closed. So I'm closing this as well

PR #291, opened 2019-04-25
Fix JS instrument test

nhnt11:
The fix is in 2 parts:
1. The function argument lists are now stored as JSON arrays and we need to modify our expected values accordingly.
2. window.instrumentObject is not exposed unless we pass testing=true to the content script. We need to find a way to reliably pass this based on config['testing']


nhnt11:
@englehardt I thought of a few ways to expose the 'testing' config flag to the content script, but they're all asynchronous (the storage API is the most straightforward), which results in the page script not getting injected in time (i.e. scripts in the page being loaded run before we inject our instrumentation).

One way to get around this is for the tests to use a special build of the add-on with the testing flag hard-coded where necessary, but this feels a bit complicated.

A quick-fix, for now, would be to just always set testing=true for the JS instrumentation. The only effect this currently has is to expose `instrumentObject` in the scope of the `window` for tests to use, and a cosmetic console message saying that we're in testing mode, every time the page script loads.

What do you think? I recommend the quick-fix, and if our future content-script work gives us a way to do this better, we can implement it then.

nhnt11:
I went ahead and pushed a commit to this PR that enables testing-mode by default for the page script. Feel free to merge this if you're okay with this solution (and r+)

englehardt:
I think the quick fix is okay for now, but we will need to solve this problem eventually. See https://github.com/mozilla/OpenWPM/issues/289#issuecomment-486872204 and https://github.com/mozilla/OpenWPM/issues/68. I think the bugs linked in https://github.com/mozilla/OpenWPM/issues/242 include an approach that will work for Firefox.

Issue #284, opened 2019-04-18
Set FIREFOX_BINARY on Mac to default to the one that install-mac-dev.sh provisions


motin:
This allows OpenWPM to run without the user having to set FIREFOX_BINARY explicitly.

Issue #278, opened 2019-04-09
ConnectionAbortedError: [Errno 53] Software caused connection abort


wagnerand:
Running `demo.py` on a macOS 10.14.4 machine yields the following errors:

```
BaseAggregator       - INFO     - Received shutdown signal!
Exception in thread Thread-1-LocalListener:
Traceback (most recent call last):
  File "/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/threading.py", line 917, in _bootstrap_inner
    self.run()
  File "/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/threading.py", line 865, in run
    self._target(*self._args, **self._kwargs)
  File "/Users/awagner/mozilla/OpenWPM/automation/SocketInterface.py", line 43, in _accept
    (client, address) = self.sock.accept()
  File "/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/socket.py", line 212, in accept
    fd, addr = self._accept()
ConnectionAbortedError: [Errno 53] Software caused connection abort

Exception in thread Thread-1-loggingserver:
Traceback (most recent call last):
  File "/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/threading.py", line 917, in _bootstrap_inner
    self.run()
  File "/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/threading.py", line 865, in run
    self._target(*self._args, **self._kwargs)
  File "/Users/awagner/mozilla/OpenWPM/automation/SocketInterface.py", line 43, in _accept
    (client, address) = self.sock.accept()
  File "/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/socket.py", line 212, in accept
    fd, addr = self._accept()
ConnectionAbortedError: [Errno 53] Software caused connection abort
```

I tracked it down to the `sleep` calls in https://github.com/mozilla/OpenWPM/blob/master/automation/DataAggregator/BaseAggregator.py#L97 and https://github.com/mozilla/OpenWPM/blob/master/automation/MPLogger.py#L122. If I remove those, the errors do not occur.

motin:
[...]

PR #296, opened 2019-05-09
Some Mac-related improvements

motin:
Allows demo crawls and manual tests to be run on a Mac natively.

Fixes https://github.com/mozilla/OpenWPM/issues/284
Works around https://github.com/mozilla/OpenWPM/issues/278

### DIFF
--- a/automation/utilities/platform_utils.py
+++ b/automation/utilities/platform_utils.py
@@ -5,6 +5,7 @@ import os
 import subprocess
 from collections import OrderedDict
 from copy import deepcopy
+from sys import platform
 
 from tabulate import tabulate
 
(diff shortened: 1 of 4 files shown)


automation/utilities/platform_utils.py
```diff
@@ -44,9 +45,14 @@ def get_firefox_binary_path():
                 "Current `FIREFOX_BINARY`: %s" % firefox_binary_path)
         return firefox_binary_path
 
-    root_dir = os.path.dirname(__file__)  # directory of this file
-    firefox_binary_path = os.path.abspath(root_dir +
-                                          "/../../firefox-bin/firefox-bin")
+    root_dir = os.path.dirname(__file__) + "/../.."
```
nhnt11: Can we do `join()` here instead of `+`?

nhnt11: I see we do `+` elsewhere, so I don't care too much about this. We don't support Windows anyway.

PR #299, opened 2019-05-12
Fix the Dockerfile

eriktews:
For some reasons, the build-extension.sh script does not work as root. This can be fixed by creating an openwpm user with sudo permissions and run the install.sh script as openwpm.

Currently, the openwpm user remains in the image, but the Dockerfile could also be changed to remove the user after the build again.

Once this is merged, I would also recommend to merge https://github.com/mozilla/OpenWPM/pull/270 since that adds the Dockerfile to the travis-ci build.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ WORKDIR /app
 RUN sed -i'' 's/archive\.ubuntu\.com/us\.archive\.ubuntu\.com/' /etc/apt/sources.list
 RUN apt-get clean -y && rm -r /var/lib/apt/lists/* -vf && apt-get clean -y && apt-get update -y && apt-get upgrade -y && apt-get install sudo -y
 
-# For some reasons, the build-extension.sh script does not work as root. We create another user and allow him to run sudo without a password. Then we run the install.sh script as openwpm, which calls build-extension.sh.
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -5,5 +5,8 @@ COPY . /app
 WORKDIR /app
 RUN sed -i'' 's/archive\.ubuntu\.com/us\.archive\.ubuntu\.com/' /etc/apt/sources.list
 RUN apt-get clean -y && rm -r /var/lib/apt/lists/* -vf && apt-get clean -y && apt-get update -y && apt-get upgrade -y && apt-get install sudo -y
-RUN ./install.sh --no-flash
+
+# For some reasons, the build-extension.sh script does not work as root. We create another user and allow him to run sudo without a password. Then we run the install.sh script as openwpm, which calls build-extension.sh.
```
nhnt11: Please use gender-neutral pronouns. s/him/it/

englehardt: Fixed

englehardt:
Thanks @eriktews. It looks like we used to do this in our previous Dockerfile (https://github.com/mozilla/OpenWPM/blob/b8f23c97982ddf5a6a6739cbd656a08ca4d6fb91/Dockerfile). I've been meaning to revisit those changes and update the documentation. I'll file a new issue for that and merge this.

PR #302, opened 2019-05-16
Fix PEP8 error related to ConnectionAbortedError in python 2.7

englehardt:
This exception is not defined in 2.7. From my testing, there doesn't seem to be an alternate exception thrown on MacOS so the best option seems to be to define a dummy exception.


englehardt:
Closing in favor of #303, which doesn't include the pack lock changes

PR #303, opened 2019-05-16
Fix PEP8 Errors on python2 caused by ConnectionAbortedError

englehardt:
This exception is not defined in 2.7. From my testing, there doesn't seem to be an alternate exception thrown on MacOS so the best option seems to be to define a dummy exception.

### DIFF
--- a/automation/SocketInterface.py
+++ b/automation/SocketInterface.py
@@ -12,8 +12,9 @@ from six.moves import input
 from six.moves.queue import Queue
 
 if six.PY2:
-    class ConnectionAbortedError(Exception):
-        pass
+    ConnectionAbortedException = socket.error
+else:
+    ConnectionAbortedException = ConnectionAbortedError
 
 # TODO - Implement a cleaner shutdown for server socket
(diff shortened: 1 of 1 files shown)


eriktews:
I'm not so sure what happens with Python 2 here. Maybe we can do something like:

```
if six.PY2:
    connectionException = socket.error
else:
    connectionException = ConnectionAbortedError
```

And then catch connectionException?

But I'm not sure whether socket.error is thrown when the bug is triggered in Python 2 or whether another exception is thrown or something else happens.

nhnt11:
@englehardt, want to merge this? Or are you looking into Erik's question about whether another exception is thrown or something else happens?

englehardt:
@nhnt11 I think Erik's suggestion is a good one. I'm going to do a quick test to see if I can determine the equivalent error in Python 2.7, even though I can't get it to trigger when running OpenWPM.

englehardt:
I decided to keep the original patch since I was unable to trigger a socket error on MacOS and this fixes the PEP8 warnings.

PR #309, opened 2019-05-21
Use base64 encoding for the response body

eriktews:
The response body might contain binary content that cannot be correctly represented by a string. This patch assumes that the openwpm-webext-instrumentation returns the response body as an Uint8Array instead of a string. It is then encoded using base64 to make it string-safe and then send to the DataAggregator of OpenWPM to be saved in a database there.

This patch needs a change in openwpm-webext-instrumentation which is currently not available in the official repository. As long as mozilla/openwpm-webext-instrumentation#45 has not been merged, the patch will not work correctly.

### DIFF
--- a/automation/DataAggregator/LocalAggregator.py
+++ b/automation/DataAggregator/LocalAggregator.py
@@ -1,5 +1,6 @@
 from __future__ import absolute_import, print_function
 
+import base64
 import json
 import os
 import sqlite3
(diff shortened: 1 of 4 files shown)


englehardt:
Thanks again for the update. This looks good!

Please also add the base64 decoding to https://github.com/mozilla/OpenWPM/blob/9d0b99cacef95235685d9bef062f667b04e6e0fb/automation/DataAggregator/S3Aggregator.py#L250-L251. Unfortunately we don't have tests for this right now, but I'm working on that in a separate PR. After that we are ready to merge.

eriktews:
I added this in 170feafdfc581690039f51ceb6e2d81f250c2102

> Thanks again for the update. This looks good!
> 
> Please also add the base64 decoding to
> 
> https://github.com/mozilla/OpenWPM/blob/9d0b99cacef95235685d9bef062f667b04e6e0fb/automation/DataAggregator/S3Aggregator.py#L250-L251
> . Unfortunately we don't have tests for this right now, but I'm working on that in a separate PR. After that we are ready to merge.

PR #310, opened 2019-05-21
Fix two HTTP tests

eriktews:
The new version of Firefox returns the HTTP POST body as key-value pairs and does not provide raw access to the body. This patch adjusts the tests to expect this behaviour.

### DIFF
--- a/test/test_http_instrumentation.py
+++ b/test/test_http_instrumentation.py
@@ -686,14 +686,16 @@ class TestPOSTInstrument(OpenWPMTest):
         post_format = "noKeyValue"
         db = self.visit("/post_request_ajax.html?format=" + post_format)
         post_body = self.get_post_request_body_from_db(db)
-        assert post_body == "test@example.com + name surname"
+        assert json.loads(post_body) == {
+            "test@example.com   name surname": [None]}
 
     def test_record_post_data_ajax_no_key_value_base64_encoded(self):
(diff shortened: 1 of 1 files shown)


test/test_http_instrumentation.py
```diff
@@ -687,14 +687,18 @@ def test_record_post_data_ajax_no_key_value(self):
...
+            "test@example.com   name surname": [None]}
 
     def test_record_post_data_ajax_no_key_value_base64_encoded(self):
         """Test Base64 encoded AJAX payloads (no key=value form)."""
         post_format = "noKeyValueBase64"
         db = self.visit("/post_request_ajax.html?format=" + post_format)
         post_body = self.get_post_request_body_from_db(db)
-        assert post_body == "dGVzdEBleGFtcGxlLmNvbSArIG5hbWUgc3VybmFtZQ=="
+        assert json.loads(post_body) == {
+            "dGVzdEBleGFtcGxlLmNvbSArIG5hbWUgc3VybmFtZQ": [""]}
```
nhnt11: Could you please file an issue to investigate this in Firefox? Assuming that we are right that Firefox is the culprit here, this is definitely a bug that the folks who work on WebExtension APIs will want to know about and fix.

nhnt11: This particular case is especially eye-catching since the key-value parsing is stripping the "==" base64 padding.

PR #311, opened 2019-05-28
Use XHR directly instead of jquery post for plaintext request bodies.

nhnt11:
Untested. This should fix the issue Erik found where the POST tests are detecting unexpectedly mutated data. See https://github.com/mozilla/OpenWPM/pull/310#pullrequestreview-242449003

### DIFF
--- a/test/test_pages/post_request_ajax.html
+++ b/test/test_pages/post_request_ajax.html
@@ -9,7 +9,7 @@
     <script type="application/javascript">
 
       function do_xhr(data) {
-        var xhr = new XMLHttpRequest;
+        var xhr = new XMLHttpRequest();
         xhr.open("POST", "bogus_submit.html");
         xhr.send(data);
       }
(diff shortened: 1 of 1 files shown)


eriktews:
I am a bit surprised by that error:

https://travis-ci.org/mozilla/OpenWPM/jobs/538062525#L2628

It looks like the request has never been sent or it is not properly stored by OpenWPM.

test/test_pages/post_request_ajax.html
```diff
@@ -8,14 +8,18 @@
     <script type="text/javascript" src="shared/utils.js"></script>
     <script type="application/javascript">
 
+      function do_xhr(data) {
+        var xhr = new XMLHttpRequest;
```
nhnt11: This should be `new XMLHttpRequest()`

eriktews: That sounds much better. Can you update your branch so that the Travis-CI tests will run again?

englehardt: Fixed this and triggered a re-run of the tests.

englehardt:
This was fixed in #316

PR #315, opened 2019-06-08
Handle HTTP POST requests with raw form data

eriktews:
A recent version of openwpm-webext-instrumentation also supports HTTP
POST request in which the POST body is only available as raw binary
data that is stored in the post_body_raw column.

This change updates the database structure with that column and adjusts
the tests so that those cases are correctly tested.

Before this can be merged, https://github.com/mozilla/openwpm-webext-instrumentation/pull/47 needs to be merged.

### DIFF
--- a/automation/DataAggregator/parquet_schema.py
+++ b/automation/DataAggregator/parquet_schema.py
@@ -83,6 +83,7 @@ fields = [
     pa.field('req_call_stack', pa.string()),
     pa.field('resource_type', pa.string(), nullable=False),
     pa.field('post_body', pa.string()),
+    pa.field('post_body_raw', pa.string()),
     pa.field('time_stamp', pa.string(), nullable=False),
 ]
 PQ_SCHEMAS['http_requests'] = pa.schema(fields)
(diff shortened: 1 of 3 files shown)


eriktews:
Maybe a few more notes about how how the raw form data is currently stored.

Firefox returns an array of 2-tuples, always a filename and the contents. The contents might be an arbitrary sequence of bytes, so it's not safe to interpret them as a string. So I decided to base64 encode the contents and then turn the entire structure in a JSON string. This string is stored in the post_body_raw field of the http_requests table.

There are a few alternatives to that:
- Use EJSON which defines a standarf for embedding binary data in JSON. In the end, this is just base64 as well, but we might use a standard de- and en-coder instead of handling the base64 manually.
- Use the standard post_body column of the table, but extend the data there a bit. We might always store a JSON encoded dict there, either with {'form': old-contents} to store the old contents as they are now, or {'raw': current-contents-of-post-body-raw} to handle the new case
- Add a new table for the raw body contents with three columns, a reference to the http_request id and then the filename and the content as blob.
- A similar structure, but with a SHA256 hash to identify the content (similar to the leveldb approach that currently exists)

Every alternative has it's own advantages and disadvantages. I'm happy to change the pull request to a different format. The good news is, that this is as far as I know the last place in OpenWPM where binary content is not yet handled properly.

PR #316, opened 2019-06-11
Mozilla rmjquery raw post

eriktews:
This is an updated PR for https://github.com/mozilla/OpenWPM/pull/311

In a nutshell, the code in the previous PR was correct, but the post body was considered to be "raw" by Firefox, so it only works when https://github.com/mozilla/OpenWPM/pull/315 has been merged.

This PR includes https://github.com/mozilla/OpenWPM/pull/315, otherwise it would not work.

### DIFF
--- a/test/test_http_instrumentation.py
+++ b/test/test_http_instrumentation.py
@@ -693,15 +693,16 @@ class TestPOSTInstrument(OpenWPMTest):
         """Test AJAX payloads that are not in the key=value form."""
         post_format = "noKeyValue"
         db = self.visit("/post_request_ajax.html?format=" + post_format)
-        post_body = self.get_post_request_body_from_db(db)
-        assert post_body == "test@example.com + name surname"
+        post_body = self.get_post_request_body_from_db(db, True)
+        assert post_body.decode('utf8') == "test@example.com + name surname"
 
     def test_record_post_data_ajax_no_key_value_base64_encoded(self):
         """Test Base64 encoded AJAX payloads (no key=value form)."""
         post_format = "noKeyValueBase64"
         db = self.visit("/post_request_ajax.html?format=" + post_format)
-        post_body = self.get_post_request_body_from_db(db)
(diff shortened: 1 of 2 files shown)


eriktews:
@nhnt11 Can you take a look at this.

test/test_http_instrumentation.py
```diff
@@ -685,15 +693,16 @@ def test_record_post_data_ajax_no_key_value(self):
         """Test AJAX payloads that are not in the key=value form."""
         post_format = "noKeyValue"
         db = self.visit("/post_request_ajax.html?format=" + post_format)
-        post_body = self.get_post_request_body_from_db(db)
-        assert post_body == "test@example.com + name surname"
+        post_body = self.get_post_request_body_from_db(db, True)
+        assert post_body == "test@example.com + name surname".encode('utf8')
```
englehardt: nit: You should use `post_body.decode('utf8')` here as you do above, rather than encode the literal.

test/test_http_instrumentation.py
```diff
@@ -685,15 +693,16 @@ def test_record_post_data_ajax_no_key_value(self):
...
 
     def test_record_post_data_ajax_no_key_value_base64_encoded(self):
         """Test Base64 encoded AJAX payloads (no key=value form)."""
         post_format = "noKeyValueBase64"
         db = self.visit("/post_request_ajax.html?format=" + post_format)
-        post_body = self.get_post_request_body_from_db(db)
-        assert post_body == "dGVzdEBleGFtcGxlLmNvbSArIG5hbWUgc3VybmFtZQ=="
+        post_body = self.get_post_request_body_from_db(db, True)
+        assert post_body == (
```
englehardt: same here.

test/test_http_instrumentation.py
eriktews: Done

test/test_http_instrumentation.py
eriktews: Done

PR #318, opened 2019-06-12
Fix test_js_profile_cookies

eriktews:
Previously, test_js_profile_cookies used _dump_profile_cookies_ which dumps the Firefox internal SQLite database to a table. This doesn't work very well anymore in the current version of Firefox and OpenWPM. Instead, the new extension provides a _javascript_cookies_ table that is filled with the cookies when _cookie_instrument_ is set to true. This commit updates the test so that the new table is used now and checks whether a cookie set by JavaScript is correctly recorded there.

### DIFF
--- a/test/test_storage_vectors.py
+++ b/test/test_storage_vectors.py
@@ -22,12 +22,19 @@ expected_lso_content_b = [
     u'REPLACEME']
 
 expected_js_cookie = (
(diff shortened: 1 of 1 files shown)


englehardt:
LGTM! I'll also make a quick PR that rips out the `dump_profile_cookies` command.

PR #319, opened 2019-06-12
Remove dump_profile_cookies command.

englehardt:
This command is no longer necessary with the new instrumentation. It was
broken by the latest Firefox upgrade, so it no longer makes sense to
keep it around.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,18 +114,10 @@ available [below](#output-format).
     * NOTE: Flash cookies are shared across browsers, so this instrumentation
         will not correctly attribute flash cookie changes if more than 1
         browser is running on the machine.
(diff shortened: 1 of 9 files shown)


nhnt11:
Just one comment that won't block my r+: I see that you changed the call to `dump_profile_cookies` to `save_screenshot` and `dump_flash_cookies`. This is technically a change of behavior, but I assume you were just being opportunistic here. Thanks for the patch!

englehardt:
Both of these changes are just in docstrings / documentation that show how to use the `CommandSequence` interface, so they should work equally well as documentation.

Thanks for reviewing!

PR #322, opened 2019-06-18
Disable file upload test

eriktews:
Firefox is currently not able to return the FormData with the file contents, currently only the filenames are returned. This is due to a limitation in the current API implementation. Therefore, this commit disables the file upload test.


englehardt:
Thanks for this! We finally have a test suite that is passing again :)

I filed https://github.com/mozilla/OpenWPM/issues/324 as a follow up.

Issue #277, opened 2019-04-03
64-bit integer columns truncated in Javascript


englehardt:
The max safe integer value in JS is [53 bits](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER). We generate 64-bit values for `visit_id` when the S3 aggregator is used, meaning some values get clipped when sent to the Javascript instrumentation. Since this same clipping doesn't happen in python, this means that the `visit_id` column is inconsistent between tables written directly in python (e.g., `site_visits`) and those written by the javascript instrumentation (e.g., `http_requests`). 

The simple solution here is to mask the [generated visit IDs](https://github.com/mozilla/OpenWPM/blob/b3ead7e38892095950806e8bcbb2e1129c27ca96/automation/DataAggregator/S3Aggregator.py#L320-L323) to 53 bits. We should also add a test for visit_id consistency across all tables when fixing this.

Note that the `expires` column of the `javascript_cookies` table is [also set to use 64 bits](https://github.com/mozilla/OpenWPM/blob/b3ead7e38892095950806e8bcbb2e1129c27ca96/automation/DataAggregator/parquet_schema.py#L149).

nhnt11:
Another simple solution is to use doubles in python-land, I think.

nhnt11:
I don't think we should mess with bits here - I assume 64-bit ints in pythonland are represented in 2's complement; whereas JS Numbers are IEEE 754 floats. Bitmasking would mess with the encoding.

englehardt:
I'm not sure that the respective implementations matter. I think things will "just work" as long as we stay within a range of numbers that can be represented both in python and in JS. My suggestion was just to make sure that we keep the integers within that safe range with generating them in python.

nhnt11:
Got it - I just saw the code that generates the ID and now understand what you mean!

englehardt:
The ability to write tests for this depends on #294, since we only generate random visit IDs for the S3 Aggregator.

PR #323, opened 2019-06-18
Make random ids JavaScript safe

eriktews:
JavaScript might only support integers up to 53 bits. This commit limits the randomly generated visid_id to 53 bits and also changes the generation of the random ids from uuid.uuid4 to random.getrandbits.

Since the S3Aggregator is currently not covered by TravisCi, the code here is unfortunately untested.

### DIFF
--- a/automation/DataAggregator/S3Aggregator.py
+++ b/automation/DataAggregator/S3Aggregator.py
@@ -4,7 +4,7 @@ import base64
 import gzip
 import hashlib
 import json
-import uuid
+import random
 from collections import defaultdict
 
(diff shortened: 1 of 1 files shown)


eriktews:
This should fix #277

englehardt:
@eriktews would you mind to add a test for this to the test file added in #333? A possible test setup would be to enable all of the instrumentation, visit a handful of the test pages, and ensure that the visit ids are consistent across all tables.

englehardt:
If we merge #386 we should close this.

PR #328, opened 2019-06-25
Merge openwpm-webext-instrumentation repo

### LINKED ISSUE #300: Adopt monorepo - ie move openwpm-webext-instrumentation into this repo?
Opening this for discussion.

I am finding various friction points when attempting to customize OpenWPM that relates to [openwpm-webext-instrumentation](https://github.com/mozilla/openwpm-webext-instrumentation) being in a separate repository. 

The main advantage of having openwpm-webext-instrumentation (the general data gathering bits that can be used in any web extension) in a separate repo is to be able to let it's development and consumption be decoupled from the OpenWPM crawl orchestration logic. 

However, the mix between monorepo and multirepo currently seems to cause a net loss in development progress, with issues and PRs being split across repos, especially for modifications that require changes in both codebases to make sense.

Also, with the decision of moving the new OpenWPM web extension from a separate repo into the main OpenWPM repo, it seems we tend towards monorepo structure also for the new webext-based OpenWPM branch.

Another (smaller) advantage of having openwpm-webext-instrumentation as a separate repository is the ability to consume it directly from github using npm without having to have an npm package published, but this can be alleviated by simply publishing an npm package.

If @englehardt and @nhnt11 are also pro going monorepo here, I can setup the PR to make that happen (landing it in `OpenWPM/automation/Extension/webext-instrumentation`). I will also publish the appropriate npm packages.

motin:
For #300


eriktews:
Hi, about the Dockerfile: Could you move both extensions in a top level folder? I'm currently working on an improved Dockerfile and having the extension code in a dedicated top level folder would make things much more beautiful and also easier.

motin:
@eriktews There is only one extension, but if I understand correctly you' suggest that we move `automation/Extension` to a top-level folder like `extension`. Sounds reasonable to me but quite a lot of scripts and documentation needs to be updated to reflect that, so I suggest it'd be done in a separate PR that specifically addresses it.
(Btw, I'll happily review new PRs related to Dockerfile changes. :))

eriktews:
> @eriktews There is only one extension, but if I understand correctly you' suggest that we move `automation/Extension` to a top-level folder like `extension`. Sounds reasonable to me but quite a lot of scripts and documentation needs to be updated to reflect that, so I suggest it'd be done in a separate PR that specifically addresses it.
> (Btw, I'll happily review new PRs related to Dockerfile changes. :))

I created a PR for that: #332 

So right now, I would just suggest to merge your PR as it is since it works with #332 too.

automation/Extension/firefox/web-ext-config.js
```diff
@@ -10,7 +10,7 @@ const defaultConfig = {
     overwriteDest: true,
   },
   run: {
-    firefox: process.env.FIREFOX_BINARY || "firefox",
+    firefox: process.env.FIREFOX_BINARY || "firefoxdeveloperedition",
```
englehardt: Why this change? From what I can tell the unbranded build is registered as "Nightly". It seems like we shouldn't need any fallback though, as we really only want to test/crawl with the firefox binary grabbed from our crawl script and throwing and error seems preferable to silently falling back to the system's firefox or dev edition binary. (If I'm misunderstanding how this is used, lmk).

motin:
> We should remove/merge/update these files:

With the idea of later splitting back this code into it's own repository, I suggest we keep some of these files:

>     automation/Extension/webext-instrumentation/.gitignore

I'd like to keep this one, since it is specific for the webext-instrumentation folder and should not be coupled with the framework gitignore rules.

>     automation/Extension/webext-instrumentation/.travis.yml

This one can be removed.

>     automation/Extension/webext-instrumentation/README.md

I'd like to keep a README / set of docs specific to this folder rather than merging with the framework docs. Again, it makes it easier to split later. 

> It may also make sense to move the editor config and npm ignore files as well (at least up one directory so they apply to both the module and the extension).

Npm ignore should be in the npm package root (ie where it is), but I am happy with merging editorconfig / code style configuration with the extension.

automation/Extension/firefox/web-ext-config.js
motin: It'd be better if this can run the bundled Fx installation, but not sure how to set that up in a cross platform way. Using firefox dev edition to start the extension is at least better than the branded firefox (which doesn't run web extension experiments).

motin: > (If I'm misunderstanding how this is used, lmk).

`process.env.FIREFOX_BINARY` allows developers to supply their preference when running npm commands, eg, `FIREFOX_BINARY=nightly npm start` or `FIREFOX_BINARY=/Applications/Firefox.app/Contents/MacOS/firefox-bin npm start`.

englehardt:
Thanks for the follow ups. Moving back to a separate repo is a good reason to keep things as is.

PR #331, opened 2019-06-25
Add .dockerignore

motin:
Somewhat synchronized with gitignore but dedicated to ignoring the largest files and directories that may be lying around in a developer's openwpm folder.

Before (on my machine):
Sending build context to Docker daemon  1.715GB

After:
Sending build context to Docker daemon  71.97MB

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,25 @@
+# Docker volume
+docker-volume
+
+# pytest
+.pytest_cache
+
+# firefox directories
+firefox-bin
+*.app
+
+# PyCharm
+.idea
+
+# Local Python virtual environments
+venv
+
+# npm packages
+automation/Extension/firefox/node_modules
+automation/Extension/webext-instrumentation/node_modules
+
+# built extension artifacts
+automation/Extension/firefox/dist
+automation/Extension/firefox/openwpm.xpi
+automation/Extension/firefox/src/content.js
+automation/Extension/firefox/src/feature.js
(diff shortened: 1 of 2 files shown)


.dockerignore
```diff
@@ -0,0 +1,25 @@
...
+firefox-bin
+*.app
+
+# PyCharm
+.idea
+
+# Local Python virtual environments
+venv
+
+# npm packages
```
eriktews: Hi, I would suggest to keep the _node_modules_ since I'm currently working on a Dockerfile-dev, which includes these files. The reason for that is that with the Dockerfile-dev, you should get a container that allows you to easily modify and rebuild OpenWPM, so the _node_modules_ are definitely useful for that.

I also created a PR for that: #332

motin: The main reason of skipping node_modules is however not performance-related, but compatibility. Node modules many times contain binary builds that are compiled for the current OS, and sending node_modules from a Mac/Windows machine into the Docker context for building a Ubuntu image is a recipe for weirdness :)

felix4webscience:
Hey Frederik,
I am just wondering, if this could be the reason, why I have problems with
node modules, when reinstalling openwpm on Linux?
Felix

Fredrik Wollsén <notifications@github.com> schrieb am Mi. 26. Juni 2019 um
10:32:

motin:
> I am just wondering, if this could be the reason, why I have problems with
node modules, when reinstalling openwpm on Linux?

Could be, but I an not sure how your dev setup looks like. A golden rule is to use different folders / checked out repositories for non-Dockerized vs Dockerized development, in order to rule out possible incompatibilities between the systems.

PR #332, opened 2019-06-25
Mozilla dockerfile dev

eriktews:
Add a development Dockerfile

The current Dockerfile is optimized to run OpenWPM, but it is less useful for development. This commit adds a Dockerfile-dev, which contains also the tools that are required to modify and test OpenWPM.

This PR makes the current Dockerfile also a bit lighter by removing everything that is not needed to just _run_ OpenWPM.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,6 +46,17 @@ jobs:
       install:
       script:
         - docker build -f Dockerfile -t openwpm .
+    - language:
+      python:
+      env:
+        - TESTS="Dockerfile-dev"
+      services:
+        - docker
+      before_install:
+      before_script:
+      install:
+      script:
+        - docker build -f Dockerfile-dev -t openwpm-dev .
     - language: node_js
       python:
       env:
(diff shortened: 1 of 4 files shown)


motin:
I am not sure what the underlying reason to a separate Dockerfile for development is? Development aided by Docker usually implies a workflow and a docker-compose file for mounting local code into the container(s) and thus have the ability to modify code and run those modifications in the Dockerized environment. This doesn't seem to support that workflow (the docs are not mentioning / describing it yet anyhow). This PR seems to introduce duplication and increased disparity between production and staging environments, so we should be gaining something / addressing a clearly articulated need. :)

eriktews:
The idea is that the container contains everything that is also present in the normal version and you can use it as a drop-in replacement to run OpenWPM. Should something fail, you can run a shell with docker exec and start debugging in the container without having to rebuild it.

Having a third Dokerfile so that you can have the OpenWPM development directory mounted in the container from your host but compile and run everything inside the container could be interesting as well. I might think about adding such a Dockerfile.

englehardt:
I prefer to avoid the complexity of introducing a second Dockerfile. Closing.

PR #333, opened 2019-06-25
Add support for testing the S3Aggregator with localstack.

englehardt:
This builds on top of the work done in #297, but cleans up the localstack code and isolates it to the test suite. The test file currently only contains a very basic correctness check.


englehardt:
@motin r? This should get #297 over the line so we can start adding additional tests in future PRs.

englehardt:
Looks like the tests are failing on CI. The test pass locally with Python 2.7. I'm getting a different set of errors with 3.6 locally. I'll have to continue to investigate.

englehardt:
I wonder if we should use the `aws_stack` interface (https://github.com/localstack/localstack/blob/master/tests/integration/test_s3.py). Maybe that will play nicer with CI...

motin:
> I wonder if we should use the aws_stack interface (https://github.com/localstack/localstack/blob/master/tests/integration/test_s3.py). Maybe that will play nicer with CI...

The current approach attempts to silently configure s3 throughout openwpm to use the mock s3 endpoint, but it was setup that way to avoid having to refactor and modify the s3 client and resource usage throughout the rest of the code. The aws_stack interface seems to allow easy setting up of a mocked s3 client, and if it can also be used with s3fs etc and some refactoring is ok, then yes it may very well be the way to go.

motin:
Conspicuously missing from the test logs (eg https://travis-ci.org/mozilla/OpenWPM/jobs/550712100) is `Starting mock S3 (http port 4572)`... but the `Ready.` output is found. It seems that S3 is not even attempted to be started when running in Travis.

motin:
The debug output for shows:

Python 2:
```
infra.py                   116 DEBUG    Unable to load plugins from module localstack: No module named google_compute_engine
```
Python 3:
```
infra.py                   116 DEBUG    Unable to load plugins from module localstack: No module named 'google_compute_engine'
```

motin:
Passing tests in Travis :) Solution/workaround was in https://github.com/travis-ci/travis-ci/issues/7940#issuecomment-311411559

englehardt:
@motin good find :) Merging.

PR #368, opened 2019-06-27
Enable navigation and js instruments for demo crawl

motin:
Just to make sure that newcomers don't get the impression that OpenWPM only does HTTP and Cookies :)

Enabling JS instrument also works around https://github.com/mozilla/OpenWPM/issues/289


englehardt:
Happy to include this in the demo! But we do need a real fix for #289.

PR #370, opened 2019-06-27
Add large payloads test page

motin:
Created this for use in manual testing mode while developing the JESTr study add-on. Seems relevant to add to the general test page collection since there are several pages out there that use canvas heavily and the platform should handle it.


englehardt:
Similar to #371, please add a test to go along with this.

englehardt:
Merging test page as-is so we don't lose track of it.

PR #372, opened 2019-06-27
Start the JS instrumentation listener as early as possible

motin:
...in order not to miss events that may arrive before the logging backend is configured (which is done asynchronously and thus page scripts may already be sending events before it has been set up).

May mitigate some timing issues somewhat.

### DIFF
--- a/automation/Extension/firefox/feature.js/index.js
+++ b/automation/Extension/firefox/feature.js/index.js
@@ -8,6 +8,13 @@ import {
 import * as loggingDB from "./loggingdb.js";
 
 async function main() {
+
+  // Start the JS instrumentation listener as early as possible
+  // in order not to miss events that may arrive before the
+  // logging backend is configured
+  let jsInstrument = new JavascriptInstrument(loggingDB);
+  jsInstrument.listen();
+
   // Read the browser configuration from file
   let filename = "browser_params.json";
   let config = await browser.profileDirIO.readFile(filename);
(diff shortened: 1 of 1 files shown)


englehardt:
I don't think we should do this. The JS instrumentation is messy -- it overwrites a number of APIs in the DOM and has caused a lot of issues in the past (in addition to making OpenWPM trivially discoverable). I know that we inject the instrumentation for all pages regardless of configuration, but as I wrote here https://github.com/mozilla/OpenWPM/issues/289#issuecomment-486872204 we should stop doing that. This patch relies on that functionality, rather than fixing it.

motin:
Hmm, the changes in this PR does not have a negative effect itself since starting the background listeners does not affect the initiation of the content/page JS instrumentation or interfere with the DOM.

Thus, in cases where the JS instrumentation is not configured for the crawl, the cost is minimal, and when JS instrumentation is configured, the changes may result in less missed early messages (during the slight delay between the background script being loaded and the configuration is available (https://github.com/mozilla/OpenWPM/pull/372/files#diff-d9ac6cc2ee6f385e2e843bbbfc099564R20)

But, I am 100% with you that we should find a good solution on how to not inject the content/page JS instrumentation all the time. Hopefully we can use the same approach to not have to wait for the configuration and can thus avoid this workaround altogether. I filed an issue to track that endeavor specifically: https://github.com/mozilla/OpenWPM/issues/374

PS I forgot to mention that this PR is the enabler of the actual functionality that was reviewed and merged in https://github.com/mozilla/openwpm-webext-instrumentation/pull/44.

motin:
@englehardt I have come around. :) I am happy to close this in favor of actually delaying the content scripts to be injected dynamically first after the configuration is loaded. Setting it up that way makes this PR useless in the context of OpenWPM crawls.

PR #375, opened 2019-06-27
Only register the JS instrument content script if JS instrument is enabled

### LINKED ISSUE #374: Only inject the JS instrumentation if the JS instrument is enabled in the config
From https://github.com/mozilla/OpenWPM/pull/372#issuecomment-506479054:
> The JS instrumentation is messy -- it overwrites a number of APIs in the DOM and has caused a lot of issues in the past (in addition to making OpenWPM trivially discoverable). I know that we inject the instrumentation for all pages regardless of configuration, but we should stop doing that.

Another symptom is #289.

This is a discussion issue to find a decent solution for this.

motin:
Fixes #374


motin:
This relates to https://github.com/mozilla/OpenWPM/issues/224 since one of the possible side effects is that the JS instrument content script may not be loaded as early as it did previously since it waits for the configuration to be available before registering the content script.

PR #382, opened 2019-07-01
Disable profile saving

englehardt:
Profile saving has long had intermittent issues. The update to Firefox 68 has fully broken profile saving. I've created [a project](https://github.com/mozilla/OpenWPM/projects/2) to track these issues, have added exceptions to OpenWPM for codepaths that rely on a browser profile (stateful crawling and profile dumping), and have disabled the profile tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -303,6 +303,11 @@ left out of this section.
 Browser Profile Support
 -----------------------
 
+**WARNING: Stateful crawls are currently not supported. Attempts to run
+stateful crawls will throw `NotImplementedError`s. The work required to
+restore support is tracked in
+[this project](https://github.com/mozilla/OpenWPM/projects/2).**
+
 ### Stateful vs Stateless crawls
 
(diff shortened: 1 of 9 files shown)


englehardt:
In 451c3babf9659fcd5816967442d1653faddbc721 I decided to take a slightly less extreme approach. Stateful crawls are still permitted, but we display a warning with each new commandsequence that's not set to clear the profile afterwards. I decided to do this in part because a bunch of our tests keep the same browser window open for performance reasons (and don't care about state). I suspect some users may have the same desire.

motin:
> In 451c3ba I decided to take a slightly less extreme approach. Stateful crawls are still permitted, but we display a warning with each new commandsequence that's not set to clear the profile afterwards. I decided to do this in part because a bunch of our tests keep the same browser window open for performance reasons (and don't care about state). I suspect some users may have the same desire.

Good compromise. Make sure to update the PR name to reflect that profile saving is actually not disabled.

PR #386, opened 2019-07-02
Fix visit_id range for JS limits.

### LINKED ISSUE #277: 64-bit integer columns truncated in Javascript

englehardt:
Will fix #277. Alternative to https://github.com/mozilla/OpenWPM/pull/323, which has errors related to signed/unsigned integers and does not pass the new included test. I verified that the master branch fails this test due to the truncation issue.

This also fixes two other issues:
* The S3 Aggregator would silently drop the last page visit (or set of visits).
* `utilities::get_test_config` accepts a browser count as an argument, but only makes the first browser headless.

### DIFF
--- a/automation/DataAggregator/BaseAggregator.py
+++ b/automation/DataAggregator/BaseAggregator.py
@@ -93,10 +93,9 @@ class BaseListener(object):
 
     def drain_queue(self):
         """ Ensures queue is empty before closing """
-        self.sock.close()
         time.sleep(3)  # TODO: the socket needs a better way of closing
-        while not self.sock.queue.empty():
-            record = self.sock.queue.get()
+        while not self.record_queue.empty():
+            record = self.record_queue.get()
             self.process_record(record)
 
 
--- a/automation/DataAggregator/S3Aggregator.py
+++ b/automation/DataAggregator/S3Aggregator.py
@@ -257,8 +257,8 @@ class S3Listener(BaseListener):
     def drain_queue(self):
         """Process remaining records in queue and sync final files to S3"""
         super(S3Listener, self).drain_queue()
-        for crawl_id in self.browser_map.keys():
-            self._create_batch(self.browser_map[crawl_id])
+        for visit_id in self.browser_map.values():
+            self._create_batch(visit_id)
         self._send_to_s3(force=True)
 
 
--- a/automation/DataAggregator/S3Aggregator.py
+++ b/automation/DataAggregator/S3Aggregator.py
@@ -257,6 +257,8 @@ class S3Listener(BaseListener):
     def drain_queue(self):
         """Process remaining records in queue and sync final files to S3"""
         super(S3Listener, self).drain_queue()
+        for crawl_id in self.browser_map.keys():
+            self._create_batch(self.browser_map[crawl_id])
         self._send_to_s3(force=True)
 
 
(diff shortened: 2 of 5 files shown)


englehardt:
Should be rebased on #382

automation/DataAggregator/BaseAggregator.py
```diff
@@ -93,10 +93,9 @@ def shutdown(self):
 
     def drain_queue(self):
         """ Ensures queue is empty before closing """
-        self.sock.close()
```
eriktews: Is that change intentional?

englehardt: Yes, we end up calling `sock.close()` twice. Once here and once in `BaseAggregator::close`.

automation/DataAggregator/S3Aggregator.py
```diff
@@ -257,6 +257,8 @@ def process_content(self, record):
     def drain_queue(self):
         """Process remaining records in queue and sync final files to S3"""
         super(S3Listener, self).drain_queue()
+        for crawl_id in self.browser_map.keys():
```
nhnt11: Can we iterate on the values instead of the keys, and directly call `_create_batch` on these? Maybe there's a python idiom to do it all in one line?

test/openwpmtest.py
```diff
@@ -40,7 +40,8 @@ def get_test_config(self, data_dir="",
             load_default_params(num_browsers)
         manager_params['data_directory'] = data_dir
         manager_params['log_directory'] = data_dir
-        browser_params[0]['headless'] = True
+        for i in range(num_browsers):
```
nhnt11: If you think it makes sense, maybe add an OS check here and exclude this on darwin

automation/DataAggregator/S3Aggregator.py
englehardt: I updated to iterating on values. I think we could also do `map(self._create_batch, self.browser_map.values())`, but I think using a loop that names the values (`visit_id`) makes it more understandable as to what we're looping over.

test/openwpmtest.py
englehardt: This is a good idea, but I think we should address it globally. I'll make a new PR that disables it in the browser deployment script.

englehardt:
Rebased on #382, actually added the test page to the PR :), and addressed Nihanth's comments. Will merge if tests pass.

test/openwpmtest.py
englehardt: See #388

PR #388, opened 2019-07-03
Disable virtual display creation for MacOS

englehardt:
We know MacOS will throw XVFB errors when we attempt to start a virtual display using `pyvirtualdisplay`. Rather than relying on a user to skip the `headless` configuration, we prevent OpenWPM from initializing the display on darwin even when the config parameter is set.

@nhnt11 or @motin: do you mind to test this on MacOS?


englehardt:
I just tested this on my macbook and it works. Thanks!

Issue #294, opened 2019-04-27
Add local saving to S3Aggregator alongside unit tests


englehardt:
Both data aggregators are currently without unit tests, however the SqliteAggregator is at least tested through our integration tests. The `S3Aggregator` is never used in our test suite, which has led to a number of bugs:  https://github.com/mozilla/OpenWPM/issues/286, c01971088a21a47348ce8b861e2ee252ef9e60f6, among others.

Ideally we could do end-to-end tests that write and read from S3, but that would be a bit of a pain to set up. Instead, I think we should add in a config option to save parquet locally and use that to write a set of unit tests for the aggregator. This is a relatively easy change, and was a feature request from some folks at Mozilla. We'll also probably want to rename the Aggregator to ParquetAggregator for clarity.

motin:
To clarify, the support for locally testing S3 using mock implementation was landed in https://github.com/mozilla/OpenWPM/pull/333, but this issue (#294) is about adding a config option to save parquet locally and use that to write a set of unit tests for the aggregator, without any involvement of S3. 

@englehardt, to clarify this in the code, we could consider adding ParquetAggregator, with S3ParquetAggregator extending it.

englehardt:
> @englehardt, to clarify this in the code, we could consider adding ParquetAggregator, with S3ParquetAggregator extending it.

@motin That's an option. I actually just renamed the class to `ParquetAggregator`. See my WIP PR #390 and note the issues I identified while working on it. That's why I decided to stop working on it for the time being.

motin:
Related: Using Mock S3 to run local crawls with crawl data stored in parquet format: https://github.com/mozilla/openwpm-crawler/tree/master/deployment/local

englehardt:
I'm going to close this as I don't think it's the right direction anymore. The S3Aggregator should be optimized for cloud crawls and minimizing data loss. Constraining the future development there to also support an alternate local format doesn't seem right.

If we do want something like this, adding a new aggregator is the right approach.

Issue #247, opened 2018-12-15
Saving page content as individual gzip files causes leads to the "small files problem" during analysis


englehardt:
Basically, the cost of reading a bunch of small files from S3 is much too high (see: http://garrens.com/blog/2017/11/04/big-data-spark-and-its-small-files-problem/).

We should think about how we can batch the saving of content or apply a post-crawl batching of the individual files. Ideally, we'd preserve our ability to relatively easily retrieve the content for random individual files.

englehardt:
I've updated the title to reflect that our current way of saving these files leads to issues while crawling + while analyzing the data. See https://github.com/mozilla/openwpm-crawler/issues/33 for a description of the issues while crawling.

We should consider both of these constraints when we investigate an alternate way to handle saved files.

vringar:
Single Document (html, js, jpg...)saved and looked up randomly by key.

englehardt:
May be able to use a key-value store like DynamoDB from Amazon (or whatever the equivalent is on GCP)?

vringar:
GCP equivalents seem to be [BigTable](https://cloud.google.com/bigtable/) and [FireStore](https://cloud.google.com/firestore/)

PR #390, opened 2019-07-03
[WIP] Convert S3Aggregator to ParquetAggregator

englehardt:
See #294 for motivation. I've stopped working on this because I think it's going to be more complex than I expected. While it's true that we can easily save the parquet database locally, we need to handle the de-duplicated content somehow. We currently write content to a `content` directory on S3 rather than as a table in the database. This lets us de-duplicate across independent crawls (but leads to #247). So how do we save these when saving locally? Also as individual files? But we know that's super inefficient. The current `LocalAggregator` uses a LevelDB database to store the files. This is very performant, but I didn't want to duplicate that logic in the ParquetAggregator class. So we'll need to do some refactoring or rethink how we save content files.

### DIFF
--- a/automation/DataAggregator/ParquetAggregator.py
+++ b/automation/DataAggregator/ParquetAggregator.py
@@ -341,7 +341,7 @@ class ParquetAggregator(BaseAggregator):
         else:
             raise ValueError(
                 "The ParquetAggregator supports saving data to the local disk "
-                "(`local`) or to Amazon S3 (`s3`). You set "
+                "(`local_parquet`) or to Amazon S3 (`s3`). You set "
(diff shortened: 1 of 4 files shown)


englehardt:
Also note that this patch has some debugging code for issues that have since been fixed in master (i.e., encoding issues with the `visit_id` column, among others). So it needs to be rebased and cleaned up.

motin:
Related: A demo of how to setup local parquet-saving crawls using localstack: https://github.com/motin/OpenWPM/pull/1

englehardt:
Closing since I don't have immediate plans to pick this back up. Feel free to open a new PR.

PR #399, opened 2019-07-11
Update default_browser_params to include "navigation_instrument" 

aliamcami:
Added "navigation_instrument" parameter to default_browser_params.json file


motin:
Thanks @aliamcami, indeed it looks like we forgot this after adding that instrument :)

PR #401, opened 2019-07-13
Openwpm crawler support

motin:
This merges efforts from @englehardt and me in order to make OpenWPM support containerized crawls via https://github.com/mozilla/openwpm-crawler


crawler.py
```diff
@@ -0,0 +1,81 @@
...
+from automation import CommandSequence, TaskManager
+from automation.utilities import rediswq
+from test.utilities import LocalS3Session, local_s3_bucket
+
+# Configuration via environment variables
+NUM_BROWSERS = int(os.getenv('NUM_BROWSERS', '1'))
+REDIS_QUEUE_NAME = os.getenv('REDIS_QUEUE_NAME', 'crawl-queue')
+CRAWL_DIRECTORY = os.getenv('CRAWL_DIRECTORY', 'crawl-data')
+S3_BUCKET = os.getenv('S3_BUCKET', 'openwpm-crawls')
+HTTP_INSTRUMENT = os.getenv('HTTP_INSTRUMENT', '1') == '1'
```
englehardt: I'm not sure if it makes sense to pass this level of configuration in the yaml file. My thought for the general workflow:

1. A user closes the repo

motin: TLDR: Keeping high-level configuration allows for crawls to be made without being familiar with crawl engineering.

Long version:
That sounds like the advanced/customized workflow to me, in which maintaining a separate crawler.py file makes a lot of sense (simple Dockerfile: `FROM openwpm; ADD crawler.py`, then push), but I imagine the first-timer/general crawl workflow be something like:

1. They launch a simple crawl with their desired configuration (or the default configuration) via kubernetes (either locally or on GCP) and specify the set of sites the enqueue in redis and the name for the crawl.

Ie, using the stock openwpm image, which has been pushed to the docker registry by CI.

After performing one or many such crawls (for testing or production purposes), it may be of interest to tweak one or many aspects of the crawl strategy, at which point it makes sense going through the advanced/customized workflow above. 

I am even open to including configuration options such as `CRAWL_SEED_LIST="alexatop1k"`, `CRAWL_STRATEGY="stateless"`, `LINK_FOLLOWING_STRATEGY="random", `LINK_FOLLOWING_DEPTH="1"` etc that would allow new crawls and replication of previous studies / crawls with even more ease.

Dockerfile
```diff
@@ -59,4 +59,4 @@ COPY . .
 RUN adduser --disabled-password --gecos "OpenWPM"  openwpm
 
 # Alternatively, python3 could be used here
-CMD python demo.py
+CMD python crawler.py
```
englehardt: Do we want to change this to `crawler.py` in the main repo? Given that this script requires a bunch of environment variables to be set to run correctly, it seems like we should keep `demo.py` as the default?

motin: Sure, we can add the crawler.py command to the crawl yaml files

automation/DataAggregator/S3Aggregator.py
```diff
@@ -63,6 +63,7 @@ def __init__(
         self._batches = dict()  # maps table_name to a list of batches
         self._instance_id = instance_id
         self._bucket = manager_params['s3_bucket']
+        self._s3_content_cache = set()  # cache of filenames already uploaded
```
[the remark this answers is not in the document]
motin: Sure thing :)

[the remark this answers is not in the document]
motin: Done: https://github.com/mozilla/OpenWPM/pull/402

Dockerfile
motin: Done

crawler.py
englehardt: Makes sense; thanks for the response.

PR #414, opened 2019-07-20
Periodically drain the s3 aggregator queue

motin:
To avoid data loss in case of stalled processes.
Works around #413

### DIFF
--- a/automation/DataAggregator/S3Aggregator.py
+++ b/automation/DataAggregator/S3Aggregator.py
@@ -4,6 +4,7 @@ import base64
 import gzip
 import hashlib
 import json
+import time
 import uuid
 from collections import defaultdict
 
@@ -23,6 +24,7 @@ from .parquet_schema import PQ_SCHEMAS
 CACHE_SIZE = 500
 SITE_VISITS_INDEX = '_site_visits_index'
 CONTENT_DIRECTORY = 'content'
+QUEUE_DRAIN_INTERVAL = 15  # seconds
 
 
 def listener_process_runner(
@@ -34,6 +36,7 @@ def listener_process_runner(
 
     while True:
         listener.update_status_queue()
+        listener.periodically_drain_queue()
         if listener.should_shutdown():
             break
         try:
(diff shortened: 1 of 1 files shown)


motin:
@englehardt Could you finish this PR (or equivalent to work around #413)? I see that this attempt of a solution currently at least once has lead to duplicate data (see https://travis-ci.org/mozilla/OpenWPM/jobs/561646341), and when I ran it locally I got:

```
Process Process-2:
  File "/usr/local/lib/python2.7/dist-packages/multiprocess/process.py", line 267, in _bootstrap
    self.run()
  File "/usr/local/lib/python2.7/dist-packages/multiprocess/process.py", line 114, in run
    self._target(*self._args, **self._kwargs)
  File "/opt/OpenWPM/automation/DataAggregator/S3Aggregator.py", line 39, in listener_process_runner
    listener.periodically_drain_queue()
  File "/opt/OpenWPM/automation/DataAggregator/S3Aggregator.py", line 222, in periodically_drain_queue
    self.drain_queue()
  File "/opt/OpenWPM/automation/DataAggregator/S3Aggregator.py", line 291, in drain_queue
    self._send_to_s3(force=True)
  File "/opt/OpenWPM/automation/DataAggregator/S3Aggregator.py", line 199, in _send_to_s3
    table = pa.Table.from_batches(batches)
  File "pyarrow/table.pxi", line 1315, in pyarrow.lib.Table.from_batches
    raise ValueError('Must pass schema, or at least '
ValueError: Must pass schema, or at least one RecordBatch
```

I believe that a period drain of the write queue is a good enough solution for now, but I am not sure of how to implement this properly.

englehardt:
This should work, but does mean we will lose some of the benefit of batching. Note that parquet datasets can be quite slow when the individual parquet files are small. Perhaps we should look into some one-time batching step to re-partition the dataset into option chunks after the crawl is finished?

Either way, we should implement this so our crawls actually complete.

englehardt:
I'm happy to do it, but I'm going to take a slightly different approach to preserve the batching. If there have been 30 seconds without a new record added to the batch we write out the current batch.

motin:
Closing in favor of https://github.com/mozilla/OpenWPM/pull/420

Issue #255, opened 2019-02-07
Investigate high crash rate of Firefox 60 crawls


englehardt:
In a 1 million site crawl with our most recent Firefox 60 support has a relatively high command crash rate (13% of `get` commands). Compare this to 3.5% for the current Firefox 52-based master branch.

A (partial) log file from a recent 1M site crawl is available here: https://www.dropbox.com/s/320d4my3b2prnc8/2019_01_29_alexa1m_logs.bz2?dl=0

motin:
> 13% of `get` commands

How is this number calculated? Do we store information about crashes in the resulting crawl datasets?

It seems like the right thing to do here is to set up [Sentry](https://sentry.io/) or similar solution so that patterns in these errors can be properly detected and troubleshot.

englehardt:
> How is this number calculated? Do we store information about crashes in the resulting crawl datasets?

Yes, see https://dbc-caf9527b-e073.cloud.databricks.com/#notebook/71922/command/71923

gunesacar:
> /cc @gunesacar, who also has some data on this comparing the 1M site crawls we used to run on FF52 and a crawl with the new webextensions instrumentation.

Yes, we got 4x more GET command failures with the webextensions instrumentation compared to the previous (November 2018) crawl. See the 7th figure from the top on this notebook: https://github.com/citp/openwpm-data-release/blob/master/Crawl-Data-Metrics.ipynb

Happy to share the crawl logs and the database. You can also download them as part of the [Princeton Web Census Dataset](https://webtransparency.cs.princeton.edu/webcensus/data-release/)).

motin:
> Yes, we got 4x more GET command failures with the webextensions instrumentation compared to the previous (November 2018) crawl. See the 7th figure from the top on this notebook: https://github.com/citp/openwpm-data-release/blob/master/Crawl-Data-Metrics.ipynb

This is such a great notebook. Thanks for sharing this! I am looking forward to getting Sentry setup properly (#406) so that we systematically can address this.

motin:
> > How is this number calculated? Do we store information about crashes in the resulting crawl datasets?
> 
> Yes, see https://dbc-caf9527b-e073.cloud.databricks.com/#notebook/71922/command/71923

Percentage of command failures 12.26%
Percentage of command timeouts 5.61%

motin:
[...]

PR #420, opened 2019-07-23
Add timeout to batch upload in S3Aggregator

### LINKED ISSUE #413: Crawls performed in GCP end up with no visits data saved to S3
Could be due to some bug related to high parallelization. Needs investigation.

englehardt:
Alternate to #414.
Fixes #413


englehardt:
I ran a couple gcp test crawls with this fix and data is now reliably written to GCP. See https://dbc-caf9527b-e073.cloud.databricks.com/#notebook/146267/command/146283 as an example.

Note that the total number of commands submitted is still suspiciously low. I suspect this is due to TaskManager-level crashes as documented off of #255.

Issue #417, opened 2019-07-22
Move configuration files into their own directory on S3


englehardt:
We currently write config files to the root directory. We should instead write them into `config/`

motin:
Fixed in https://github.com/mozilla/OpenWPM/pull/421

PR #421, opened 2019-07-23
Move config files into subdirectory

englehardt:
This moves the config files on S3 into a subdirectory. Since there are typically hundreds of files, this will keep the root directory much cleaner.


englehardt:
This includes changes from #420

englehardt:
Will fix #417

PR #429, opened 2019-07-30
[WIP] Conda install openwpm dependencies

birdsarah:
Task: make dependencies conda installable
Why: so we can get going on windows
Side benefit: installation for all platforms will be simplified and unified

Depends on #426

PR #430, opened 2019-07-30
[WIP] Windows hackathon

birdsarah:
@crankycoder - this should run locally if you follow the README and the notes in demo.py

Note you have to get firefox and openwpm.xpi yourself.

* openwpm.log ends up with some stuff in it
* the sqlite database ends up with some stuff in tables that are not populated by the web extension

Earlier today I thought I knew why the table was not getting populated but I don't. 

The plyvel swap for python-leveldb is on #429 and you can test that it works and is populating fine on linux (I haven't tried on osx yet so if you want to give that a walk through on osx that's cool).

So, it's not that and I would hazard a guess that it's something to do with sockets. If you wouldn't mind starting there - getting data into the sqlite database OR getting it into parquet with S3, it doesn't really matter, but I would hunch that they both don't currently work.

This may all be late night ramblings so hit me up in the morning for some clarity.

Also note your headless mode pr is working on windows too :D 

Note would recommend not running in headless mode because my pid code does not work properly so this will help to remind you. Because pid code doesn't work properly you'll have a bunch of firefox and geckodriver processes lurking around that need manually killing.


englehardt:
Closing due to inactivity

Issue #406, opened 2019-07-16
Configure Sentry for error/crash report aggregation


motin:
This supports resolving https://github.com/mozilla/OpenWPM/issues/255

Docs: https://sentry.io/for/python/, https://github.com/getsentry/sentry-python

motin:
With PR #415, activation of Sentry works fine (tried in a GCP crawl) and uncaught exceptions are caught. This at least allows us to get some notification on uncaught exceptions in a crawl.

In an additional PR, I hope that we can also report the caught exceptions to sentry so that we can get a good overview of what is causing driver-related crashes in the crawls. 

Also, integrating with the OpenWPM logger (eg sending log messages to Sentry as well as per https://docs.sentry.io/platforms/python/logging/#usage) allows accessing the relevant log scrollback for each reported exception.

@englehardt Care to set that up? I am not sure where to best add that code.

motin:
Possibly useful for crawls as well, adding Redis commands as breadcrumbs: https://docs.sentry.io/platforms/python/redis/

PR #432, opened 2019-07-30
[WIP] Add sentry logging to MPLogger

englehardt:
This allows Sentry to capture the full set of logs that are normally only sent to the local log file.

### DIFF
--- a/automation/MPLogger.py
+++ b/automation/MPLogger.py
@@ -9,6 +9,8 @@ import struct
 import sys
 import time
 
+import sentry_sdk
+from sentry_sdk.integrations.logging import LoggingIntegration
 from six.moves.queue import Empty as EmptyQueue
 
 from .SocketInterface import serversocket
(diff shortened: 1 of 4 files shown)


automation/MPLogger.py
```diff
@@ -67,14 +69,85 @@ def loggingserver(log_file, status_queue):
...
+
+    # If running a cloud crawl, we can pull the sentry endpoint and related
+    # config varibles from the environment
+    SENTRY_DSN = os.getenv('SENTRY_DSN', None)
+    if SENTRY_DSN:
+        sentry_logging = LoggingIntegration(
+            level=logging.DEBUG,
+            event_level=logging.ERROR
+        )
+        sentry_sdk.init(
```
motin: I am not sure that having multiple sentry_sdk.init is a good idea. Not sure what the alternative is but init is already in https://github.com/mozilla/OpenWPM/blob/a337b52eb172ceed9e2ab092a4ba0e1f7421ad29/crawler.py#L30

englehardt: I believe this is required since `crawler.py` and `MPLogger.py` execute in different processes.

motin: Removing the crawler.py code sounds like a good idea to me

englehardt: For some background, MP logger was written to serialize writes from multiple processes (the task manager, each browser manager, and each extension). See https://mattgathu.github.io/multiprocessing-logging-in-python/ for a description of the problem and an alternate solution (using queues instead of sockets). So by initializing and attaching sentry to MPLogger's logging module, we should capture logs from all of those processes -- basically everything that would normally be written to the log file on disk. However, any uncaught exceptions in the main process (i.e., the process that runs the `crawler.py` crawl script and the TaskManager) will not be captured by MPLogger.

englehardt: Thus, by initializing in both we should capture exceptions in both.

motin: Ok great to know. Initializing in both seems fine, but we should move the init code into a separate file and avoid dry.

englehardt: Makes sense! I also think it might make sense to just run the logging server within a thread of the main process -- it doesn't really require a new process. Then only one initialization should be fine (if I understand correctly how the logger is scoped).

motin: Feel free to move that into the same process in a separate PR - or we could consider using fluentd or similar which would allow us to configure persistent logging for containerized deployments - but ultimately out of scope for this particular PR. :)

motin:
This is related to https://github.com/mozilla/OpenWPM/issues/406

motin:
Closing in favor of #434

PR #434, opened 2019-07-31
Refactor MPLogger to make cross-process sentry logging work

### LINKED ISSUE #406: Configure Sentry for error/crash report aggregation

englehardt:
This is an alternate approach to #432. MPLogger's current structure is quite broken (in that it requires child processes to explicitly connect to the central logging processing) and doesn't play well with Sentry. This refactor allows us to serialize logs from all processes in a single sentry report. The only remaining issue is that it [doesn't appear to be possible](https://github.com/getsentry/sentry-python/issues/291) to capture uncaught exceptions of child processes in the new Sentry SDK.

Fixes #406


automation/MPLogger.py
```diff
@@ -5,13 +5,21 @@
...
+import threading
 import time
 
+import sentry_sdk
+from sentry_sdk.integrations.logging import BreadcrumbHandler, EventHandler
 from six.moves.queue import Empty as EmptyQueue
 
-from .SocketInterface import serversocket
+# from .SocketInterface import serversocket
+from SocketInterface import serversocket
```
motin: This change led to `E   ImportError: No module named SocketInterface`

motin:
This is related to https://github.com/mozilla/OpenWPM/issues/406

motin:
@englehardt This looks promising, and I fixed the obvious issues reported in CI, but I'll leave it to you to finish the work since it requires a good sense of how to change the rest of the code base to properly interact with the refactored logger class.

englehardt:
This is in pretty good shape for review. This PR:

* Refactors how loggers are attached within the parent and child processes. This is much cleaner than the previous instrumentation, and issues like https://github.com/mozilla/OpenWPM/pull/260/files won't require passing around a reference to the logging object. All you need to do is attach to the 'openwpm' logger (via `logging.getLogger('openwpm')`) and the logs will be written to console, disk, and sentry (if enabled). This also means that the sentry breadcrumbs will be consistent with the other logfiles (i.e., will include logs across processes).
* Only initializes sentry once, meaning we don't need to replicate the config across processes
* Wraps child processes in a try-catch, and logs any resulting exception and traceback to all loggers (including sentry). These were previously only printed to stderr.
* Parses sentry log errors to group by message and error value. This removes a lot of the redundant logs that were grouped by browser ID.

There are a couple remaining issues:
* There's a race condition in the initialization of MPLogger that occasionally causes it hang. I think I know why, and it should be a pretty simple fix.
* We need better formatting within `_sentry_before_send` and at the two points we add tracebacks to the `extra` data to log calls to fix two issues.
    * https://sentry.prod.mozaws.net/operations/web-crawls/issues/6110000/
    * https://sentry.prod.mozaws.net/operations/web-crawls/issues/6110002/

One other thing to note that I don't understand: Sentry [claims to support the `exc_info` logging parameter](https://docs.sentry.io/platforms/python/logging/#usage), which automatically attaches traceback info to the logs. This seems to work just fine for console and file logging, but the sentry logs do not include traceback information irrespective of this parameter. As a workaround I've attached the traceback text in the `extra` parameter. This is better than including the info in the event message itself, as event messages are truncated. However, it appears that there is still some truncation (despite claims of a 16kb limit) depending on the encoding. I'll investigate more tomorrow as I address the two issues above.

englehardt:
https://docs.sentry.io/clients/java/config/#max-message-size may help with truncated messages.

englehardt:
Looking through https://github.com/getsentry/sentry-python/pull/352 and https://github.com/getsentry/sentry-python/issues/344, it seems there currently isn't a way to avoid the string limits imposed by Sentry without monkey patching the library. I've chosen instead to chunk the traceback string into multiple "additional data" keys.

Here's [an example](https://sentry.prod.mozaws.net/operations/web-crawls/issues/6109868/). The formatting isn't great in the [breadcrumbs](https://sentry.prod.mozaws.net/operations/web-crawls/issues/6109868/#breadcrumbs) section, but thankfully the [additional data](https://sentry.prod.mozaws.net/operations/web-crawls/issues/6109868/#extra) section sorts the data by key, so the traceback is easy to read.

englehardt:
@motin This PR is ready to be reviewed and merged. I've fixed the issues identified in https://github.com/mozilla/OpenWPM/pull/434#issuecomment-517569977.

Here's an example of logs with the new formatting:
* https://sentry.prod.mozaws.net/operations/web-crawls/issues/6111952/
* https://sentry.prod.mozaws.net/operations/web-crawls/issues/6111967/

I suggest we file a follow-up issue to further improve the messages for these errors. We can strip out the "Crash in driver..." and "Child process exception" text and perhaps strip individual URLs from the neterror messages (or at least update the fingerprint for those events) to allow all neterror messages of a certain type to group together. All of this can be done in `MPLogger::_sentry_before_send`.

englehardt:
> I suggest we file a follow-up issue to further improve the messages for these errors. We can strip out the "Crash in driver..." and "Child process exception" text and perhaps strip individual URLs from the neterror messages (or at least update the fingerprint for those events) to allow all neterror messages of a certain type to group together. All of this can be done in MPLogger::_sentry_before_send.

I decided to do that here, the latest commits include these changes. see https://sentry.prod.mozaws.net/operations/web-crawls/?query=CRAWL_REFERENCE%3A%22openwpm-crawls%2F2019-08-02_test_pr_438_2%22 as an example.

PR #436, opened 2019-08-01
Configurable Redis host

motin:
Allows for use of Google Cloud Memorystore for Redis in GCP for instance, as implemented in https://github.com/mozilla/openwpm-crawler/pull/20


englehardt:
Thanks!

Issue #439, opened 2019-08-05
localstack dependency import issues


englehardt:
The latest release of localstack (0.10.X) appears to have some dependency install issues. Specifically, it appears to fail to install any of the required dependent packages.

For examples, see:
* https://travis-ci.org/mozilla/OpenWPM/jobs/567950566#L2225
* https://travis-ci.org/mozilla/OpenWPM/jobs/568040169#L2236

Pinning to an older version (0.9.6) seems to fix the issue (https://github.com/mozilla/OpenWPM/pull/438/commits/413ce79c2f1075baac90cc0f44d0cdb91eedcf0a)

motin:
> Pinning to an older version (0.9.6) seems to fix the issue

Sounds like a good solution for now.

englehardt:
Closed by #440

PR #438, opened 2019-08-02
Fix S3Aggregator crash when records are flushed twice

englehardt:
Will rebase once #434 is merged into master. See the commit message for https://github.com/mozilla/OpenWPM/commit/4a0e4e4ede9120c28cabe2c35fef0487f9835220 for a description of the issue.

This should fix errors of this type: https://sentry.prod.mozaws.net/operations/web-crawls/issues/6111973/, which are likely lead to the socket errors we're seeing here: https://sentry.prod.mozaws.net/operations/web-crawls/issues/6101037/.

PR #440, opened 2019-08-05
Pin localstack to older version to fix dependency issues

englehardt:
Broken out from #438.

Localstack released 10.X on August 3rd. The newest version has installation issues. See #439. Pinning to `0.9.6` then introduced a bunch of `flake8` errors that we didn't have before. This is caused by the fact that localstack includes `flake8-quotes` as a dependency (https://travis-ci.org/mozilla/OpenWPM/jobs/568044495#L2129-L2130) and this plugin flags any use of double quotes as an error. Unfortunately flake8 automatically loads any installed plugins and doesn't provide a way to disable an installed plugin or restrict the set of possible plugins (https://gitlab.com/pycqa/flake8/issues/133). Thus, I've decided to disable the entire `QXXX` error code range, which as of 2 years ago isn't used by any plugins we plan to install (https://gitlab.com/pycqa/flake8/issues/339).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,11 +25,10 @@ before_install:
   - "export BOTO_CONFIG=/dev/null" # https://github.com/travis-ci/travis-ci/issues/7940
 install:
   - echo "y" | ./install.sh
-  - pip install -r requirements.txt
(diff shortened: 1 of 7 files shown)


englehardt:
Fixes #439

PR #441, opened 2019-08-05
[WIP] Replace print calls with logging

englehardt:
This is especially important in `SocketInterface.py`, which is currently `print`ing important serialization errors.


englehardt:
Will rebase on #440 once merged.

englehardt:
We should fix #258 as part of this

englehardt:
Closing this since I never managed to get the tests passing.

PR #445, opened 2019-08-06
Add specific webdriver exception reporting

birdsarah:
Specifically capture Selenium Exceptions and report.

@englehardt this is the other "fix" from my frankenstein branch. You already did the s3 bug in #438.

I don't know whether this should be logged as "error" or put in status queue as "failed".

Am more than happy for you to copy paste and make a PR that is in line with your better judgement on how to report.

### DIFF
--- a/automation/BrowserManager.py
+++ b/automation/BrowserManager.py
@@ -427,9 +427,9 @@ def BrowserManager(command_queue, status_queue, browser_params,
         err_info = sys.exc_info()
         status_queue.put(('CRITICAL', pickle.dumps(err_info)))
         return
-	except WebDriverException as e:
+    except WebDriverException as e:
         logging.error("BROWSER %i: Selenium WebDriver Exception: %s" % (
-             browser_params['crawl_id'], str(e))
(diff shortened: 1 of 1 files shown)


birdsarah:
Thanks @motin.

englehardt:
@birdsarah I don't think this is necessary anymore.  In https://github.com/mozilla/OpenWPM/pull/434 I starting [removing the `BROWSER X:` prefix](https://github.com/mozilla/OpenWPM/blob/e9c1baa0582a6ede58e0230d5c947952cbf607a6/automation/MPLogger.py#L115-L119) from all logs sent to sentry. As an example, take a look at the logs for your recent crawl:  https://sentry.prod.mozaws.net/operations/web-crawls/?query=CRAWL_REFERENCE%3A%22srg-team-bucket%2F2019-07-jestr-vs-crawl%2F2019_08_09_main_crawl_7a%22. The webdriver exceptions are nicely batched by type, which I think will make the errors easier to tackle.

If you disagree let's reopen.

PR #442, opened 2019-08-06
Encode post bodies as utf-8 before sending

englehardt:
Post bodies that contain non-ascii characters are currently discarded by the SocketInterface during deserialization.

PR #447, opened 2019-08-06
[WIP] Add non-ascii characters to post body tests

englehardt:
Creating this to show that the current master branch fails to parse post bodies that contain non-ascii characters. Will fold into #442.


englehardt:
Closing since these tests have been folded into #442.

Issue #12, opened 2015-04-20
Close browser gracefully when possible


englehardt:
It's definitely not best practice to kill everything indiscriminately. We should attempt to first close each child process and kill if the close operation times out.

Issue #404, opened 2019-07-15
Alerts causing `WebDriverException: Message: Failed to decode response from marionette`?


motin:
Not sure what this one is about, apparently something related to reading the value of alerts (`self.driver.execute(Command.W3C_GET_ALERT_TEXT)["value"]`) causing this crash?

Log output:
```
BrowserManager       - INFO     - BROWSER -1312820091: EXECUTING COMMAND: ('GET', u'http://ltn.com.tw', 10, -1446733757290209L)
BrowserManager       - INFO     - BROWSER -1312820091: Crash in driver, restarting browser manager 
 Traceback (most recent call last):
  File "/opt/OpenWPM/automation/BrowserManager.py", line 420, in BrowserManager
    browser_params, manager_params, extension_socket)
  File "/opt/OpenWPM/automation/Commands/command_executor.py", line 15, in execute_command
    extension_socket=extension_socket)
  File "/opt/OpenWPM/automation/Commands/browser_commands.py", line 131, in get_website
    WebDriverWait(webdriver, .5).until(EC.alert_is_present())
  File "/usr/local/lib/python2.7/dist-packages/selenium/webdriver/support/wait.py", line 71, in until
    value = method(self._driver)
  File "/usr/local/lib/python2.7/dist-packages/selenium/webdriver/support/expected_conditions.py", line 401, in __call__
    alert = driver.switch_to.alert
  File "/usr/local/lib/python2.7/dist-packages/selenium/webdriver/remote/switch_to.py", line 55, in alert
    alert.text
  File "/usr/local/lib/python2.7/dist-packages/selenium/webdriver/common/alert.py", line 67, in text
    return self.driver.execute(Command.W3C_GET_ALERT_TEXT)["value"]
  File "/usr/local/lib/python2.7/dist-packages/selenium/webdriver/remote/webdriver.py", line 321, in execute
    self.error_handler.check_response(response)
  File "/usr/local/lib/python2.7/dist-packages/selenium/webdriver/remote/errorhandler.py", line 242, in check_response
    raise exception_class(message, screen, stacktrace)
WebDriverException: Message: Failed to decode response from marionette
TaskManager          - INFO     - BROWSER -1312820091: Received failure status while executing command: GET
```

englehardt:
Related to #255. We should take note of all of these types errors in the WebExtension crawls and fix where possible. The current crash rate is quite high.

englehardt:
[...]

PR #449, opened 2019-08-10
Improved stability and error logging

englehardt:
This is addressing some of the errors observed in a recent large crawl.

### DIFF
--- a/automation/Commands/browser_commands.py
+++ b/automation/Commands/browser_commands.py
@@ -133,7 +133,7 @@ def get_website(url, sleep, visit_id, webdriver,
         alert = webdriver.switch_to_alert()
         alert.dismiss()
         time.sleep(1)
-    except TimeoutException:
+    except (TimeoutException, WebDriverException):
         pass
 
     close_other_windows(webdriver)
--- a/automation/BrowserManager.py
+++ b/automation/BrowserManager.py
@@ -6,6 +6,7 @@ import os
 import shutil
 import signal
 import sys
+import threading
 import time
 import traceback
 
@@ -107,6 +108,9 @@ class Browser:
             crash_recovery = True
         else:
         """
+        self.logger.info(
+            "BROWSER %i: Launching browser..." % self.crawl_id
+        )
         tempdir = None
         crash_recovery = False
         self.is_fresh = not crash_recovery
@@ -174,7 +178,7 @@ class Browser:
                 self.logger.error(
                     "BROWSER %i: Spawn unsuccessful %s" % (self.crawl_id,
                                                            error_string))
-                self.kill_browser_manager()
+                self.close_browser_manager()
                 if 'Profile Created' in launch_status:
                     shutil.rmtree(spawned_profile_path, ignore_errors=True)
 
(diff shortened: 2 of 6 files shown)


englehardt:
@motin This PR is ready for review. I ran a 10k site crawl with this branch and here are the improvements:

First, some positives:
1. Out of the 10k sites, 9489 are in the `crawl_history` table. This means that the aggregator is much more stable than our previous crawls (with >30% data loss).
2. The new serialization of exceptions means the sentry logs are much cleaner. Look at the groupings in https://sentry.prod.mozaws.net/operations/web-crawls/?query=CRAWL_REFERENCE%3A%22openwpm-crawls%2F2019-08-12_test_pr_449%22.
3. I didn't receive any threading related errors, perhaps as a result of more safely closing the webdriver (https://github.com/mozilla/OpenWPM/pull/449/commits/8903f9d287e949e716d07e9c227f461a84ef7277)?

Remaining problems:
1. There was still a single broken pipe error in the main process (https://sentry.prod.mozaws.net/operations/web-crawls/issues/6101037/?query=CRAWL_REFERENCE:%22openwpm-crawls/2019-08-12_test_pr_449%22). This means there's still a crasher in the aggregator, but there's no associated Sentry log from what i can tell. I'll need to do a bit more digging in the GCP logs (https://console.cloud.google.com/logs/viewer?organizationId=442341870013&project=senglehardt-openwpm-test-1&minLogLevel=500&expandAll=false&timestamp=2019-08-13T00:15:08.481000000Z&customFacets=&limitCustomFacetWidth=true&dateRangeStart=2019-08-12T00:15:08.856Z&dateRangeEnd=2019-08-13T00:15:08.856Z&interval=P1D&resource=container&scrollTimestamp=2019-08-13T00:00:30.414629184Z).
2. The rate of WebDriverExceptions is quite high. After ignoring the exceptions for the alert handling code (https://github.com/mozilla/OpenWPM/pull/449/commits/0b9ccb780685052a2a47d7dbd11bb91fa07c7050), we received almost no errors of the type `Message: Failed to decode response from marionette`. Instead, we received a bunch of `InvalidSessionIdException: Message: Tried to run command without establishing a connection`, which is just caused by the next command that interacts with the webdriver. Thus, there must be something that's causing the webdriver / marionette session to disconnect from the browser after it finishes loading the page. Perhaps the browser has crashed at that point? If I could find a site that reliably reproduces the error, it would be easier to debug.

englehardt:
Fixes #12. Partially addresses #450.

automation/Commands/browser_commands.py
```diff
@@ -133,7 +133,7 @@ def get_website(url, sleep, visit_id, webdriver,
         alert = webdriver.switch_to_alert()
         alert.dismiss()
         time.sleep(1)
-    except TimeoutException:
+    except (TimeoutException, WebDriverException):
```
motin: This will completely suppress the exceptions that are reported in https://github.com/mozilla/OpenWPM/issues/404 and prevent being able to properly address these errors as per https://github.com/mozilla/OpenWPM/issues/404#issuecomment-517879065. When addressing #404, we must remember to change this back to throwing exceptions, so that we can reproduce the root cause via tests, report upstream etc.

PR #456, opened 2019-08-14
Allow content saving to be configured by any resource type

### LINKED ISSUE #344: Allow content saving to be configured by any resource type
* **I'm submitting a ...**
[ ] bug report
[X] feature request
[ ] question about the decisions made in the repository
[ ] question about how to use this project

* **Summary**
Right now we have a `saveJavascript` and a `saveAllContent` setting. These should be removed in favor of an array where one can specify the resource types for which they want to have content saved. This array should accept any of the available types [listed here](https://github.com/mozilla/openwpm-webext-instrumentation/blob/1af4022730104b80c10be51573363342ca2dd7fc/src/background/http-instrument.ts#L42-L64).

For example, the config parameter could be:

`saveContent: ['script', 'main_frame', 'sub_frame']`

motin:
Fixes #344  

Possible ways to specify "save_content": 
* script,main_frame,sub_frame
* True
* False

But not:
* ['script', 'main_frame', 'sub_frame']
* '0'
* '1'

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,13 +96,16 @@ available [below](#output-format).
 * Response body content
     * Saves all files encountered during the crawl to a `LevelDB`
         database de-duplicated by the md5 hash of the content.
-    * Set `browser_params['save_all_content'] = True`
+    * Set `browser_params['save_content'] = True`
     * The `content_hash` column of the `http_responses` table contains the md5
         hash for each script, and can be used to do content lookups in the
         LevelDB content database.
     * NOTE: this instrumentation may lead to performance issues when a large
         number of browsers are in use.
-    * Set `browser_params['save_javascript'] = True` to save only Javascript
+    * Set `browser_params['save_content']` to a comma-separated list of 
+        [resource_types](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/webRequest/ResourceType)
+        to save only specific types of files, for instance 
+        `browser_params['save_content'] = "script"` to save only Javascript
         files. This will lessen the performance impact of this instrumentation
         when a large number of browsers are used in parallel.
 * Flash Cookies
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -96,13 +96,16 @@ available [below](#output-format).
...
         database de-duplicated by the md5 hash of the content.
-    * Set `browser_params['save_all_content'] = True`
+    * Set `browser_params['save_content'] = True`
     * The `content_hash` column of the `http_responses` table contains the md5
         hash for each script, and can be used to do content lookups in the
         LevelDB content database.
     * NOTE: this instrumentation may lead to performance issues when a large
         number of browsers are in use.
-    * Set `browser_params['save_javascript'] = True` to save only Javascript
+    * Set `browser_params['save_content']` to a comma-separated list of
```
englehardt: In your PR summary you mention that you can also initialize this with a list of strings. Should we document that here?

automation/Extension/webext-instrumentation/src/background/http-instrument.ts
```diff
@@ -177,6 +178,49 @@ export class HttpInstrument {
     }
   }
 
+  private shouldPossiblýSaveContent(saveContentOption: SaveContentOption) {
```
englehardt: I think this is better described as `isContentSavingEnabled`.

README.md
englehardt: If this is an acceptable config type, be sure to check for it in the misconfiguration checks described in my comment above. I think it's better to only support a single multi-option config format (either comma-separated string or a list of strings).

englehardt:
To test saving documents, you can probably reuse the test pages I added in this (unmerged) fork: https://github.com/englehardt/OpenWPM/commit/3ee26fa0fd3a8913a72a7e6665860fe1a569212c

README.md
motin: > I think it's better to only support a single multi-option config format (either comma-separated string or a list of strings).

Agree. That is simpler. Removed support for list of strings.

PR #458, opened 2019-08-14
Add logSettings.logFunctionGets

motin:
... for also logging access (gets) to instrumented function properties, not only their execution (calls).


motin:
> Thanks! What is the use case for this?

Sometimes scripts only check for the existence of a function `if (window.doThis)` in order to gain knowledge about the environment (browser support/version etc). This currently goes undetected in the current instrumentation. 

> It would be great to have a test for it as well, but I don't think it's a blocker since writing this test would require some piece of built-in instrumentation to use it.

Yes, once we get there we'll add tests :)

PR #459, opened 2019-08-14
Js instrument retain prototype and constructor

### LINKED ISSUE #461: Instrumenting parent object property breaks instrumentation of object properties
When instrumenting both window.Foo.bar AND window.Foo itself, the instrumentation of the latter breaks the instrumentation for class instances from `new window.Foo()`. 

Steps to reproduce:
1. Instrument window.Foo.bar
2. Instrument window.Foo
3. Execute js: `const foo = new window.Foo();`
4. One JS instrumentation call event for `window.Foo`
5. Execute js: `foo.bar;`

Expected results:
6. One JS instrumentation get event for `foo.bar`

Actual results:
6. No JS instrumentation get event for `foo.bar`

motin:
Restore the original prototype and constructor so that instrumented classes remain intact.

Fixes #461

This PR currently also includes https://github.com/mozilla/OpenWPM/pull/458, will be rebased after its merge.

### DIFF
--- a/automation/Extension/webext-instrumentation/src/lib/js-instruments.ts
+++ b/automation/Extension/webext-instrumentation/src/lib/js-instruments.ts
@@ -609,12 +609,21 @@ export function jsInstruments(event_id, sendMessagesToLogger) {
                 logSettings,
               );
             }
-            return instrumentFunction(
+            const instrumentedFunctionWrapper = instrumentFunction(
               objectName,
               propertyName,
               origProperty,
               logSettings,
             );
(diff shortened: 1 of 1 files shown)


motin:
> This PR currently also includes #458, will be rebased after its merge.

Rebased and ready for merge.

motin:
> Did you find some breakage that was caused by this?

Yes, retroactively filed https://github.com/mozilla/OpenWPM/issues/461 for clarity :)

PR #475, opened 2019-08-26
Check for Firefox process crashes explicitly

englehardt:
In #255 we saw a lot of unclear error messages caused by the Firefox process crashing after the initial `get()` command. In this PR we check the webdriver's `current_url` attribute, which will cause marionette to query the Firefox process. If the process is dead or stalled, or if the marionette connection was severed for some reason, we will see a crash from this call and raise a more helpful error message.

### DIFF
--- a/automation/Commands/browser_commands.py
+++ b/automation/Commands/browser_commands.py
@@ -19,6 +19,7 @@ from selenium.webdriver.support import expected_conditions as EC
 from selenium.webdriver.support.ui import WebDriverWait
 from six.moves import range
 
+from ..Errors import BrowserCrashError
 from ..SocketInterface import clientsocket
 from .utils.lso import get_flash_cookies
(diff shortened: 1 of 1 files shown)


automation/Commands/browser_commands.py
```diff
@@ -127,13 +128,22 @@ def get_website(url, sleep, visit_id, webdriver,
     # Sleep after get returns
     time.sleep(sleep)
 
+    # Check if the browser crashed
+    try:
+        webdriver.current_url
+    except Exception:
```
motin: This seems a too generous catch. Is the exception thrown here of class Exception or are there specific exceptions that we want to catch? Maybe sentry-log the Exception so that its details show up as breadcrumbs?

englehardt:
I think this could end up hiding issues that we want to learn about. I'm going to leave this out for now.

PR #483, opened 2019-08-30
Avoid TypeError: logSettings.nonExistingPropertiesToInstrument is undefined

motin:
Found this error during manual testing. Introduced in #452. Happens when attempting to instrument a non-existing property without using the nonExistingPropertiesToInstrument config attribute.

### DIFF
--- a/automation/Extension/webext-instrumentation/src/lib/js-instruments.ts
+++ b/automation/Extension/webext-instrumentation/src/lib/js-instruments.ts
@@ -589,9 +589,13 @@ export function jsInstruments(event_id, sendMessagesToLogger) {
     // Store original descriptor in closure
     const propDesc = Object.getPropertyDescriptor(object, propertyName);
 
+    // Property descriptor must exist unless we are instrumenting a
+    // non-existing property
     if (
       !propDesc &&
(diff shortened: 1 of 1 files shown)


automation/Extension/webext-instrumentation/src/lib/js-instruments.ts
```diff
@@ -589,9 +589,13 @@ export function jsInstruments(event_id, sendMessagesToLogger) {
     // Store original descriptor in closure
     const propDesc = Object.getPropertyDescriptor(object, propertyName);
 
+    // Property descriptor must exist unless we are instrumenting a
+    // non-existing property
     if (
       !propDesc &&
-      logSettings.nonExistingPropertiesToInstrument.indexOf(propertyName) == -1
+      (!logSettings.nonExistingPropertiesToInstrument ||
+        logSettings.nonExistingPropertiesToInstrument.indexOf(propertyName) ==
```
englehardt: Why not use `includes()` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes? It might save us from wrapping text here :)

PR #488, opened 2019-09-11
Add page visit test including a worker script making a request

jallmann:
This patch adds a new testcase for the http instrumentation. 
A page is loaded that includes a worker script. The page and the script both load the same image, resulting in different requests with differen top-level-urls. 
As mentioned [here](https://github.com/mozilla/OpenWPM/pull/469#discussion_r321275524), I'm not sure this behaviour is actually desired. The request made by the script would have been attributed to the page's URL before [PR#469](https://github.com/mozilla/OpenWPM/pull/469), now it gets the URL of the script itself. The URL-attribution code might need to be adapted to handle requests by worker scripts differently.
Also, as mentioned, I wasn't able to create a similar test for a page with a Service Worker because that would reuqire the page to be served over https.

### DIFF
--- a/test/test_http_instrumentation.py
+++ b/test/test_http_instrumentation.py
@@ -364,8 +364,8 @@ HTTP_CACHED_REDIRECTS = {
      u'http://localtest.me:8000/test_pages/shared/test_image_2.png'),
 }
 
-# Test URL attribution for servie worker requests
-HTTP_SERVICE_WORKER_REQUESTS = {
+# Test URL attribution for worker script requests
+HTTP_WORKER_SCRIPT_REQUESTS = {
     (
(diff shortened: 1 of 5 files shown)


englehardt:
The current test failures on Travis were introduced by #469. Would you mind to fix those as part of this PR?

englehardt:
> I'd like to, but so far I can't find any connection between failures errors and the changes I made. Also, the first Travis build after #469 was merged was successful, then they suddenly started to fail. Any idea what might be the cause?

You're right. It looks like these are caused by `localstack`, which we've had issues with in the past. I created #490 as a fix.

englehardt:
The new service worker test is much appreciated and looks good! I'm going to merge this in.

PR #490, opened 2019-09-14
Update localstack to newest version

englehardt:
The S3Aggregator tests are failing again, but we didn't make any recent changes. I suspect this is due to `localstack` dependency issues, though I'm unable to reproduce the failures locally. Let's try removing the pin to an old version.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,2 +1,2 @@
 autopep8
-localstack==0.9.6
+localstack
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,2 +1,3 @@
 autopep8
 localstack
+requests
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,3 +1,2 @@
 autopep8
-localstack
-requests
+localstack[full]
--- a/test/test_env.py
+++ b/test/test_env.py
@@ -25,7 +25,7 @@ class TestDependencies(OpenWPMTest):
             line = line.strip()
             if line == "" or line[0] == "#":
                 continue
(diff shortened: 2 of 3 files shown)


englehardt:
It looks like using newest version of `localstack` fixes the test failures we're seeing on master. We have to install it with the `full` profile, as described [here](https://github.com/localstack/localstack#starting-locally-non-docker-mode) since we aren't launching it within a docker container.

requirements-dev.txt
```diff
@@ -1,2 +1,2 @@
 autopep8
-localstack==0.9.6
+localstack[full]
```
nhnt11: What does [full] mean? I couldn't find much from a quick google search and thought it'd be more productive to just ask here.

test/test_env.py
```diff
@@ -25,7 +25,7 @@ def check_py_pkgs(self, requirements_file):
             line = line.strip()
             if line == "" or line[0] == "#":
                 continue
-            pkg = re.split(r'[>=<]', line)[0]
+            pkg = re.split(r'[>=<\[]', line)[0]
```
nhnt11: Would be nice to have a comment here to indicate that this regex effectively extracts the package name and excludes the version info.

requirements-dev.txt
englehardt: I've also never seen this before, but it was recommended in the `localstack` README. My understanding is that it let's package managers specify "versions" of the same release of their package. We're installing the "full" version. See: https://setuptools.readthedocs.io/en/latest/pkg_resources.html#requirement-objects.

test/test_env.py
englehardt: I've added a comment for this.

PR #509, opened 2019-10-18
Use extension built within Dockerfile instead of from the local system

### LINKED ISSUE #508: Make dockerfile work from unconfigured repo
For some reason running ```build-extension.sh``` was required to build a proper docker image from the Dockerfile.
From what I can tell the only difference between the Dockerfile and the script is the ```cp dist/*.zip ./openwpm.xpi```at the end of the script.

vringar:
Made the Dockerfile do the same steps as build-extension.sh
Hopefully closes #508

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,6 +14,8 @@ COPY automation/Extension/webext-instrumentation/. ../webext-instrumentation
 RUN npm install
 RUN npm run build
 
+COPY dist/*.zip ./openwpm.xpi
+
 # Stage 2, build the main OpenWPM image
 FROM ubuntu:18.04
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -13,6 +13,7 @@ COPY automation/Extension/firefox/. ./
 COPY automation/Extension/webext-instrumentation/. ../webext-instrumentation
 RUN npm install
 RUN npm run build
+RUN cp dist/openwpm-1.0.zip openwpm.xpi
```
englehardt: I think we should use `COPY` here, instead of `RUN cp`?

vringar: `COPY` moves files from the host system, while `RUN cp` copies filed around inside the image.
At least this is my understanding after reading the documentation

vringar: Thus far we seem to have accidentally copied in the extension build on the host and never used the extension built inside of the image.

englehardt: Makes sense. Good work finding this!

PR #514, opened 2019-10-25
Removed one mention of python2 in README.md

### LINKED ISSUE #513: python(2) doesn't work in docker anymore
When I ran
 ``` 
docker run -v $PWD/docker-volume:/root/Desktop \   
 -it openwpm python /opt/OpenWPM/demo.py
``` 
as it is currently in the README, I got the error ` no module named psultil`.  
Since we've sunset python2 support I'd just go ahead and change the README to reflect that.

vringar:
Removed it under "Running Measurements from inside the Container" as the instruction wasn't working anymore.
Closes #513


englehardt:
Good catch! Thanks for the update :)

Issue #448, opened 2019-08-09
Reduce the surface for bot detection


englehardt:
There are likely a number of ways to identify that we're running Firefox with Selenium/geckodriver. Back in the Selenium 2 days these were injected by the Selenium extension. We made some efforts to prevent that (https://github.com/mozilla/OpenWPM/pull/108). We later removed them with the upgrade to Selenium 3 because, at least at the time, Selenium 3 didn't self-identify via `navigator.webdriver` (https://github.com/mozilla/OpenWPM/pull/152). I'm guessing that's no longer the case.

See also:
* https://developer.mozilla.org/en-US/docs/Web/API/Navigator/webdriver
* https://antoinevastel.com/bot%20detection/2017/08/05/detect-chrome-headless.html
* https://github.com/antoinevastel/fpscanner

englehardt:
It looks like WebGL doesn't work properly in headless mode:

```
selenium_firefox     - DEBUG    - BROWSER -1382707880: driver: JavaScript warning: https://login.taobao.com/member/login.jhtml?tpl_redirect_url=https%3A%2F%2Fwww.tmall.com&style=miniall&enup=true&newMini2=true&full_redirect=true&sub=true&from=tmall&allp=assets_css%3D3.0.10/login_pc.css&pms=1566085516350, line 450: Error: WebGL warning: getContext: Disallowing antialiased backbuffers due to blacklisting.
selenium_firefox     - DEBUG    - BROWSER -1382707880: driver: JavaScript warning: https://login.taobao.com/member/login.jhtml?tpl_redirect_url=https%3A%2F%2Fwww.tmall.com&style=miniall&enup=true&newMini2=true&full_redirect=true&sub=true&from=tmall&allp=assets_css%3D3.0.10/login_pc.css&pms=1566085516350, line 450: Error: WebGL warning: <SetDimensions>: Can't use WebGL in headless mode (https://bugzil.la/1375585).
selenium_firefox     - DEBUG    - BROWSER -1382707880: driver: JavaScript warning: https://login.taobao.com/member/login.jhtml?tpl_redirect_url=https%3A%2F%2Fwww.tmall.com&style=miniall&enup=true&newMini2=true&full_redirect=true&sub=true&from=tmall&allp=assets_css%3D3.0.10/login_pc.css&pms=1566085516350, line 450: Error: WebGL warning: <SetDimensions>: Failed to create WebGL context: WebGL creation failed:
[...]

PR #526, opened 2019-11-05
Spoof `navigator.webdriver` to `false`

Flnch:
### What?
Regarding to  #448 this PR extends the OpenWPM instrumentation extension (WebExtension) to spoof the `navigator.webdriver` attribute to `false`. This is done by overwriting `window.navigator` with a JavaScript proxy.

### Why a JavaScript proxy?
Why not simply do a `Object.defineProperty(navigator, 'webdriver', {enumerable: true, value: false})`? Well, this changes the `webdriver` attribute but also the order of the `window.navigator` attributes. That is, the overwritten `webdriver` attribute is listed as first while normally listed later:
```
"permissions": {},
"doNotTrack": "unspecified",
...
"webdriver": true,
...
"storage": {}
```
becomes
```
"webdriver": false,
"permissions": {},
"doNotTrack": "unspecified",
...
"storage": {}
```
Thus, it would still be detectable.

### How does it work?
Simply set `spoof_navigator` to `True` and the instrumentation extension registers the content script that spoofs the `navigator` object with a JavaScript proxy. This option defaults to `False`.

### Other remarks
- I did my best to comply with the project structure but I am keen to get your feedback.
- The JavaScript proxy code is based on another Firefox extension (also GPLv3) as indicated in `automation/Extension/firefox/spoof.js/index.js` and `LICENSE`.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -26,3 +26,4 @@ automation/Extension/firefox/dist
 automation/Extension/firefox/openwpm.xpi
 automation/Extension/firefox/src/content.js
 automation/Extension/firefox/src/feature.js
+automation/Extension/firefox/src/spoof.js
--- a/.gitignore
+++ b/.gitignore
@@ -89,3 +89,4 @@ automation/Extension/firefox/dist
 automation/Extension/firefox/openwpm.xpi
 automation/Extension/firefox/src/content.js
 automation/Extension/firefox/src/feature.js
+automation/Extension/firefox/src/spoof.js
--- a/LICENSE
+++ b/LICENSE
@@ -41,6 +41,11 @@ https://github.com/redline13/selenium-jmeter
 By Richard Friedman
 Licensed GPLv3+
 
+Incorporating code from User Agent Switcher
+https://gitlab.com/ntninja/user-agent-switcher/
+Copyright © 2017 – 2019  Alexander Schlarb
+Licensed GPLv3+
+
 Text of GPLv3 License:
 ======================
                     GNU GENERAL PUBLIC LICENSE
--- a/README.md
+++ b/README.md
@@ -296,6 +296,8 @@ left out of this section.
   * **NOT SUPPORTED.** See [#101](https://github.com/citp/OpenWPM/issues/101).
   * Set to `True` to enable Firefox's built-in
     [Tracking Protection](https://developer.mozilla.org/en-US/Firefox/Privacy/Tracking_Protection).
+* `spoof_navigator`
+  * Set to `True` to spoof the JavaScript DOM attribute `webdriver` to be `false`.
 
 Browser Profile Support
 -----------------------
--- a/automation/Extension/firefox/feature.js/index.js
+++ b/automation/Extension/firefox/feature.js/index.js
@@ -22,6 +22,7 @@ async function main() {
       js_instrument:true,
       js_instrument_modules:"fingerprinting",
       http_instrument:true,
+      spoof_navigator:false,
       save_content:false,
       testing:true,
       crawl_id:0
@@ -60,9 +61,11 @@ async function main() {
                        config['save_content']);
   }
 
-  loggingDB.logDebug("Now spoofing `webdriver` attribute to `false`"); // TODO Add config option
-  let spoofNavigator = new SpoofNavigator();
-  await spoofNavigator.registerContentScript();
+  if (config['spoof_navigator']) {
+    loggingDB.logDebug("Now spoofing `webdriver` attribute to `false`");
+    let spoofNavigator = new SpoofNavigator();
+    await spoofNavigator.registerContentScript();
+  }
 }
 
 main();
--- a/automation/Extension/firefox/feature.js/index.js
+++ b/automation/Extension/firefox/feature.js/index.js
@@ -3,6 +3,7 @@ import {
   JavascriptInstrument,
   HttpInstrument,
   NavigationInstrument,
+  SpoofNavigator,
 } from "openwpm-webext-instrumentation";
 
 import * as loggingDB from "./loggingdb.js";
@@ -58,6 +59,10 @@ async function main() {
     httpInstrument.run(config['crawl_id'],
                        config['save_content']);
   }
+
+  loggingDB.logDebug("Now spoofing `webdriver` attribute to `false`"); // TODO Add config option
+  let spoofNavigator = new SpoofNavigator();
+  await spoofNavigator.registerContentScript();
 }
 
 main();
--- a/automation/Extension/firefox/spoof.js/index.js
+++ b/automation/Extension/firefox/spoof.js/index.js
@@ -1,0 +1,1 @@
+console.log("CODE WOULD BE EXECUTED");
--- a/automation/Extension/firefox/spoof.js/index.js
+++ b/automation/Extension/firefox/spoof.js/index.js
@@ -1,1 +1,74 @@
-console.log("CODE WOULD BE EXECUTED");
+/*
+/* This code to spoof values of the `window.navigator` object using a
+ * JavaScript proxy is based on:
+ * User Agent Switcher
+ * Copyright © 2017 – 2019  Alexander Schlarb (https://gitlab.com/ntninja)
+/* For the used part see: https://gitlab.com/ntninja/user-agent-switcher/blob/6aacc15ed6651317776f7abb3a85d6f34fc1a254/content/override-navigator-data.js
+ * 
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ * 
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * Set of all object that we have already proxied to prevent them from being
+ * proxied twice.
+ */
+let proxiedObjects = new Set();
+
+/**
+ * Convenience wrapped around `cloneInto` that enables all possible cloning
+ * options by default.
+ */
+function cloneIntoFull(value, scope) {
+  return cloneInto(value, scope, {
+    cloneFunctions: true,
+    wrapReflectors: true
+  });
+}
+
+/**
+ * Spoof `navigator` by overwriting the `navigator` of the given
+ * (content script scope) `window` object with a proxy, if applicable.
+ */
+function spoofNavigator(window) {
+  if (!(window instanceof Window)) { // Not actually a window object
+    return window;
+  }
+
+  let origNavigator = window.navigator.wrappedJSObject; // `navigator` of the page scope
+  if (proxiedObjects.has(origNavigator)) { // Window was already shadowed
+    return window;
+  }
+
+  let spoofedGet_PageScope = cloneIntoFull({
+    get: (target, prop, receiver) => {
+      if (prop === "webdriver") {
+        return false;
+      } else {
+        return Reflect.get(origNavigator, prop);
+      }
+    }
+  }, window.wrappedJSObject); // The `get` function, defined in privileged code (that is here in the extension / content script), is cloned into the target scope (that is the web page / `window.wrappedJSObject`) and thus accessible there. The return value is the reference to the cloned object in the defined scope.
+
+  let origProxy = window.wrappedJSObject.Proxy;
+  let navigatorProxy = new origProxy(origNavigator, spoofedGet_PageScope);
+
+  proxiedObjects.add(origNavigator);
+
+  let returnFunc_PageScope = exportFunction(() => {
+    return navigatorProxy;
+  }, window.wrappedJSObject);
+  // Using `__defineGetter__` here our function gets assigned the correct
+  // name of `get navigator`. Additionally its property descriptor has
+  // no `set` function and will silently ignore any assigned value. – This
+  // exact configuration is not achievable using `Object.defineProperty`.
+  Object.prototype.__defineGetter__.call(window.wrappedJSObject, "navigator", returnFunc_PageScope);
+  return window;
+}
+
+spoofNavigator(window);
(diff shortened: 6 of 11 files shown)


englehardt:
This is very well done, and I apologize for taking so long to get to this review. Thanks so much for the contribution! I've just tested this on the Google login flow (which prevent browsers that have been automated from logging in), and the spoofing fools it.

Would you mind to change `spoof_navigator` config option over to `hide_webdriver`. I would also appreciate a simple test page that checks if the webdriver attribute is set with and without this change. Aside from that, these changes look great!

Flnch:
> Would you mind to rename the `spoof_navigator` config option to `hide_webdriver`. I think that's more descriptive and agnostic to the underlying changes necessary to hide the fact that we're using webdriver. We could also tie future protections into that same config option (which may not change the navigator).

Of course. Your suggested name is more appropriate. I will change this tomorrow.

> I would also appreciate a simple test page that checks if the webdriver attribute is set with and without this change.

Do you mean something like [this](https://github.com/Flnch/webdriver-testpage)? I created this test that reads the `navigator.webdriver` property in three different scopes. This confirmed my assumption: The current solution does not spoof the `navigator.webdriver` property for an iframe created during runtime (see test page). I am already working on a fix for this problem.

englehardt:
> > Would you mind to rename the `spoof_navigator` config option to `hide_webdriver`. I think that's more descriptive and agnostic to the underlying changes necessary to hide the fact that we're using webdriver. We could also tie future protections into that same config option (which may not change the navigator).
> 
> Of course. Your suggested name is more appropriate. I will change this tomorrow.
> 
> > I would also appreciate a simple test page that checks if the webdriver attribute is set with and without this change.
> 
> Do you mean something like [this](https://github.com/Flnch/webdriver-testpage)? I created this test that reads the `navigator.webdriver` property in three different scopes. This confirmed my assumption: The current solution does not spoof the `navigator.webdriver` property for an iframe created during runtime (see test page). I am already working on a fix for this problem.

Yes exactly! And thanks for letting me know that it doesn't apply to dynamically created iframes. Do you know the cause? Just wondering if that's also true for the way we [inject instrumentation](https://github.com/mozilla/OpenWPM/blob/8ed3798c45ce01729458a794748fe018a81727e2/automation/Extension/webext-instrumentation/src/background/javascript-instrument.ts#L113-L141).

As an example of how to build an automated test from your current test page you can look at [our test page](https://github.com/mozilla/OpenWPM/pull/108/files#diff-9ad8dd052529add585f30d586c08c2c7) from when we previously masked this attribute.

Flnch:
> And thanks for letting me know that it doesn't apply to dynamically created iframes. Do you know the cause? Just wondering if that's also true for the way we [inject instrumentation](https://github.com/mozilla/OpenWPM/blob/8ed3798c45ce01729458a794748fe018a81727e2/automation/Extension/webext-instrumentation/src/background/javascript-instrument.ts#L113-L141).

The [MDN page about content scripts](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/content_scripts) mentions:

> Note that in Firefox, content scripts won't be injected into empty iframes at `document_start` even if you specify that value in `run_at`.

I think this is the cause. Then, indeed, it would also apply to the injected instrumentation.

The creator of the user-agent-switcher extension I oriented the code at also [dealt with this issue](https://gitlab.com/ntninja/user-agent-switcher/blob/master/content/override-navigator-data.js#L165-249). He tracks added iframes and applies the JavaScript proxy overwrite to them. A first test shows using his solution would work as a fix for the `navigator.webdriver` attribute.

englehardt:
> The [MDN page about content scripts](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/content_scripts) mentions:
> 
> > Note that in Firefox, content scripts won't be injected into empty iframes at `document_start` even if you specify that value in `run_at`.
> 
> I think this is the cause. Then, indeed, it would also apply to the injected instrumentation.

Thanks, I think we missed that during the move from the addon-sdk to WebExtensions. I filed #545 to investigate.

englehardt:
I think your approach may cause some breakage. Take a look at: `https://www.reddit.com/r/videos/` with and without the spoofing enabled. When it's enabled you'll see the error message: `TypeError: 'javaEnabled' called on an object that does not implement interface Navigator.`. The page also fails to load any advertisements. This error is not present and advertisements are loaded when the spoofing is disabled (but webdriver is still used).

Flnch:
> I think your approach may cause some breakage. Take a look at: `https://www.reddit.com/r/videos/` with and without the spoofing enabled. When it's enabled you'll see the error message: `TypeError: 'javaEnabled' called on an object that does not implement interface Navigator.`. The page also fails to load any advertisements. This error is not present and advertisements are loaded when the spoofing is disabled (but webdriver is still used).

Indeed. I pushed a fix for this issue. [Here](https://gitlab.com/ntninja/user-agent-switcher/blob/master/content/override-navigator-data.js#L80) they did fix the problem but because of the comment "Workarounds for Firefox 68 and below" I did not suspect this to be still a problem.

Flnch:
So, there is a further detection method. Using
```
window.open('').navigator.webdriver
```
the original `webdriver` value is revealed because it accesses the new page's DOM that is not spoofed. In regular Firefox, such pop-up windows are blocked by default but for OpenWPM not.

@englehardt Is it for a specific reason OpenWPM disabled the pop-up window protection? As far as I understand the setting modification originates from `OpenWPM/automation/Extension/firefox/node_modules/firefox-profile/lib/firefox_profile.js:45: 'dom.disable_open_during_load': 'false'`. It is possible to re-activate the pop-up protection in `configure_firefox.py`.

englehardt:
> So, there is a further detection method. Using
> 
> ```
> window.open('').navigator.webdriver
> ```
> 
> the original `webdriver` value is revealed because it accesses the new page's DOM that is not spoofed. In regular Firefox, such pop-up windows are blocked by default but for OpenWPM not.

Do you know why the extension doesn't inject a content script into the newly created window? Is it simply that the script is able to access the webdriver attribute before the happens or is this related to https://github.com/mozilla/OpenWPM/pull/526#issuecomment-567464231?

> @englehardt Is it for a specific reason OpenWPM disabled the pop-up window protection? As far as I understand the setting modification originates from `OpenWPM/automation/Extension/firefox/node_modules/firefox-profile/lib/firefox_profile.js:45: 'dom.disable_open_during_load': 'false'`. It is possible to re-activate the pop-up protection in `configure_firefox.py`.

This wasn't changed by us. The file you've referenced is part of the firefox profile librarr (https://github.com/saadtazi/firefox-profile-js), which is required by the `web-ext` package. I don't think we actually use that package, and instead that the pref is [set by Selenium](https://github.com/SeleniumHQ/selenium/blob/8a0691f55df5868b518e0f53c3c90b6251607dd4/third_party/js/selenium/webdriver.json#L28). It's a frozen pref, but we can [overwrite](https://github.com/mozilla/OpenWPM/blob/9686b8402925336fc0fc9d69780a6551da3b589e/automation/DeployBrowsers/deploy_firefox.py#L57-L60) it.

I suspect the reason it's set to false is that Selenium is primarily meant for website testing where site owners may want to run through tests that continually open new windows without user interaction. That's not as important to us, and it seems more true to the actual use experience to block those popups.

So, I'm happy to set that pref back to true. Are you up for doing that in this PR?

Also, do you expect to be able to add automated tests to the test suite to check for this property? If not, I'll merge this as is and file a follow-up bug for tests.

Flnch:
> Do you know why the extension doesn't inject a content script into the newly created window? Is it simply that the script is able to access the webdriver attribute before the happens or is this related to [#526 (comment)](https://github.com/mozilla/OpenWPM/pull/526#issuecomment-567464231)?

I could not find out yet why this happens. It opens an `about:blank` page where the injection in principle  [should happen](https://github.com/mozilla/OpenWPM/blob/80ef59ad6e0731aa5d512c2f950783050d46f6fc/automation/Extension/webext-instrumentation/src/background/spoof-navigator.ts#L13). But even in the console of the opened page, `webdriver` is not spoofed. However, if I manually open a new window (via Ctrl+N or the menu), `webdriver` is spoofed when checking in the console.

> This wasn't changed by us. The file you've referenced is part of the firefox profile library (https://github.com/saadtazi/firefox-profile-js), which is required by the `web-ext` package. I don't think we actually use that package, and instead that the pref is [set by Selenium](https://github.com/SeleniumHQ/selenium/blob/8a0691f55df5868b518e0f53c3c90b6251607dd4/third_party/js/selenium/webdriver.json#L28). It's a frozen pref, but we can [overwrite](https://github.com/mozilla/OpenWPM/blob/9686b8402925336fc0fc9d69780a6551da3b589e/automation/DeployBrowsers/deploy_firefox.py#L57-L60) it.
> 
> I suspect the reason it's set to false is that Selenium is primarily meant for website testing where site owners may want to run through tests that continually open new windows without user interaction. That's not as important to us, and it seems more true to the actual use experience to block those popups.
> 
> So, I'm happy to set that pref back to true. Are you up for doing that in this PR?

Already done :) I hope `configure_firefox.optimize_prefs` is the correct place. Otherwise I can change that. In each case this fixes the `window.open('')` case.

Further, I migrated the iframe solution [mentioned earlier](https://github.com/mozilla/OpenWPM/pull/526#issuecomment-567464231). This does not prevent all access methods but does its best, see the [commit description](https://github.com/mozilla/OpenWPM/pull/526/commits/84e4b4811b4bcc8dca4f817e18252d97c850a02d). I also updated the [test page](https://github.com/Flnch/webdriver-testpage) to cover all combinations I am aware off, including the `window.open('')` method.

> Also, do you expect to be able to add automated tests to the test suite to check for this property? If not, I'll merge this as is and file a follow-up bug for tests.

That depends on how easy it is to integrate my test page. I looked into [your reference](https://github.com/mozilla/OpenWPM/pull/108/files#diff-9ad8dd052529add585f30d586c08c2c7) how testing was done previously. JavaScript code was executed via Selenium. But for my test page the timing is relevant as earlier access to JavaScript properties makes a difference. Do you think it is easy to read the results of my test page from the HTML page (they are dynamically added in a HTML list) into the Python test? Or is there a more appropriate solution? I am not advanced in writing such tests so maybe adjusting my test page to also write the results in a JSON structure that can be accessed by Selenium in Python or so will be easier.

Flnch:
@englehardt Any news on this? Momentarily, I have not much time. But if you want to merge the current status I am fine with adding automated tests in a few weeks when I have resources to do so.

englehardt:
Thanks for poking me on this!

> @englehardt Any news on this? Momentarily, I have not much time. But if you want to merge the current status I am fine with adding automated tests in a few weeks when I have resources to do so.

It looks like your previous updates broke a bunch of the tests. Would you mind to check into those test failures to see whether you've introduced an unexpected bug?

> Already done :) I hope `configure_firefox.optimize_prefs` is the correct place. Otherwise I can change that. In each case this fixes the `window.open('')` case.

That's fine!
 
> That depends on how easy it is to integrate my test page. I looked into [your reference](https://github.com/mozilla/OpenWPM/pull/108/files#diff-9ad8dd052529add585f30d586c08c2c7) how testing was done previously. JavaScript code was executed via Selenium. But for my test page the timing is relevant as earlier access to JavaScript properties makes a difference. Do you think it is easy to read the results of my test page from the HTML page (they are dynamically added in a HTML list) into the Python test? Or is there a more appropriate solution? I am not advanced in writing such tests so maybe adjusting my test page to also write the results in a JSON structure that can be accessed by Selenium in Python or so will be easier.

To make sure I understand, you'd like to be able to check for the webdriver attribute as early as possible and not introduce the lag that comes with executing JS through Selenium?

In that case your suggestion sounds reasonable. The test page can run the full set of tests and put the results of those tests somewhere that the JS we execute through Selenium can access (e.g., on the `window` object).

If we manage to get the current set of tests passing again, I'm happy to merge this without additional tests and add those in a follow-up PR.

Flnch:
> It looks like your previous updates broke a bunch of the tests. Would you mind to check into those test failures to see whether you've introduced an unexpected bug?

I merged remote upstream to see whether this caused the failed tests. Travis-CI tests do fail but it seems to be a problem with automated testing, e.g. with `flake8`. Local testing does not work because `py.test -vv` states `ValueError: attempted relative import beyond top-level package`, which I do not understand as the current working directory correctly is inside the `test` folder.

> To make sure I understand, you'd like to be able to check for the webdriver attribute as early as possible and not introduce the lag that comes with executing JS through Selenium?

Exactly.

englehardt:
> > Do you know why the extension doesn't inject a content script into the newly created window? Is it simply that the script is able to access the webdriver attribute before the happens or is this related to [#526 (comment)](https://github.com/mozilla/OpenWPM/pull/526#issuecomment-567464231)?
> 
> I could not find out yet why this happens. It opens an `about:blank` page where the injection in principle [should happen](https://github.com/mozilla/OpenWPM/blob/80ef59ad6e0731aa5d512c2f950783050d46f6fc/automation/Extension/webext-instrumentation/src/background/spoof-navigator.ts#L13). But even in the console of the opened page, `webdriver` is not spoofed. However, if I manually open a new window (via Ctrl+N or the menu), `webdriver` is spoofed when checking in the console.

I wonder if this is related to the fact that you're loading `about:blank`. We've recently done a bunch of work to ensure that sites can't inject code into privileged `about:` pages. Do you see the same issue on other non-privileged pages?


> > It looks like your previous updates broke a bunch of the tests. Would you mind to check into those test failures to see whether you've introduced an unexpected bug?
> 
> I merged remote upstream to see whether this caused the failed tests. Travis-CI tests do fail but it seems to be a problem with automated testing, e.g. with `flake8`.

This means you have style issues that must be fixed. You can try running https://github.com/hhatto/autopep8 to fix these automatically, but looking at the outputs they seem pretty simple to fix by hand (just sorting issues for dependencies).

> Local testing does not work because `py.test -vv` states `ValueError: attempted relative import beyond top-level package`, which I do not understand as the current working directory correctly is inside the `test` folder.

You're running `pytest -vv` from the `test/`?

Also just to note that you'll also need to run the dev install script (i.e., `./install-dev.sh`) to run all of the tests, but that is unrelated to the error you're seeing.

PR #528, opened 2019-11-14
Registering signal handler to handle shutdown gracefully

### LINKED ISSUE #450: Minimize data loss during unexpected shutdown
In a recent pair of 108k crawls in GCP, this is what happened:

1. After an initial crawl run, about 83k crawl_history records were found in the S3 bucket (about 70%)
2. After creating a new crawl list (n = 25k) with the items of the original seed list that did not yet have a crawl_history record, about 18k more, or 101k crawl_history records in total, were found in the S3 bucket (about 72% of the 25k crawled made it)
3. After step 2 repeated with now a 7k list, about 107k crawl_history records were found in total  (about 85% of the 7k crawled made it)

I am filing this separate to https://github.com/mozilla/OpenWPM/issues/255, which addresses percentage of command failures in relation to the total number of commands submitted. This issue does not relate the value of bool_success in the crawl_history table, but to the existence of a record at all, and currently it seems that the data for around 15-30% of crawled sites in each crawl-attempt gets dropped.

vringar:
To reduce data loss in Kubernetes
Closes #450


vringar:
the while loop doesn't work
somehow when I run it from a bash shell attached to the running image bash writes `Terminated` to stdout and exits with a 132 return code.

vringar:
After spending 1,5 h on this without any meaningful progress I'm considering writing the cleaner solution of propagating the signal from the main thread as that would also guarantee flushing for `docker kill`

vringar:
The scripts name contained python and thus it gets listed by `pgrep python` leading to it's termination.
Renaming the script should fix this behavior

vringar:
I gave up on getting the script to work. 
Instead I've now moved the signal handlers to the TaskManager and handle the shutdown from there. I also had to change the way the dockerfile starts by default, as `CMD` spawns a bash shell as PID 1 which doesn't pass on signals.
There are multiple problems arising from the current state.

1. ctrl+c doesn't work properly anymore (as in it creates a huge mess of output)
2. The `TaskManager` now throws an exception when it gets called while closed, because I haven't figured out a better way to communicate to the client/user, that it should also shutdown rn.
3. As PID 1 it is our responsibility to deal with zombie processes (aka killed firefox instances and more). Here we could use [an existing wrapper](https://github.com/Yelp/dumb-init) that forwards the signals to our python script and handles the rest for us.

vringar:
Some additional thoughts and comments:
1. seems acceptable
2. Maybe we should call `sys.exit` once shutdown returns or maybe we should move the signal handlers to the crawler script instead of the platform. Because at the end of the day it's not the platforms job to figure out if it should shutdown but the users. But I'm not sure on this.
3. If we don't use the wrapper long running containers will accumulate hundreds of zombie processes which doesn't seem acceptable.

vringar:
Fixed 3. by using the wrapper that docker uses if you pass `--init` to `run`.
So now we don't have to deal with zombie processes and get our signals forwarded.

vringar:
This branch is no longer under active developement as we have found a better approach to things

PR #532, opened 2019-11-20
Removed all traces of python2

vringar:
Breaking all the backwards compat to move forward to a less cluttered future


automation/MPLogger.py
```diff
@@ -265,10 +263,22 @@ def _handle_serialized_writes(self, obj):
         This is currently records that are written to a file on disk
         and those sent to Sentry.
         """
+        def ensure_str(s):
+            """Ensures we get a string
+
+                Stolen and simplified from six
```
englehardt: This looks right, but please just clean up this comment to "Taken and simplified from the six library".

englehardt: In fact I think this is simple enough that we don't need to reference six.

requirements.txt
```diff
@@ -2,7 +2,7 @@
 # This file is autogenerated by pip-compile
 # To update, run:
 #
-#    pip-compile requirements.in
+#    pip-compile
```
englehardt: The instructions on how to update these files were incorrect. Mind to review and merge https://github.com/mozilla/OpenWPM/pull/533 (assuming the tests pass) and then use those commands to re-compile the `.txt` files?

vringar: Done

automation/MPLogger.py
```diff
@@ -265,10 +263,21 @@ def _handle_serialized_writes(self, obj):
...
+            """
+            if not isinstance(s, (str, bytes)):
+                raise TypeError("not expecting type '%s'" % type(s))
+            elif isinstance(s, bytes):
+                s = s.decode('utf-8', 'strict')
+            return s
+
         if obj['exc_info']:
-            obj['exc_info'] = dill.loads(six.ensure_str(obj['exc_info']))
+            obj['exc_info'] = dill.loads(ensure_str(obj['exc_info']))
```
vringar: @englehardt Has this ever worked?
Dill says it converts objects into an arbitrary byte stream.
So from my understanding we would need to do it the other way around.
```suggestion
            obj['exc_info'] = ensure_str(dill.loads(obj['exc_info']))
```

automation/MPLogger.py
```diff
@@ -265,10 +263,21 @@ def _handle_serialized_writes(self, obj):
...
         if obj['args']:
-            obj['args'] = dill.loads(six.ensure_str(obj['args']))
+            obj['args'] = dill.loads(ensure_str(obj['args']))
```
vringar: Same here
```suggestion
            obj['args'] = ensure_str(dill.loads(obj['args']))
```

automation/MPLogger.py
vringar: This code was introduced [here](https://github.com/mozilla/OpenWPM/commit/8a4a293ab4f58cd5885ce3067d6629df374090c3)

vringar: Just tried switching the order, that doesn't work either as exc_info contains a tuple, so I'm updating my suggestion to just remove `ensure_str` unless there's something I'm missing here.

```suggestion
            obj['exc_info'] = dill.loads(obj['exc_info'])
```

vringar: after introducing some hacky logging I see this:
``` python
obj['exc_info'] is of type <class 'tuple'> :  (<class 'selenium.common.exceptions.WebDriverException'>, WebDriverException("invalid argument: can't kill an exited process", None, None), <traceback object at 0x7f1a36320188>)
```

vringar: This would mean that we never sent exceptions to sentry, because we would always error out before sending them?

englehardt: I filed https://github.com/mozilla/OpenWPM/issues/534 to investigate why our tests failed to capture this.

englehardt:
I added a note to the README as we discussed. Merging now.

PR #540, opened 2019-12-13
Callstacks

### LINKED ISSUE #352: Restore instrumentation regarding what code is causing requests
**I'm submitting a ...**
[ ] bug report
[X] feature request
[ ] question about the decisions made in the repository
[ ] question about how to use this project

**Summary**
Since refactoring to Web Extensions and supporting multi-process Firefox, it is no longer possible to capture a meaningful stack trace in the request listener. 

The purpose of the previously gathered stacktrace was primarily to understand what scripts are sending XHR requests or adding elements with src-attributes to the DOM (thus initiating requests). 

It has been used it to determine which scripts are responsible for leaking PII to external domains. For example, we've captured PII leaked to major data brokers by unrelated companies. If we only have the URL of the request, we wouldn't know whether the data broker itself was responsible for the request.

Original discussion at https://github.com/mozilla/openwpm-webext-instrumentation/pull/7/files#r229142721

**Other information**
We can probably instrument some of this via the page script (overloading XHR-related prototypes, watching DOM changes etc), or possibly create a webextension experiment that specifically listens to the origin of requests and notes down stacktraces or similar (given that such a strategy is feasible in multi-process Firefox).

vringar:
Just opening this so I can write my thoughts inline.
Closes #352


automation/Extension/firefox/feature.js/callstack-instrument.js
```diff
@@ -0,0 +1,19 @@
+/*
+    This script catches the stacktraces of JavaScript that ran on a crawled site and made a WebRequest
```
nhnt11: nit: 2-space indent throughout this file (do we have linting for this)

nhnt11: Can we re-word this a bit? I suggest something like:
```
We capture the JS callstack when we detect a dynamically created http request
and bubble it up via a WebExtension event API.
This instrumentation captures those and saves them to the "callstacks" table.
```

automation/Extension/firefox/feature.js/callstack-instrument.js
```diff
@@ -0,0 +1,19 @@
...
+    and saves them to the "callstacks" table
+*/
+export class CallstackInstrument {
+    constructor(dataReceiver) {
+        this.dataReceiver = dataReceiver;
+    }
+    run(crawlID) {
+        browser.stackDump.onStackAvailable.addListener((requestId, stack) => {
```
nhnt11: If we use `crawl_id`, `request_id` and `call_stack` as the parameter names, then we can use shorthand notation in the object:
```
this.dataReceiver.saveRecord("callstacks", { crawl_id, request_id, call_stack });
```

automation/Extension/firefox/src/privileged/stackDump/OpenWPMStackDumpChild.jsm
```diff
@@ -0,0 +1,170 @@
...
+  }
+}
+
+
+class OpenWPMStackDumpChild extends JSWindowActorChild {
+  constructor() {
+    super();
+  }
+  actorCreated() {
+    //We need the Controller because registering an observer on the child didn't work
```
nhnt11: Please add some details to this comment:
```
JSWindowActorChild (which is a WebIDL class) instances are not directly usable as nsIObserver instances, so we need to use a separate object to observe requests.
```

automation/Extension/webext-instrumentation/src/background/callstack-instrument.js
```diff
@@ -0,0 +1,17 @@
+
```
nhnt11: something is confusing - why does this live here as well as in the feature.js webpack dir? Aren't we doing some automated copying of files? @englehardt can you remind me how this works?

vringar: Sorry, this is a leftover from my first attempt of implementing the listener. It doesn't belong here

automation/Extension/firefox/feature.js/callstack-instrument.js
vringar: I don't think we have any linting set up for .js files, I've only seen linters for the .ts and .py files.

automation/Extension/firefox/src/privileged/stackDump/OpenWPMStackDumpChild.jsm
```diff
@@ -0,0 +1,175 @@
+"use strict";
+
+var EXPORTED_SYMBOLS = ["OpenWPMStackDumpChild"];
+const { Services } = ChromeUtils.import("resource://gre/modules/Services.jsm");
+
+class Controller {
+  constructor(actor) {
+    this.actor = actor;
+    Services.obs.addObserver(this, "http-on-opening-request");
```
englehardt: Why do we need to register two observers here?

automation/Extension/webext-instrumentation/package.json
```diff
@@ -20,7 +20,6 @@
     "fix:prettier": "prettier \"src/**/*.ts\" --write",
     "fix:tslint": "tslint -t verbose --fix --project .",
     "test": "run-s build test:*",
-    "test:lint": "tslint -t verbose --project . && prettier \"src/**/*.ts\" --list-different",
```
englehardt: Why remove this?

deploy-extension.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
```
englehardt: Isn't this effectively doing the same thing as `/test/manual_test.py`? Is there a reason to prefer this over that?

automation/Extension/firefox/src/privileged/stackDump/OpenWPMStackDumpChild.jsm
vringar: From my understanding it's just copied from [here](https://searchfox.org/mozilla-central/source/devtools/server/actors/network-monitor/stack-trace-collector.js#37) also the [comments in the IDL](https://searchfox.org/mozilla-central/source/netwerk/protocol/http/nsIHttpProtocolHandler.idl#127) indicate that document and http channels are two different things.
Maybe @nhnt11 can expand

automation/Extension/webext-instrumentation/package.json
vringar: @nhnt11

deploy-extension.sh
vringar: No, there isn't. I just forgot `test/manual_test.py` already does this

automation/Extension/firefox/src/privileged/stackDump/OpenWPMStackDumpChild.jsm
nhnt11: mmmm, ok, technically I think we can use just the http notification here. document channel is a more abstract channel type that exists before we're sure that this is an http channel. I'm not entirely sure why the devtools stack trace collector is using this. we should verify that removing it doesn't do anything bad first though, for sanity.

automation/Extension/webext-instrumentation/package.json
nhnt11: no reason to remove it, if this came from my original patch then it's likely just because I wanted to speed up my workflow when building.

automation/Extension/firefox/src/privileged/stackDump/OpenWPMStackDumpChild.jsm
vringar: https://bugzilla.mozilla.org/show_bug.cgi?id=1556489#c12
As per this comment this was intentional, so I'd just keep it as is

automation/Extension/webext-instrumentation/package.json
vringar: Reverted

deploy-extension.sh
vringar: I have deleted it

pickizme:
@vringar I now can not interpret the  `callstacks` table anymore. I also couldn't find documentation about that. Can you please help me to understand that table so that I can work with it?

Issue #519, opened 2019-10-29
Firefox 70 changed the ExtensionCommon.EventManager constructor


vringar:
After setting everything up by running `python -m test.manual_test --selenium`
Open up the extesion debug window and run
 ```JS
 let y = await browser.sockets.createServerSocket();
browser.sockets.startListening(y)
browser.sockets.onDataReceived.addListener ((socketID,data, aJson) => console.log(socketID));
``` 
and then
```Python
from automation.SocketInterface import clientsocket
client = clientsocket(verbose = True) 
client.connect("localhost", y)
client.send(3)
```
Nothing gets logged to console. Meaning the data is not getting through the experimental part of the API
Blocks #512

vringar:
Use `about:devtools-toolbox?type=extension&id=openwpm%40mozilla.org` to directly access the extension toolbox

vringar:
The connection gets established and the content arrives in the browser which can be demonstrated by sending a string and the extension logging into the browser console that encoding "u" is not supported.

vringar:
Okay, so after logging everything I can see that `gManager.onDataReceivedListeners` doesn't change when `browser.sockets.onDataReceived.addListener (x)` gets called.

vringar:
TODO: Figure out if something changed [here](https://dxr.mozilla.org/mozilla-central/source/toolkit/components/extensions/ExtensionCommon.jsm)

PR #554, opened 2020-01-17
Switched to latest release instead of specific version

### LINKED ISSUE #553: Update to Firefox 72
We should also check if there's a new geckodriver release and update that if necessary.

vringar:
With the link that @nhtn11 has received we no longer are bound to a specific version
but instead run on the lastest stable release

Closes #553

### DIFF
--- a/install-mac-dev.sh
+++ b/install-mac-dev.sh
@@ -31,7 +31,7 @@ brew install node || true
 # Use the Unbranded build that corresponds to a specific Firefox version (source: https://wiki.mozilla.org/Add-ons/Extension_Signing#Unbranded_Builds)
 brew install wget || true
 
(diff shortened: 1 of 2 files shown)


Flnch:
I wondered whether automatically using the latest stable version is so wise. Previously, new Firefox versions broke functionality of OpenWPM, e.g. #519. Maybe give the user the choice to use the latest version tested for OpenWPM compatibility or the latest stable version?

vringar:
Yeah :) Others had that concern as well. That's why I started searching for a link that would give me the addon build for a released version of Firefox.
Sorry for that I didn't document that thoguht here

vringar:
This is not a good UX and we can do better.

PR #556, opened 2020-01-31
[WIP] Code cleanup

englehardt:


README.md
```diff
@@ -144,7 +149,7 @@ available [below](#output-format).
         browser is running on the machine.
 * Cookie Access
     * Set `browser_params['cookie_instrument'] = True`
-    * Data is saved to the `javascript_cookies` table.
+    * Data is saved to the `cookies` table.
```
birdsarah, line 152: Is the cookies table a table of all cookies (including those set from http requests) or is it only those from JS.set_cookie?

englehardt: It's all cookies

birdsarah: I did not know that! Has it always been? (Just thinking about past data sources).

vringar: We just used to differentiate between `javascript_cookies` and `flash_cookies`, that's why there was any sort of prefix

PR #575, opened 2020-03-10
Setting site_rank in crawler.py

### LINKED ISSUE #564: Set site_rank in crawler.py
Since we now have the ability to save site rank in the DB we should set it in crawler.py

svensevenslow:
Closes #564


englehardt:
@skim1102

PR #578, opened 2020-03-11
Type annotation and refactoring in TaskManager.py

vringar:
Should be merged before #574


englehardt:
This looks great to me. Would you mind to update the README to include a note on types? We should request that folks to add them in iteratively.

PR #580, opened 2020-03-12
Adding Commands.py file

svensevenslow:
This a WIP PR. I have currently just added the base class and subclasses for the commands in the Commands.py file. Please let me know if this is the way you wanted it to be implemented(I have used factory method pattern) and if the naming of classes etc is ok. 
Also a few questions
1. Do we intend to pass the entire object to function calls for eg [here](https://github.com/mozilla/OpenWPM/blob/59e79a7adbb41b2a801b7f8fb9c2863cb6222cc7/automation/TaskManager.py#L457) or send something like obj.command
2. Also we are getting a visit_id [here](https://github.com/mozilla/OpenWPM/blob/59e79a7adbb41b2a801b7f8fb9c2863cb6222cc7/automation/TaskManager.py#L452) and also a visit_id [here](https://github.com/mozilla/OpenWPM/blob/master/automation/Commands/command_executor.py#L12). Which one am I supposed to maintain in the class?
3. How do you want me write the class for [this](https://github.com/mozilla/OpenWPM/blob/master/automation/Commands/command_executor.py#L59)?

### DIFF
--- a/automation/BrowserManager.py
+++ b/automation/BrowserManager.py
@@ -17,6 +17,7 @@ from selenium.common.exceptions import WebDriverException
 from tblib import pickling_support
 
 from .Commands import command_executor
+from .Commands.Commands import Shutdown_Command
 from .DeployBrowsers import deploy_browser
 from .Errors import BrowserConfigError, BrowserCrashError, ProfileLoadError
 from .SocketInterface import clientsocket
@@ -266,7 +267,8 @@ class Browser:
             return
 
         # Send the shutdown command
-        self.command_queue.put(("SHUTDOWN",))
+        command = Shutdown_Command()
+        self.command_queue.put((command))
 
         # Verify that webdriver has closed (30 second timeout)
         try:
@@ -456,11 +458,9 @@ def BrowserManager(command_queue, status_queue, browser_params,
                 time.sleep(0.001)
                 continue
 
-            # reads in the command tuple of form:
-            # (command, arg0, arg1, arg2, ..., argN) where N is variable
             command = command_queue.get()
 
-            if command[0] == "SHUTDOWN":
+            if type(command) is Shutdown_Command:
                 # Geckodriver creates a copy of the profile (and the original
                 # temp file created by FirefoxProfile() is deleted).
                 # We clear the profile attribute here to prevent prints from:
--- a/automation/Commands/Commands.py
+++ b/automation/Commands/Commands.py
@@ -1,0 +1,96 @@
+class Command:
+    def set_visit_crawl_id(self, visit_id, crawl_id):
+        self.visit_id = visit_id
+        self.crawl_id = crawl_id
+
+    def set_start_time(self, start_time):
+        self.start_time = start_time
+
+
+class Get_Command(Command):
+    def __init__(self, url, sleep):
+        self.url = url
+        self.sleep = sleep
+
+    def __repr__(self):
+        return "Get_Command({},{})".format(self.url, self.sleep)
+
+
+class Browse_Command(Command):
+    def __init__(self, url, num_links, sleep):
+        self.url = url
+        self.num_links = num_links
+        self.sleep = sleep
+
+    def __repr__(self):
+        return "Browse_Command({},{},{})".format(
+            self.url, self.num_links, self.sleep)
+
+
+class Dump_Flash_Cookies_Command(Command):
+    def __init__(self):
+        pass
+
+    def __repr__(self):
+        return "Dump_Flash_Cookies_Command()"
+
+
+class Dump_Prof_Command(Command):
+    def __init__(self, dump_folder, close_webdriver, compress):
+        self.dump_folder = dump_folder
+        self.close_webdriver = close_webdriver
+        self.compress = compress
+
+    def __repr__(self):
+        return "Dump_Prof_Command({},{},{})".format(
+            self.dump_folder, self.close_webdriver, self.compress)
+
+
+class Dump_Page_Source_Command(Command):
+    def __init__(self, suffix):
+        self.suffix = suffix
+
+    def __repr__(self):
+        return "Dump_Page_Source_Command({})".format(self.suffix)
+
+
+class Recursive_Dump_Page_Source_Command(Command):
+    def __init__(self, suffix):
+        self.suffix = suffix
+
+    def __repr__(self):
+        return "Recursive_Dump_Page_Source_Command({})".format(self.suffix)
+
+
+class Save_Screenshot_Command(Command):
+    def __init__(self, suffix):
+        self.suffix = suffix
+
+    def __repr__(self):
+        return "Save_Screenshot_Command({})".format(self.suffix)
+
+
+class Screenshot_Full_Page_Command(Command):
+    def __init__(self, suffix):
+        self.suffix = suffix
+
+    def __repr__(self):
+        return " Screenshot_Full_Page_Command({})".format(self.suffix)
+
+
+class Run_Custom_Function_Command(Command):
+    def __init__(self, function_handle, func_args):
+        self.function_handle = function_handle
+        self.func_args = func_args
+
+    def __repr__(self):
+        return "Run_Custom_Function_Command({},{})".format(
+            self.function_handle, self.func_args)
+
+
+class Shutdown_Command(Command):
+    def __init__(self):
+        pass
+
+    def __repr__(self):
+        return "Shutdown_Command()"
--- a/automation/Commands/command_executor.py
+++ b/automation/Commands/command_executor.py
@@ -1,5 +1,11 @@
 
+from ..Errors import CommandExecutionError
 from . import browser_commands, profile_commands
+from .Commands import (Browse_Command, Dump_Flash_Cookies_Command,
+                       Dump_Page_Source_Command, Dump_Prof_Command,
+                       Get_Command, Recursive_Dump_Page_Source_Command,
+                       Run_Custom_Function_Command, Save_Screenshot_Command,
+                       Screenshot_Full_Page_Command)
 
 
 def execute_command(command, webdriver, browser_settings, browser_params,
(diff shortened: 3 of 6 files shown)


vringar:
Hey, sorry for not getting back to you yesterday and thank you for seeking for feedback so quickly.
First off I think your names are good and while I don't think we'll need the factory here, it quite a nice pattern.
I think all of your questions are related to how a CommandSequence gets passed around the system so I'm gonna lay out it's way through the system first.
The way a Sequence of Commands is executed is:
1. User creates a CommandSequence (CS) with all the steps
2. User passes that CS to the TaskManager
3. TaskManager send the CS command per command to the BrowserManager [via the command_queue](https://github.com/mozilla/OpenWPM/blob/59e79a7adbb41b2a801b7f8fb9c2863cb6222cc7/automation/TaskManager.py#L457)
4. The BrowserManager then passes the command to the command_executor which then unpacks the sent command and executes it.

Answering your questions:
1. For the code linked we would pass the entire object
2. You would set the visit_id property on the object at the location of the first link and then have use this object in the second location
3. I'd want to you to modify the executor to make use of the named fields and use isInstance there

The following will just be pseudocode to illustrate how I was thinking it could be implemented:

In [CommandSequence.get()](https://github.com/mozilla/OpenWPM/blob/59e79a7adbb41b2a801b7f8fb9c2863cb6222cc7/automation/CommandSequence.py#L53) could look like:
```python
 def get(self, sleep=0, timeout=60):
        """ goes to a url """
        self.total_timeout += timeout
        command = Get_Command(self.url, sleep)
        self.commands_with_timeout.append((command, timeout))
        self.contains_get_or_browse = True
```
and the [command_executor](https://github.com/mozilla/OpenWPM/blob/master/automation/Commands/command_executor.py#L10) would look like:
```python
if type(command) is Get_Command:
        browser_commands.get_website(
            url=command.url, sleep=command.sleep, visit_id=command.visit_id,
            webdriver=webdriver, browser_params=browser_params,
            extension_socket=extension_socket)
```

I hope this helps, if not please follow up.

svensevenslow:
1. I have a confusion regarding how you want me to maintain the crawl_id and visit_id in the parent class. I don't think they need to be maintained as class variables. And according the example of Get_Command you have mentioned above we don't want it in the constructor. So then is this how you want me to maintain it? This will be a method in the parent class. The only problem is that these attributes won't be available until you call this method

```
def set_visit_crawl_id(self, visit_id, crawl_id):
        self.visit_id = visit_id
        self.crawl_id = crawl_id
```
2. And from what I understood above we will be passing the Command object everywhere, starting from the CommandSequence file

vringar:
Hey,
1. I think your approach for this is good. Nobody should access the crawl_id or visit_id before they are set, but that is an assumption that the code already makes, so it should work.
2. Yes, that would be the goal. 
3. I would expect you to implement __repr__ and change those calls to repr(command)  instead of str(command/arguments)
4. Good catch, that should also be a class
5. I was thinking, we would still create a tuple of command and timeout, but if you want to create a field for it feel free.

Also would you mind pushing your intermediate progress, so that I can get a better idea of how far you are in completing this issue.

svensevenslow:
I've pushed all the changes I made. 3 tests seem to be failing. They are failing because of [this](https://github.com/mozilla/OpenWPM/blob/master/test/test_extension.py#L291) and [this](https://github.com/mozilla/OpenWPM/blob/master/test/test_extension.py#L277) failing. Also I think this part 

> implement repr and change those calls to repr(command) instead of str(command/arguments)

I haven't implemented correctly. Let me know what you think.

automation/Commands/Commands.py
```diff
@@ -0,0 +1,63 @@
...
+    def __init__(self, url, sleep):
+        self.url = url
+        self.sleep = sleep
+
+
+class Browse_Command(Command):
+    def __init__(self, url, num_links, sleep):
+        self.url = url
+        self.num_links = num_links
+        self.sleep = sleep
```
vringar:
```suggestion
class Get_Command(Command):
    def __init__(self, url, sleep):
        self.url = url
        self.sleep = sleep
        
     def __repr__(self):
         return "Get_Command({},{})".format(self.url,self.sleep)
```

vringar:
Hey, 
Thank you for sharing your progress. It seems like this is almost done.
I've suggested an implementation of `__repr__` for Get_Command which you should be able to adapt for the other commands.
Do you have an idea why these tests are failing or should I take a look at them tomorrow?

svensevenslow:
One thing I wanted to know is if[command.timeout](https://github.com/mozilla/OpenWPM/pull/580/files#diff-066daebfc15ae59b0a062006d79c53aeR30) is the correct value to pass for start_time  Right now haven't figured out why the tests are failing.

vringar:
I thinks start time should be the current time, as it used to be.
https://github.com/mozilla/OpenWPM/pull/580/files#diff-df517bce823800d647ba110df4679453L374

svensevenslow:
I have made the changes you suggested above. I have set the value of `start_time=None` [here](https://github.com/mozilla/OpenWPM/blob/master/automation/Commands/command_executor.py#L25), as according to the code written [here](https://github.com/mozilla/OpenWPM/blob/master/automation/TaskManager.py#L377) for the DUMP_FLASH_COOKIES command, the start_time value used is None. Not sure if this start_time value is required, I mean do we maybe need to change the function definition given [here ](https://github.com/mozilla/OpenWPM/blob/42fd133bff2b145d4112ff11cd513506099e62f3/automation/Commands/browser_commands.py#L176). If yes, could we handle that by opening a separate issue?

Also the `test_js_time_stamp`[test](https://github.com/mozilla/OpenWPM/blob/master/test/test_extension.py#L277) in text_extension.py failing is the reason for the tests failing in Travis CI. It throws this assertion `assert (utc_now - js_time).seconds < MAX_TIMEDELTA`. Not sure how to go about correcting this issue

svensevenslow:
Hi! I have updated my code with all the requested changes. The [`test_js_time_stamp`  ](https://github.com/mozilla/OpenWPM/blob/master/test/test_extension.py#L277) test in test_extension.py failing is the reason for the Travis CI checks failing. It is throwing the assertion `assert (utc_now - js_time).seconds < MAX_TIMEDELTA`. Not sure how to go about getting this test to pass. Let me know what you think.

automation/TaskManager.py
```diff
@@ -425,13 +423,13 @@ def _issue_command(self, browser: Browser,
...
                     "BROWSER %i: Timeout while executing command, %s, killing "
-                    "browser manager" % (browser.crawl_id, command[0]))
+                    "browser manager" % (browser.crawl_id, repr(command)))
 
             self.sock.send(("crawl_history", {
                 "crawl_id": browser.crawl_id,
                 "visit_id": browser.curr_visit_id,
-                "command": command[0],
-                "arguments": str(command_arguments),
+                "command": repr(command),
```
vringar:
```suggestion
                "command": type(command),
```
Because otherwise you're writing the same string twice

svensevenslow:
I've updated

svensevenslow:
Are there any other changes required to be done?

automation/BrowserManager.py
```diff
@@ -460,7 +462,7 @@ def BrowserManager(command_queue, status_queue, browser_params,
             # (command, arg0, arg1, arg2, ..., argN) where N is variable
```
vringar: Please remove this comment as it is no longer accurate.

automation/BrowserManager.py
```diff
@@ -266,7 +267,8 @@ def close_browser_manager(self):
             return
 
         # Send the shutdown command
-        self.command_queue.put(("SHUTDOWN",))
+        command = Shutdown_Command()
+        self.command_queue.put((command,))
```
vringar:
```suggestion
        self.command_queue.put(command)
```
Because currently the shutdown code doesn't run shutdown code as `type((command,))` is `tuple` and not `Shutdown_Command` as expected

automation/Commands/command_executor.py
```diff
@@ -1,66 +1,74 @@
...
-    if command[0] == 'RUN_CUSTOM_FUNCTION':
+    if type(command) is Run_Custom_Function_Command:
         arg_dict = {"command": command,
                     "driver": webdriver,
                     "browser_settings": browser_settings,
                     "browser_params": browser_params,
                     "manager_params": manager_params,
                     "extension_socket": extension_socket}
-        command[1](*command[2], **arg_dict)
+        command.function_handle(*command.func_args, **arg_dict)
```
vringar: Could you change all of these to `elif`s and throw an exception when an invalid command gets passed?
This way invalid commands can be caught in development instead of just being ignored

automation/TaskManager.py
```diff
@@ -366,19 +367,16 @@ def _issue_command(self, browser: Browser,
...
             command, timeout = command_and_timeout
-            if command[0] in ['GET', 'BROWSE',
-                              'SAVE_SCREENSHOT',
-                              'SCREENSHOT_FULL_PAGE',
-                              'DUMP_PAGE_SOURCE',
-                              'RECURSIVE_DUMP_PAGE_SOURCE']:
+            command.set_visit_crawl_id(browser.curr_visit_id, browser.crawl_id)
+            if isinstance(command, Dump_Flash_Cookies_Command) or isinstance(
+                    command, Dump_Prof_Command):
+                command.set_start_time(start_time)
```
vringar: Could you please set `start_time` to `time.time()` and remove the local variable.
I know this is a change in behavior but the current behavior was incorrect even though nobody noticed it

svensevenslow:
Hi! I've updated all the changes you requested. Currently for throwing an exception I have done a simple `raise Exception("Invalid Command")`, let me know if you wanted this to be done differently, I mean write a custom exception in [Errors.py](https://github.com/mozilla/OpenWPM/blob/master/automation/Errors.py) or use the existing [CommandExecutionError](https://github.com/mozilla/OpenWPM/blob/master/automation/Errors.py#L4) exception? Let me know if there are any other changes required

automation/TaskManager.py
```diff
@@ -425,13 +417,13 @@ def _issue_command(self, browser: Browser,
...
-                    "browser manager" % (browser.crawl_id, command[0]))
+                    "browser manager" % (browser.crawl_id, repr(command)))
 
             self.sock.send(("crawl_history", {
                 "crawl_id": browser.crawl_id,
                 "visit_id": browser.curr_visit_id,
-                "command": command[0],
-                "arguments": str(command_arguments),
+                "command": type(command),
+                "arguments": command_arguments,
```
vringar: @englehardt and I discussed what format would be the easiest to analyze and we came to the assumption, that JSON might just be the better choice than the `repr`
```suggestion
                "arguments": json.dumps(command.__dict__).encode('utf-8),
```

automation/TaskManager.py
```diff
@@ -363,22 +363,14 @@ def _issue_command(self, browser: Browser,
...
-                command += (browser.curr_visit_id,)
-            elif command[0] in ['DUMP_FLASH_COOKIES', 'DUMP_PROFILE_COOKIES']:
-                command += (start_time, browser.curr_visit_id,)
+            command.set_visit_crawl_id(browser.curr_visit_id, browser.crawl_id)
+            command.set_start_time(time.time())
             browser.current_timeout = timeout
             # passes off command and waits for a success (or failure signal)
             browser.command_queue.put(command)
-            command_arguments = command[1] if len(command) > 1 else None
+            command_arguments = repr(command)
```
vringar: And remove this line as it's no longer used.

vringar:
Hey,
I think a `CommandExecutionError` would be a better description of what was going on.
Besides that and the change I suggested, everything LGTM

svensevenslow:
So then I'll be removing the `__repr__` method from the Commands.py file and replacing all `repr(command)` with `json.dumps(command.__dict__).encode('utf-8)`?
Also do you want me to add some error handling for the invalid command or let the outermost [Exception block](https://github.com/mozilla/OpenWPM/blob/master/automation/BrowserManager.py#L508) handle it

vringar:
No, I think the `__repr__` has value in all other places, where we want to have an easily understandable string, that contains as much information as possible. (for logs and debugging)
I just want you to change this one line so that the database entry gets easier to analyse with scripts.
I think the outermost exception block should handle this as there is no reasonable recovery from such a logic error

svensevenslow:
Hi! I've updated witht the changes you requested, the [travis ](https://travis-ci.org/github/mozilla/OpenWPM/builds/666846730?utm_source=github_status&utm_medium=notification)checks are passing

vringar:
LGTM
@skim1102 Do you want to review this as well

PR #581, opened 2020-03-12
Update python dependencies

englehardt:
Updates pinned dependencies.


vringar:
Apparently the current version of iPython is not provided for Python 3.5.
Since I planned on raising the minimum supported version to 3.6 anyways this shouldn't be a problem

vringar:
By merging in the new master we now only run current versions of python aka >3.6

PR #583, opened 2020-03-13
Remove --user flag from pip install

### LINKED ISSUE #577: Running install.sh in a virtual environment fails
The installation instructions in the README suggest running `install.sh` in a virtual environment, but the `--user` flag is used in `install-pip-and-packages.sh` (specifically in the commands `python3 get-pip.py --user` and `pip3 install --user --upgrade -r requirements.txt`).

This results in `install.sh` failing with `ERROR: Can not perform a '--user' install. User site-packages are not visible in this virtualenv.`

Removing the `--user` flag seems to work. What was the rationale behind using this flag in this case? Is it necessary?

boolean5:
This PR resolves #577 by removing the `--user` flag from `pip3 install` and `python3 get-pip.py` commands.

This flag caused `install.sh` and `install-dev.sh` to fail with the following error when run in a virtual environment: `ERROR: Can not perform a '--user' install. User site-packages are not visible in this virtualenv.`

As @englehardt [pointed out](https://github.com/mozilla/OpenWPM/issues/577#issuecomment-598475227), pip now behaves by default as if the `--user` flag was used when running outside of a virtual environment and there is no write permission for the global site-packages directory. Given that, using this flag is no longer necessary.

### DIFF
--- a/install-dev.sh
+++ b/install-dev.sh
@@ -2,4 +2,4 @@
 set -e
 
 # Dependencies for OpenWPM development -- NOT needed to run the platform.
-pip3 install --user -U -r requirements-dev.txt
+pip3 install -U -r requirements-dev.txt
--- a/install-pip-and-packages.sh
+++ b/install-pip-and-packages.sh
@@ -5,8 +5,8 @@ set -e
 # Python requirements are already installed by .travis.yml on Travis
 if [ "$TRAVIS" != "true" ]; then
   wget https://bootstrap.pypa.io/get-pip.py
-  python3 get-pip.py --user
+  python3 get-pip.py
   export PATH=~/.local/bin:$PATH
   rm get-pip.py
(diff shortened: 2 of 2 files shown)


skim1102:
@boolean5, some tests on travis is failing and they are all failing to build the `pyarrow` package on Python 3.8 environment. I think this needs further scrutiny/reading documentation.

vringar:
The issue is the numpy package being out of date. 
@boolean5 please merge the current master into your branch into your branch and see if that works.

boolean5:
@skim1102, as @vringar said, the test failure was not related to the changes in this PR. It was because tests  for Python 3.8 were run but some dependencies were pinned to versions that did not support Python 3.8. Those dependencies have been updated [now](https://github.com/mozilla/OpenWPM/pull/581). I rebased the changes on the current master and it should be alright.

PR #585, opened 2020-03-16
Tagging-Browsers-Changing-Default-Browser-params

Akhila499:


vringar:
Hey, could you please write tests for this.
Testcases that come to mind are:
1. Labels actually work
1. Index overwrites label
2. Throw an error when the user schedules a command_sequence for a label that doesn't exist

You should be able to test all of these by having a command_sequence that contains a custom function in which you make assertions about certain browser parameters being set.
Also please fix all the flake8 errors.

Akhila499:
Hi @vringar @skim1102.. I added test cases and fixed the flake8 errors, please can you check this PR.

PR #589, opened 2020-03-19
Mark visits during shutdown as incomplete

### LINKED ISSUE #443: Make it possible to know whether or not a site visit lead to fully saved visit-related data
In https://github.com/mozilla/openwpm-crawler/issues/23#issuecomment-518279520 we discussed this. 

By for instance saving a flag in crawl_history data_saved = True or making sure that the whole crawl_history record is only saved after all visit-related data is saved, we can be more confident in using that visit-related data in analysis. 

We can also employ a retry-strategy (such as https://github.com/mozilla/openwpm-crawler/issues/23) to incrementally increase the crawl success rate.

vringar:
Closes #443 
Current implementation problems:
- We aren't writing out whether a certain command sequence ran to completion
- We don't have a unique identification per entry in the reddis queue that we persist
- We currently discard the whole batch and not just the actually interrupted crawls

Resulting problems:
- If we have multiple data entries for the same top level domain, we can't just throw the first one away as there might have been other things that weren't captured

Suggestion:
- Create a column in site_visits that contains a unique id (that would have to be generated when inserting into the redis queue)
   This would be pretty easy to do
- Create a field to indicate unsuccessful crawl
  This is really hard as we need to write it out during shutdown


vringar:
@englehardt the code and problems we discussed yesterday

vringar:
Add a was_interrupted flag to the completion queue, and pass that to the callback? Or just drop the callback entirely

vringar:
S3aggregator just drops the last visit on shutdown, so there is no duplicate data

vringar:
TODO:
- [ ]  Fix test check_basic_properties
- [ ] Make "Visit was interrupted" groupable and turn it into a warning

vringar:
The last two open boxes + "S3Aggregator - ERROR - Had unfinished records during relaxed shutdown" should become it's own issue in which I try to draw out a flow chart or state machine for the state of any given record so we can figure out why these are happening
The data saving seems to be very robust but some edge cases might pop up in the future

automation/DataAggregator/BaseAggregator.py
```diff
@@ -127,22 +135,41 @@ def handle_special(self, data: Dict[str, Any]) -> None:
             - finalize: A message sent by the extension to
```
englehardt, line 135: We should also add initialize to this docstring.

automation/DataAggregator/BaseAggregator.py
```diff
@@ -99,9 +103,13 @@ def startup(self):
         self.record_queue = self.sock.queue
 
     def should_shutdown(self):
-        """Return `True` if the listener has received a shutdown signal"""
+        """Return `True` if the listener has received a shutdown signal
+        Sets `self._relaxed` and `self.shutdown_flag`
```
englehardt, line 107: Documenting what self._relaxed indicates would be helpful.

automation/DataAggregator/S3Aggregator.py
```diff
@@ -73,8 +73,12 @@ def __init__(self,
                  manager_params: Dict[str, Any],
                  instance_id: int) -> None:
         self.dir = manager_params['s3_directory']
+
+        def factory_function():
+            return defaultdict(list)
+
         self._records: Dict[int, DefaultDict[str, List[Any]]] =\
```
englehardt, line 80: Is this type correct? Or should it be nested `DefaultDict`?

automation/DataAggregator/parquet_schema.py
```diff
@@ -196,3 +196,10 @@
     pa.field('call_stack', pa.string())
 ]
 PQ_SCHEMAS['callstacks'] = pa.schema(fields)
+
+fields = [
+    pa.field('visit_id', pa.int64(), nullable=False),
+    pa.field('instance_id', pa.uint32(), nullable=False)
+]
+
```
englehardt, line 204:
```suggestion
```

For consistency

automation/TaskManager.py
```diff
@@ -275,34 +277,39 @@ def _launch_aggregators(self) -> None:
...
         """
         Wait for current commands to finish, close all child processes and
         threads
-        <during_init> flag to indicator if this shutdown is occuring during
+
+        Parameters
+        ----------
+        during_init :
+            flag to indicator if this shutdown is occuring during
                       the TaskManager initialization
```
englehardt, line 290:
```suggestion
          the TaskManager initialization
```

automation/TaskManager.py
```diff
@@ -275,34 +277,39 @@ def _launch_aggregators(self) -> None:
...
+        relaxed :
+            If `True` the function will wait for all active
+            `CommandSequences` to finish
```
englehardt, line 293:
```suggestion
            `CommandSequences` to finish before shutting down
```

test/test_s3_aggregator.py
```diff
@@ -64,7 +64,11 @@ def test_basic_properties(self):
 
         # Test visit_id consistency
         visit_ids = defaultdict(set)
-        for table_name in PQ_SCHEMAS:
+        useful_schemas = dict(PQ_SCHEMAS)
```
englehardt, line 67:
```suggestion
        expected_tables = dict(PQ_SCHEMAS)
```

test/test_s3_aggregator.py
```diff
@@ -64,7 +64,11 @@ def test_basic_properties(self):
...
+        # We don't expect incomplete visits to exist
+        # since the visit shouldn't be interrupted
+        useful_schemas.pop("incomplete_visits")
```
englehardt, line 70:
```suggestion
        expected_tables.pop("incomplete_visits")
```

test/test_s3_aggregator.py
```diff
@@ -64,7 +64,11 @@ def test_basic_properties(self):
...
+        for table_name in useful_schemas:
```
englehardt, line 71:
```suggestion
        for table_name in expected_tables:
```

automation/TaskManager.py
```diff
@@ -498,11 +518,18 @@ def _issue_command(self, browser: Browser,
                 browser.restart_required = True
                 self.logger.debug("BROWSER %i: Browser restart required" % (
                     browser.crawl_id))
+
             else:
                 with self.threadlock:
                     self.failurecount = 0
 
             if browser.restart_required:
+                self.sock.send((RECORD_TYPE_SPECIAL, {
```
englehardt, line 527: Note that `browser.restart_required` is also set to True by the memory monitor, which can lead to bugs and might be the cause of the duplicate records you've noticed. See:
https://github.com/mozilla/OpenWPM/blob/9c8e2ea8678da2d0916bb7d9a4855341b20a9bae/automation/TaskManager.py#L209-L232

The browser may successfully complete the last command, but also run over the memory limit at the same time. The extension will send back success and then this code will send back failure (which is incorrect, the visit *did* finish). We can mitigate this by having the memory monitor use a different browser property (e.g., `browser.memory_exceeded`) and perhaps `restart_required` can then just be a local variable.

englehardt:
> The last two open boxes + "S3Aggregator - ERROR - Had unfinished records during relaxed shutdown" should become it's own issue in which I try to draw out a flow chart or state machine for the state of any given record so we can figure out why these are happening
> The data saving seems to be very robust but some edge cases might pop up in the future

Do you know how common this is? If relatively uncommon we should definitely not block on it.

automation/TaskManager.py
englehardt: Also is finalize command always sent twice, or only sometimes?

automation/CommandSequence.py
```diff
@@ -42,6 +43,12 @@ def __init__(self, url: str, reset: bool = False,
         site_rank : int, optional
             Integer indicating the ranking of the page to visit, saved
             to `site_visits`
+        callable :
+            A callback to be invoked once all data regarding this
+            CommandSequence has been saved out or it has been interrupted
+            The function will be passed a bool that's true if everything has
+            been successfully saved out if it's false something bad happened
+            and no or only incomplete data may have been saved out
```
vringar, line 51:
```suggestion
            If the command sequence completes and all data is saved successfully, `True` will be passed to the callback.
            Otherwise `False` will be passed. A value of `False` indicates that the data saved from the site visit may be 
            incomplete or empty.
```

automation/DataAggregator/BaseAggregator.py
```diff
@@ -99,9 +103,13 @@ def startup(self):
...
+        """
```
vringar, line 108:
```suggestion
        `self._relaxed means this shutdown is happening after all visits have completed and all data can be seen as
        complete
        """
```

automation/TaskManager.py
vringar: The finalize command is only very rarely called twice. And it only happened after the crawl had been running for a while. I should investigate if it relates to the memory watchdog

vringar: Also if we have a false positive (as in mark a visit incomplete when it's not) it isn't a big deal as the job will just retry and run to completion

Issue #565, opened 2020-02-16
Remove is_full_page and is_frame_load in http_requests


eriktews:
In the current version of OpenWPM, is_full_page and is_frame_load are used a bit inconsistently. The documentation on https://github.com/mozilla/OpenWPM/wiki/Instrumentation-Schema-Documentation#http-requests says:

is_frame_load - True if the request is for an iframe document
is_full_page - True if the request is for a top-level (i.e. tab) document

However, the code that determines those attributes on https://github.com/mozilla/OpenWPM/blob/master/automation/Extension/webext-instrumentation/src/background/http-instrument.ts#L357 sets is_full_page to true for every request that is executed on behalf of the top level frame such as for example loading an image for a site that has no frame/iframes. However, is_frame_load is only set to true for requests that populate an entire frame/iframe with content, so in general only for requests that are supposed to return an HTML document. When an image is loaded from within an iframe, the is_frame_load attribute is false.

From my point of view, this is not very consistent, so I would suggest to do one of the following things:
- Either update the documentation to clearly state that what exactly the meaning of is_frame_load and is_full_page is according to what is currently being done in the source code.
- Update the source code to mark only those requests that load the top of an frame/iframe true here which means checking whether the type of the request is either main_frame or sub_frame according to https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/webRequest/ResourceType
- Update the source code so that all requests that are executed on behalf of the top frame are marked as is_full_page (frameId === 0) and those which are executed on behalf of a frame/iframe are marked as is_frame_load (frameId !== 0)
- Introduce two new fields in the database (for example is_for_full_page, is_for_frame), which takes the values from the previous item.

I'm happy to submit a patch for each of the variants, but first I would like to hear which variant is best.

englehardt:
Thanks for the thorough submission!

> However, the code that determines those attributes on https://github.com/mozilla/OpenWPM/blob/master/automation/Extension/webext-instrumentation/src/background/http-instrument.ts#L357 sets is_full_page to true for every request that is executed on behalf of the top level frame such as for example loading an image for a site that has no frame/iframes.

That's an error. The intended functionality is what's in the documentation.

>However, is_frame_load is only set to true for requests that populate an entire frame/iframe with content, so in general only for requests that are supposed to return an HTML document. When an image is loaded from within an iframe, the is_frame_load attribute is false.

That's expected, and matched the documentation.

[...]

PR #590, opened 2020-03-19
removed iframe/page type check from HTTP requests

LilyLME:
Removed is_full_page and is_frame_load in http_requests as prescribed in  #565

### DIFF
--- a/automation/DataAggregator/parquet_schema.py
+++ b/automation/DataAggregator/parquet_schema.py
@@ -59,8 +59,6 @@ fields = [
     pa.field('headers', pa.string(), nullable=False),
     pa.field('request_id', pa.string(), nullable=False),
     pa.field('is_XHR', pa.bool_()),
-    pa.field('is_frame_load', pa.bool_()),
-    pa.field('is_full_page', pa.bool_()),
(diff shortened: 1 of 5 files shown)


LilyLME:
@englehardt I bet we have to take that off the documentation as well?

englehardt:
> @englehardt I bet we have to take that off the documentation as well?

Yes, thanks for the reminder. I can do that when we merge.

I think you're almost there! You just have a few test failures to fix (see: https://travis-ci.org/github/mozilla/OpenWPM/jobs/664536322#L2602). The problem here is that the hard coded records expected by the test page still contain values for the `is_full_page`/`is_frame_load` columns. You'll need to remove those as well.

LilyLME:
Got the tests to pass! After a while of missing just one uncorrected row in the HTTP_REQUEST dict. {} :/

Screenshot from 2020-03-22 02-21-54

LilyLME:
@englehardt Please do have a look again.

englehardt:
This looks good to me, thank you!

englehardt:
I've updated the wiki as well.

englehardt:
@skim1112 do you also want to look this over?

LilyLME:
Great.

Issue #550, opened 2020-01-07
Create a script folder and create one main entry script


vringar:
There should only be one `setup.sh`/`install.sh` in the top level directory.
It should automatically detect arch and have a command line argument so you can run the dev setup as well ( `./install.sh dev`). This would allow us to more easily support multiple distributions and clean up the top level directory.

vringar:
Good first issue because it's only tangentially related to the project and only requires knowledge of bash

LilyLME:
Hello @vringar please assign this to me. :) I would like to work on it

englehardt:
@LilyLME done! Thank you.

PR #592, opened 2020-03-21
added scripts folder and cmd line arg for dev requirments

LilyLME:
As discussed in #550 added a scripts folder and move main scripts into scripts folder.


LilyLME:
## Discussion

- Some people have higher versions of node installed, can we made the node script running optional?
- Should the docker script be move from root dir?
- Can `install-system.sh` also be optional?

vringar:
- In `install-node.sh`we should check if node is installed and not touch their version
- No, the docker script should remain in the top level directory
- No, it's not, as there would be too much to check for and also people might want to rerun the script to update their stuff

vringar:
We've now decided to move our installation over to Conda since it provides proper multiplattform. Thank you very much anyways.

PR #593, opened 2020-03-22
Remove support for flash cookie saving

svensevenslow:


svensevenslow:
@englehardt I've updated my PR

Issue #501, opened 2019-10-08
Memory watchdog monitors the geckodriver process instead of the Firefox process


englehardt:
The [memory watchdog](https://github.com/mozilla/OpenWPM/blob/960fc5040cbeb188310fd89c8b7b93d7a8e35989/automation/TaskManager.py#L199-L211) in the `TaskManager` is meant to monitor the memory usage of the Firefox and decide whether the process should be restarted. Since the update to a multi-process version of Firefox, this instead monitors the memory consumption of the geckodriver process since this is the process that Selenium [returns a handle to](https://github.com/mozilla/OpenWPM/blob/960fc5040cbeb188310fd89c8b7b93d7a8e35989/automation/DeployBrowsers/deploy_firefox.py#L201-L207). To restore the intended functionality, we should instead traverse the process chain and sum the total memory used by the main Firefox process and all of its child processes.

Here's some sample code on how to do that when testing with `python -m test.manual_test --selenium`:

```python
import psutil

geckodriver = psutil.Process(pid=driver.service.process.pid)
firefox = geckodriver.children()[0]
for child in firefox.children():
    print(child.memory_info()[0] / float(2**20))
```

boolean5:
Hi, I'd like to work on this. May I?

vringar:
Go right ahead :)

boolean5:
Hello again. I submitted a PR along with a few questions and comments here: https://github.com/mozilla/OpenWPM/pull/596

PR #596, opened 2020-03-25
Restore intended functionality of memory watchdog

boolean5:
This PR fixes https://github.com/mozilla/OpenWPM/issues/501 by making the `_manager_watchdog` method in the `TaskManager` monitor the memory consumption of the Firefox process and all its child processes instead of that of the geckodriver process.

### DIFF
--- a/automation/TaskManager.py
+++ b/automation/TaskManager.py
@@ -207,8 +207,17 @@ class TaskManager:
             # Check browser memory usage
             for browser in self.browsers:
                 try:
-                    process = psutil.Process(browser.browser_pid)
-                    mem = process.memory_info()[0] / float(2 ** 20)
+                    # Sum the memory used by the geckodriver process, the
+                    # main Firefox process and all its child processes.
+                    # Use the USS metric for child processes, to avoid
+                    # double-counting memory shared with their parent.
+                    geckodriver = psutil.Process(browser.browser_pid)
+                    firefox = geckodriver.children()[0]
+                    mem_bytes = (geckodriver.memory_info().rss
+                                 + firefox.memory_info().rss)
+                    for child in firefox.children():
+                        mem_bytes += child.memory_full_info().uss
+                    mem = mem_bytes / 2 ** 20
                     if mem > BROWSER_MEMORY_LIMIT:
                         self.logger.info("BROWSER %i: Memory usage: %iMB"
                                          ", exceeding limit of %iMB" %
--- a/setup.cfg
+++ b/setup.cfg
@@ -6,4 +6,4 @@ skip = venv,automation/Extension,firefox-bin
 
 [flake8]
 exclude = venv,automation/Extension,firefox-bin
-ignore = Q
+extend-ignore = Q


boolean5:
@vringar, @englehardt, I have a few questions and comments on this:

1. Is it safe to assume that the main Firefox process will always be at `geckodriver.children()[0]`?

2. Could you please explain how the memory watchdog worked before, when Firefox run as a single process? Wasn't it still monitoring the memory used by the geckodriver process?

3. I removed the redundant [type conversion to float](https://github.com/mozilla/OpenWPM/blob/960fc5040cbeb188310fd89c8b7b93d7a8e35989/automation/TaskManager.py#L203), since OpenWPM is a Python 3 application and in Python 3 the `/` operator does not do floor division.

4. I initially wrote the memory computation as follows:
  ```
  mem = (firefox.memory_info().rss + sum(
         child.memory_info().rss for child in
         firefox.children())) / 2 ** 20
  ```
  However, I chose to use a for loop instead of `sum` because the resulting code felt somewhat more readable. If you prefer this version I'm ok with changing it back.

5. I noticed the following possibly problematic [check](https://github.com/mozilla/OpenWPM/blob/960fc5040cbeb188310fd89c8b7b93d7a8e35989/automation/TaskManager.py#L224-L225) in the process_watchdog: 
`(process.name() == 'firefox' and (process.pid not in browser_pids))`
  I think that this is always going to be true, since the `browser_pids` list only holds pids of geckodriver processes. Is that right? Even though the process_watchdog is currently disabled it might be useful to keep track of this for future reference.

vringar:
1. It's pretty safe to assume that this is the case, but I don't see any harm done in monitoring all child processes of the Geckodriver as well as gecko_driver itself.
2. AFAIK there was no Geckodriver when Firefox was single process as controlling the browser was done with an extension. So the name `browser_pid` was actually correct. Whereas nowadays it should probably be called `geckodiver_pid`.
3. Thank you for that :)
4. I think the for loop is more readable
5. Changing the check is definitely a good idea. I do wonder though if it's necessary to kill everything recursively or if we could just call terminate on the geckodriver hoping that it'll clean up it's firefox process by itself. @englehardt do you have an opinion on this?

Also just out of curiosity: Do you think we could just recursively sum the memory of all child processes starting from geckodriver? (maybe also more for @englehardt )

This way we would be robust against process layout changes that Firefox might have in the future.
```python 
def memory_consumption(parent: psutil.Process) -> int: 
 """ Returns the sum of the memory consumption 
   of the given process and all it's children""" 
   memory = parent.memory_info().rss 
   for child in parent.children(recursive=True): 
       memory += child.memory_info().rss
  return memory 
```

boolean5:
> Do you think we could just recursively sum the memory of all child processes starting from geckodriver?

Yes, to me that makes even more sense.

To be clear, do you suggest implementing the memory consumption computation as a separate `memory_consumption()` function or was the code snippet above just an example? In the first case, where should `memory_consumption()` be implemented? In `automation/utilities/platform_utilities.py` or within the  `_manager_watchdog()` function since it is not likely to be used elsewhere?

 Also, regarding 5, changing the check to `process.name() == 'geckodriver'`would also get rid of the unwanted side-effect of killing random firefox instances that happen to be open during a crawl but have nothing to do with OpenWPM. In any case, this is not directly related to this PR and shouldn't block it.

vringar:
Then we should probably implement it that way.

This is just a snippet I wrote when trying to figure out how psutil works. I don't think it should be a separate function.

Yes, changing the check is definitely desirable and I'd welcome it in this PR.

boolean5:
I changed the memory computation code to recursively sum the memory of geckodriver and all its descendants.

> I do wonder though if it's necessary to kill everything recursively or if we could just call terminate on the geckodriver hoping that it'll clean up it's firefox process by itself.

This does not seem to happen. I tested it with `python -m test.manual_test --selenium` and the following:
```python
import psutil

geckodriver = psutil.Process(pid=driver.service.process.pid)
geckodriver.status()
firefox.status()

geckodriver.terminate()
geckodriver.wait()

geckodriver.status()
firefox.status()
```

The second time `geckodriver.status()` is run it raises a `NoSuchProcess` exception, while `firefox.status()` prints `sleeping`, just like before.


> Yes, changing the check is definitely desirable and I'd welcome it in this PR.

I can change the check in this PR but it is not enough to restore the functionality of the process watchdog. I'm still not sure which is the correct/preferred way to kill/terminate the process tree under geckodriver. So unless this turns out to be something trivial, I was thinking about opening a separate issue to discuss the changes required to bring the currently broken process watchdog back to life. What do you think?

vringar:
@skim1102 waiting with merging for your review

Thank you for investigating my suggestion. Seeing as the naive approach didn't work I'd agree with you that a new issue should be opened for the process watchdog.

boolean5:
Great! I opened an issue for the process watchdog here: https://github.com/mozilla/OpenWPM/issues/598

englehardt:
>     5. Changing the check is definitely a good idea. I do wonder though if it's necessary to kill everything recursively or if we could just call terminate on the geckodriver hoping that it'll clean up it's firefox process by itself. @englehardt do you have an opinion on this?

I believe we need to kill the child processes first. I remember trying to just kill the geckodriver process when we first upgraded to Selenium 3 and running into issues. We actually already have code to do this here:

https://github.com/mozilla/OpenWPM/blob/2d689772a82f1b2115a9e81bb330857a38cfa504/automation/BrowserManager.py#L326-L352
 
> Also just out of curiosity: Do you think we could just recursively sum the memory of all child processes starting from geckodriver? (maybe also more for @englehardt )

Sounds good

@boolean5 I really appreciate you working on this!

boolean5:
> Side note, could there be a chance that this sum is double counting memory as seen in shared attribute for memory_info() documentation?

I think @skim1102 is right. I'll try to investigate this further. Please don't merge yet.

boolean5:
Ok, here's what I found:

1. [This blogpost](http://www.erahm.org/2016/02/11/memory-usage-of-firefox-with-e10s-enabled/) suggests a more realistic approximation of the total memory used by multi-process Firefox. This is achieved by summing the RSS of the Firefox main process and the USS (Unique Set Size) of each child process. The USS is the memory which is unique to a process and is obtained via `memory_full_info()`. The [psutil documentation](https://psutil.readthedocs.io/en/latest/#psutil.Process.memory_full_info) for this method states the following:

	a. it usually requires higher user privileges than `memory_info()`
	b. it is considerably slower than `memory_info()`. While running `python -m test.manual_test --selenium`, timing 1000 repetitions of accessing the geckodriver process's `memory_full_info().uss` and `memory_info().rss` showed a ~10x difference (the respective results where 0.0381 sec and 0.3978 sec). Taking into account that the memory watchdog runs every 10 seconds is this an acceptable tradeoff?	

      In addition, this way of computing the total memory ignores memory that is only shared between child processes. Do Firefox child processes share memory between themselves?

2. Another approximation of the total memory usage could be obtained using `memory_full_info().pss` for all Firefox processes. According to the documentation, this divides shared memory evenly between the processes that share it. However, `memory_full_info().pss` is only supported on Linux.

3. Summing the RSS of all processes and subtracting the `memory_info().shared` memory for each child process could also provide an approximation. Unfortunately, `memory_info().shared` is only supported on Linux.

4. If accuracy is not very important in this scenario, and given that solutions 1,2 and 3 all have major drawbacks, we could stick to using RSS. In case we decide to do this, I should add a comment in the code that will highlight this inaccuracy.

Running a simple experiment using `python -m test.manual_test --selenium` gave the following results for each of the computation methods described above. These results can be compared to the difference in available memory before and after terminating the firefox and geckodriver processes as reported by `free -m` , which was 210 MiB.

```
Method | Total Memory (MiB)
  1	    295.56	
  2	    281.81
  3	    293.61
  4	    499.08
```

Please, let me know what you think.

vringar:
I'd be in favor of method 1 as long as we can figure out what higher user privileges means (my main question being "can we use it in docker?") since speed is not really an issue as this is not happening on a worker thread. 
Even if this is not entirely accurate it's a way better approximation than our naive approach and has the advantage of being cross platform.

boolean5:
I think higher user privileges are required only if we want to use `memory_full_info()` on a process owned by another user. So there shouldn't be a problem as long as the Firefox processes and the process that runs the memory watchdog are all owned by the same user. Isn't that the case?

vringar:
Since OpenWPM spawns the GeckoDriver process and the GeckoDriver spawns the Firefox process I think we should be in the clear.
So please feel free to go ahead with method 1.

boolean5:
Thanks for confirming this! I changed the code to use method 1.

boolean5:
Regarding the test failure: flake8 produces a warning both when a line break occurs before the binary operator and when it occurs after it (W503 and W504 respectively), which is conflicting behavior. According to the output of `flake8 --help`, the default configuration ignores these warnings. But because we use the "ignore" option in `setup.cfg`, the default configuration is overriden (see the [documentation](https://flake8.pycqa.org/en/latest/user/violations.html#ignoring-violations-with-flake8) of flake8 for more information on this).
I can add a commit to replace `ignore = Q` with `extend-ignore = Q` in `setup.cfg` if you agree.
Alternatively, we can ignore one of these warnings (preferrably W503 to follow the pep8 [recommendation](https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator)).

skim1102:
@vringar, @birdsarah was there a reason we used `ignore` instead of `extend-ignore` blocking @boolean5 to make the change?

@boolean5, if you replace the `ignore` to `extend-ignore` does it raise other warnings outside the changes in this PR?

vringar:
As per git-blame this line was added by @englehardt, but as per his commit message back then I don't see a reason not to switch to `extend-ignore`

boolean5:
@skim1102, no, changing  `ignore` to `extend-ignore` doesn't raise any other warnings.

boolean5:
I've updated the PR with a commit that makes `setup.cfg` extend the default flake8 configuration instead of overriding it.

PR #599, opened 2020-04-01
Fixing crawler

vringar:
The code I had in `crawler.py` wasn't capturing the variables as expected (it only bound the first job) and this PR fixes it.
It also adds more logging to entire callback chain that should help debugging.


crawler.py
```diff
@@ -111,10 +112,23 @@
 
 unsaved_jobs = list()
 unsaved_jobs_lock = Lock()
+
+
+def mark_job_done(logger: logging.Logger, unsaved_jobs_lock: Lock,
```
englehardt, line 117: I think it's confusing to still have this called `mark_job_done`, since it implies that calling this function marks the job "done". Can we change to `get_job_completion_callback` or something along those lines?

crawler.py
```diff
@@ -111,37 +115,60 @@
...
-                                         2 * (TIMEOUT + DWELL_TIME + 30)):
+                                         EXTENDED_LEASE_TIME):
                 manager.logger.error("Unsaved job: %s timed out", unsaved_job)
 
     job = job_queue.lease(
         lease_secs=TIMEOUT + DWELL_TIME + 30, block=True, timeout=5
     )
     if job is None:
-        manager.logger.info("Waiting for work")
+        if no_job_since is None:
```
vringar, line 147: This is needed since every job doesn't get saved until a new one started, so the last job of each browser stays unfinished until `close` gets called, however `close` can't get called as there are still jobs in `:processing` so the job_queue isn`t empty.
I attempt to resolve this deadlock by waiting for the extended lease time, thus ensuring all unfinished jobs are being held by other instances or ourselves and thus implying that the job didn't fail but everybody is just waiting on each other, each instance then chooses to terminate itself to resolve this deadlock and since they all call `close` everything gets saved out.

vringar: @englehardt I can write up a longer form piece on this if you want me to.
This feels very hacky and I don't like it, but it's the best I came up with.

vringar:
@nhnt11 if you have the time please comment on the changes I made to the extension

automation/Extension/firefox/feature.js/loggingdb.js
```diff
@@ -134,18 +168,18 @@ export let dataReceiver = {
...
-    while (!debugging && listeningSocket.queue.length != 0) {
-        visitID = listeningSocket.queue.shift();
-        logDebug("Visit Id: " + visitID);
-    }
-    record["visit_id"] = parseInt(visitID, 10);
-
+    record["visit_id"] = visitID;
 
     if (!visitID && !debugging) {
+        // We navigate to about blank between visits and it screws with our data
```
englehardt, line 174:
```suggestion
        // Navigations to about:blank can be triggered by OpenWPM. We drop those.
```

automation/DataAggregator/LocalAggregator.py
```diff
@@ -84,23 +85,26 @@ def _generate_insert(self, table, data):
...
     def process_record(self, record: Tuple[str, Union[str, Dict[str, Any]]]):
         """Add `record` to database"""
+
         if len(record) != 2:
             self.logger.error("Query is not the correct length")
             return
 
         table, data = record
-
         if table == "create_table":
```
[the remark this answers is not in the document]
vringar: Seems reasonable :)

PR #600, opened 2020-04-03
Update dependencies

englehardt:
I've had to add:

```
# localstack compatibility
more-itertools==5.0.0
zipp==0.6.0
```

due to localstack's outdated pinned dependencies

and we already pin `psutil` for the same reason. 

We should investigate what it would take to run localstack as a docker image to avoid these issues in the future, since the current installation approach we use [is deprecated](https://github.com/localstack/localstack/issues/2019#issuecomment-582581192).


englehardt:
This needs to be rebased on #593. @vringar you can hold off on reviewing for now.

englehardt:
This is now ready for review

automation/Extension/webext-instrumentation/package.json
```diff
@@ -64,7 +64,7 @@
     "@types/firefox-webext-browser": "^63.0.0",
```
vringar, line 64: Could we bump this to [latest version](https://www.npmjs.com/package/@types/firefox-webext-browser) 
```suggestion
    "@types/firefox-webext-browser": "^70.0.1",
```

englehardt: Done!

Issue #594, opened 2020-03-23
sqlite3.ProgrammingError: Cannot operate on a closed database in TestSimpleCommands.test_browse_wrapper_http_table_valid


englehardt:
See https://travis-ci.org/github/mozilla/OpenWPM/jobs/664956158#L3994

boolean5:
@englehardt, the above link points to a successful test run. I've seen this error appear in `TestExtension.test_canvas_fingerprinting` here though: https://travis-ci.org/github/mozilla/OpenWPM/jobs/664047781#L1862

LilyLME:
This test happens to pass locally for me.

englehardt:
Closing for now. We can file a new issue if we see this again.

PR #603, opened 2020-04-07
Fix some test warnings

boolean5:
I've seen these two warnings when running the tests for a while now:
1. https://travis-ci.org/github/mozilla/OpenWPM/jobs/671679677#L2939
2. https://travis-ci.org/github/mozilla/OpenWPM/jobs/671679677#L2948

Regarding the first warning, when `pytest.mark.slow` was added, it was perhaps assumed that it is one of the built-in marks of pytest, when it is in fact a custom mark. Registering it in the `pytest.ini` file, as mentioned in the [pytest documentation](https://docs.pytest.org/en/latest/mark.html#registering-marks) fixes the `PytestUnknownMarkWarning`. 

To address the second warning (`DeprecationWarning`), I just changed the deprecated `warn` method to `warning`.

### DIFF
--- a/automation/TaskManager.py
+++ b/automation/TaskManager.py
@@ -380,7 +380,7 @@ class TaskManager:
 
         reset = command_sequence.reset
         if not reset:
-            self.logger.warn(
+            self.logger.warning(
                 "BROWSER %i: Browser will not reset after CommandSequence "
                 "executes. OpenWPM does not currently support stateful crawls "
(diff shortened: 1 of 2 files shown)


boolean5:
Seems like the test error is not related to the changes in this PR. It is similar to the intermittent error mentioned in https://github.com/mozilla/OpenWPM/issues/594.

vringar:
Thank you very much for this PR.

PR #607, opened 2020-04-10
Update to FF75, including new setting.

### LINKED ISSUE #606: OpenWPM does not run with FF74 or higher
Tried upgrading version of FF installed and hit issues. FF 73_0_1 (TAG=5b0905233e5d29d9fbad98c1380030387cda5dd9) works 

But FF 74, 74_0_1, 75 all fail with the same error:

```
BrowserManager       - INFO     - BROWSER 1: Launching browser...
BrowserManager       - ERROR    - BROWSER 1: Crash in driver, restarting browser manager
  File "/home/bird/Dev/mozilla/OpenWPM/automation/BrowserManager.py", line 436, in BrowserManager
    with open(ep_filename, 'rt') as f:
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/rust_mozprofileNEhYpP/extension_port.txt'
BrowserManager       - ERROR    - BROWSER 1: Spawn unsuccessful  | Proxy Ready: False  | Profile Created: True  | Profile Tar: True  | Display: False  | Launch Attempted: True  | Browser Launched: True
| Browser Ready: False
```

FF74 release notes (https://www.mozilla.org/en-US/firefox/74.0/releasenotes/) says "Add-ons installed by external applications can now be removed using the Add-ons Manager (about:addons). Going forward, only users can install add-ons; they cannot be installed by an application."

birdsarah:
Fixes #606


englehardt:
Interesting. It looks like two lines of the stacktrace have been left off, causing that test to fail. @vringar may have some insight on that.

birdsarah:
These tests have run fine for me locally. I think this was just an intermittent failure caused by the browser / page loading slowly.

birdsarah:
This appears to be an intermittent test failure, I don't have permission to restart the build.

birdsarah:
cc @englehardt @vringar

test/test_callstack_instrument.py
```diff
@@ -63,25 +63,25 @@ def test_http_stacktrace(self):
...
-                    or url.endswith("Blank.gif")):  # noqa: W503
-                observed_records.add(stacktrace)
+            print(row["call_stack"])
+            url, call_stack = row
+            test_urls = (
+                "inject_pixel.js",
+                "test_image.png",
+                "Blank.gif",
+            )
+            if url.endswith(test_urls):
```
englehardt, line 83: Cool! I didn't know `endswith` accepts a set

PR #621, opened 2020-04-28
Restore xvfb

### LINKED ISSUE #504: Restore XVFB-based headless mode
By moving from XVFB to Firefox's built-in headless mode in #426 we no longer properly support WebGL. WebGL is likely not the only web API that doesn't work correctly under headless mode. This means that OpenWPM is even more detectable as a bot (see #448), but this also means that we won't be able to properly measure WebGL and may end up messing with the many fingerprinting scripts that use WebGL.

There's no need to remove the support for Firefox's built-in headless mode, but we should add back in support for XVFB under a different configuration parameter. The code removed in #426 is a good place to start.

birdsarah:
Fixes #504

### DIFF
--- a/README.md
+++ b/README.md
@@ -300,6 +300,12 @@ left out of this section.
     * Launch the browser using the X virtual frame buffer. In this mode, Firefox
       is not running in it's own headless mode, but no GUI will be displayed.
     * This mode only works on Linux.
+  * `headless` mode and `xvfb` are not equivalent. `xvfb` is a full browser, but you get
+    "headless" browsing because you do not need to be in a full X environment e.g. on a
+    server. `headless` mode is supported on all platforms and is implemented by the browser
+    but has some differences. For example webGL is not supported in headless mode.
+    https://github.com/mozilla/OpenWPM/issues/448 discusses additional factors to consider
+    when picking a `display_mode`.
 * `browser`
   * Used to specify which browser to launch. Currently only `firefox` is
     supported.
--- a/README.md
+++ b/README.md
@@ -288,11 +288,18 @@ left out of this section.
   * Performs some actions to prevent the platform from being detected as a bot.
   * Note, these aren't comprehensive and automated interaction with the site
     will still appear very bot-like.
-* `headless`
-  * Launch the browser in headless mode (supported as of Firefox 56),
-    no GUI will be visible.
-  * Use this when running browsers on a remote machine or to run crawls in the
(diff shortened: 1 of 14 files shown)


birdsarah:
I think we could parametrize one or more tests to run in "headless" and "xvfb" mode. I'm open to suggestions about which tests it would be good to do?

birdsarah:
`test_crawl/TestCrawl` ?

test/test_simple_commands.py
```diff
@@ -328,25 +351,25 @@ def test_save_screenshot_valid(self, tmpdir):
         manager.close()
 
         # Check that viewport image is not blank
-        pattern = os.path.join(str(tmpdir), 'screenshots', '1-*-test.png')
+        pattern = os.path.join(str(self.tmpdir), 'screenshots', '1-*-test.png')
```
birdsarah, line 354: Note that I'm just making use of an existing feature of `OpenWPMTest` class that sets the tmpdir as a class property.

README.md
```diff
@@ -288,11 +288,18 @@ left out of this section.
...
+    * Launch the browser normally - GUI will be visible
+  * `headless`:
+    * Launch the browser in headless mode (supported as of Firefox 56),
+        no GUI will be visible.
+    * Use this when running browsers on a remote machine or to run crawls in the
+        background on a local machine.
+  * `xvfb`:
+    * Launch the browser using the X virtual frame buffer. In this mode, Firefox
+      is not running in it's own headless mode, but no GUI will be displayed.
+    * This mode only works on Linux.
```
englehardt, line 302: We should mention here _why_ someone might want to run in `xvfb`. Maybe:

```
Firefox running in xvfb may more accurately reflect a real computing
environment than Firefox's native headless mode (e.g., for reasons like 
[Bug 1375585](https://bugzilla.mozilla.org/show_bug.cgi?id=1375585)).
Use this mode if `headless` mode doesn't support an API you'd like to 
monitor, or if you're concerned about bot detection.
```

birdsarah: I think many aspects of bot detection will be the same / alternate problems in xvfb on a server vs headless mode somewhere else. I will add a note about the problems with headless, and open an issue to document the relative merits of xvfb vs headless and we can discuss there.

englehardt: Thanks. That's fine as well. I mainly want to make it possible for someone without context to decide between the two modes based on the differences we currently know about.

PR #630, opened 2020-05-04
Dockerfile now has a develop stage

vringar:
This change allows people to run `docker build --target develop -t openwpm_develop . ` to build a dockerfile that has the dev-dependencies installed.
This enables tests to be run in the dockerfile if for some reason (be it that your developing on Windows or that your local installation of localstack is broken) you can't run the tests natively.
Then you could run
```bash
docker run --rm --name crawl  -v $PWD/docker-volume:/root/Desktop -v $PWD:/opt/OpenWPM \
 -it openwpm_develop
```
to mount the current python code and test it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,7 +16,7 @@ RUN npm run build
 RUN cp dist/openwpm-1.0.zip openwpm.xpi
 
 # Stage 2, build the main OpenWPM image
-FROM krallin/ubuntu-tini:bionic
+FROM krallin/ubuntu-tini:bionic as release
 
 WORKDIR /opt/OpenWPM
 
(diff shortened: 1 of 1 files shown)


englehardt:
> This change allows people to run `docker build --target develop -t openwpm_develop . ` to build a dockerfile that has the dev-dependencies installed.
> This enables tests to be run in the dockerfile if for some reason (be it that your developing on Windows or that your local installation of localstack is broken) you can't run the tests natively.
> Then you could run
> 
> ```shell
> docker run --rm --name crawl  -v $PWD/docker-volume:/root/Desktop -v $PWD:/opt/OpenWPM \
>  -e DISPLAY=$DISPLAY -v /tmp/.X11-unix:/tmp/.X11-unix --shm-size=2g \
>  -it openwpm_develop
> ```
> 
> to mount the current python code and test it.

Thanks! Would you mind to add some instructions README as part of this PR? I'll re-review once that is added.

birdsarah:
I think this PR is no longer necessary, as dev dependencies are installed always with #648.

PR #631, opened 2020-05-04
Save the HTTP response headers for HTTP redirects

eriktews:
Currently, the HTTP response headers of the server are not logged when
the server responds with an HTTP redirect. Wit this patch, they are
saved as well.

I can add a test for this as well, but when I would test for the exact headers present, then it might break easily once the HTTP server for the tests is upgraded.

### DIFF
--- a/automation/DataAggregator/parquet_schema.py
+++ b/automation/DataAggregator/parquet_schema.py
@@ -100,6 +100,7 @@ fields = [
     pa.field('frame_id', pa.int64()),
     pa.field('response_status', pa.int64()),
     pa.field('response_status_text', pa.string(), nullable=False),
+    pa.field('headers', pa.string()),
     pa.field('time_stamp', pa.string(), nullable=False)
 ]
 PQ_SCHEMAS['http_redirects'] = pa.schema(fields)
--- a/automation/DataAggregator/schema.sql
+++ b/automation/DataAggregator/schema.sql
@@ -120,6 +120,7 @@ CREATE TABLE IF NOT EXISTS http_redirects(
   frame_id INTEGER,
   response_status INTEGER NOT NULL,
   response_status_text TEXT NOT NULL,
+  headers TEXT NOT NULL,
   time_stamp DATETIME NOT NULL
 );
 
--- a/automation/Extension/webext-instrumentation/src/background/http-instrument.ts
+++ b/automation/Extension/webext-instrumentation/src/background/http-instrument.ts
@@ -6,6 +6,7 @@ import { PendingResponse } from "../lib/pending-response";
 import ResourceType = browser.webRequest.ResourceType;
 import RequestFilter = browser.webRequest.RequestFilter;
 import BlockingResponse = browser.webRequest.BlockingResponse;
+import HttpHeaders = browser.webRequest.HttpHeaders;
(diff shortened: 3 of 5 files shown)


automation/DataAggregator/schema.sql
```diff
@@ -120,6 +120,7 @@ CREATE TABLE IF NOT EXISTS http_redirects(
   frame_id INTEGER,
   response_status INTEGER NOT NULL,
   response_status_text TEXT NOT NULL,
+  headers TEXT NOT NULL,
```
englehardt, line 123: It would be great to have this in in the responses table, rather than redirects. The headers are actually a property of the response, and from your description it sounds like the column is otherwise empty when there are redirects. Is that right?

However, I suspect this isn't straightforward since the actual data is available in the redirect listener, and you'd need to save it alongside data saved in a separate response listener. If I'm missing something and you see a straightforward way to do this please do. Otherwise, there's no need to add complexity here.

englehardt:
> Currently, the HTTP response headers of the server are not logged when
> the server responds with an HTTP redirect. Wit this patch, they are
> saved as well.
> 
> I can add a test for this as well, but when I would test for the exact headers present, then it might break easily once the HTTP server for the tests is upgraded.

A test would be great! You don't have to test all of the details of the headers. You can check to make sure that the expected number of headers are present with and without redirects. Similarly, you can check to make sure that the expected header keys are present -- whatever level of detail seems robust.

automation/DataAggregator/parquet_schema.py
```diff
@@ -100,6 +100,7 @@
     pa.field('frame_id', pa.int64()),
     pa.field('response_status', pa.int64()),
     pa.field('response_status_text', pa.string(), nullable=False),
+    pa.field('redirects', pa.string()),
```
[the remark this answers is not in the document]
eriktews: Yes, sorry, I will fix that.

automation/DataAggregator/schema.sql
eriktews: That's a good question. In general, when a redirect is triggered, we get pretty much everything as for an response, except for the content or contentHash. So I could modify the patch in a way that a redirect would trigger two writes, one in the redirect table and one in the response table. However, both tables are currently already very similar, except for the missing old and new URL and request id fields that are only present in the redirect table and we would then save the data twice.

The other option would be to merge the response table with the redirect table, but then some fields would be empty depending on whether it's an redirect or not.

In my opinion, having the headers in the redirect table is currently the cleanest solution, but other more invasive changes could be possible as well.

eriktews:
> A test would be great! You don't have to test all of the details of the headers. You can check to make sure that the expected number of headers are present with and without redirects. Similarly, you can check to make sure that the expected header keys are present -- whatever level of detail seems robust.

Since only **Date** (varies), **Server** (includes the current Python version number, will vary as well) and **Location** (stable) are used as headers in the current test case, I decided to extend the testcase with c319e8142b7239bc0aaaef4da2be489e5d5cf119 and test for the presence of the location header in the database.

automation/DataAggregator/schema.sql
englehardt: I agree. This seems like the best approach to me as well. Thanks for thinking through it.

Issue #28, opened 2015-11-03
Record access to all window.navigator properties


englehardt:
The properties [we currently collect](https://github.com/citp/OpenWPM/blob/master/automation/Extension/firefox/data/content.js#L286) are largely those ported over from FourthParty. We should add as much from [this list](http://help.dottoro.com/ljglevrj.php) as possible. It's even fine to add shims for properties Firefox doesn't support.

Examples of things we're missing include
- `window.navigator.oscpu`
- `window.navigator.doNotTrack`
- `window.navigator.clientinformation`

englehardt:
Increased the set a bit here: 7b22664b7fa20a87987795d04853a308ebd823d8

Be sure to include any new APIs in the test files for this. See: [extension_tests.py](https://github.com/citp/OpenWPM/blob/master/test/test_extension.py#L45-L70)

PR #642, opened 2020-05-09
Arbitrary WebAPI JS instrumentation

### LINKED ISSUE #641: Instrument arbitrary JS modules
Extend `js_instrumentation_modules` to accept any WebAPI object e.g. CanvasRenderingContext2D or XMLHttpRequest.

birdsarah:
Fixes #641 

To do:
* [x] Getting the pasted below error when instrumenting XMLHttpObject. These all appear to be `Can't redefine non-configurable property "XXX"`. Need to handle this property type. There may be other cases that come up. **wontfix** - these errors don't prevent instrumentation occurring.
* [x] Implement API as below. Can also pass in a logSettings object.
* [x] Ensure do not repeat instrument - creates issues because the new prototype ends up with separate getters and setters.
* [x] Can pass, for example, "excludedProperties" or "nonExistingPropertiesToInstrument" a logSettings object if desired.
* [x] Tests
* [x] Add support to instrument arrays (plugins, mimeTypes currently missing from fingerprinting.json too) - I've added back a version of the instrumentation that could be sufficient.
* [x] Performance test - tests that: 
  * [x] (a) time how long it takes instrumentation to load 
  * [x] (b) hits an API over and over and see if we miss any. Decided there's no point in doing this as instrumentation def loads first.
  * [x] (c) will *not* cover how much instrumentation of n apis slows down a page.
* [x] Regression - add a test for not propogating down propertiesToInstrument and fix regression I introduced.
* [x] Add JSON parsing to crawler.py
* [x] Update README . 
  * [x] From @englehardt review - "Now that folks can instrumentation arbitrary JS objects, we should (a) mention that, (b) describe what a shortcut is, and which ones exist already, and (c) provide a short example of two." "I like that we've moved the documentation for instrumentObject out of JS (which is good, since it's user configurable). I'm not sure how someone who wants to write their own settings file would be able to figure out which options are available. Is the only way to read the schema file? That would be pretty clunky. (Perhaps this will be covered by a README update)."
  * [x] Add developer docs about rendering schemas to markdown.

```js
      js_instrument:true,
      js_instrument_modules: [
        // Shortcut
        "fingerprinting",
        // APIs
        "Storage",
        {"XMLHttpRequest": ["send"]},
        // Specific instances on window
        {"window.document": ["cookie", "referrer"]},
        {"window": ["name", "localStorage", "sessionStorage"]}
      ],
      http_instrument:true,
      callstack_instrument:true,
```

For follow-on issues:
* Update [openwpm-crawler](https://github.com/mozilla/openwpm-crawler) with new input
* ~~Add an "all" option~~ (nope - crashes everything)
* ~~Do you want to be able to specify just a property? e.g. `XMLHttpReqest.send` vs `XMLHttpRequest` (this could be a follow-on PR)~~

Questions:
* For "fingerprinting" - do we want to keep the current set of instrumented APIs which is some whole modules, and sometimes just a limited set of properties? For example, we only take pixelDepth and colorDepth from window.screen instead of the full 15 or so options. The difference might be starker on other modules, I haven't done a thorough review. 
  - Keep the same for now.
* I'm leaving all logSettings options default. As best as I can tell we've never used them. Thoughts? (Given that we haven't used it, I'm tempted to propose removing it to simplify the code).
  - I was wrong. They're all used (or desired in the webcompat work - https://github.com/mozilla/OpenWPM/compare/master...motin:webcompat-instrumentation). This PR hopefully makes logSettings and it's default clearer.
* Why not have it so settings are passed around as JSON and JSON is handled webext side.
  - Because by making a magic js string we are actually referring to the window object as opposed to having a string in JSON. It's' the difference between `'{object: window.CanvasRenderingContext2D.prototype, instrumentedName: "CanvasRenderingContext2D",...}'` and `{object: "window.CanvasRenderingContext2D.prototype", instrumentedName: "CanvasRenderingContext2D",...}`. In the latter case where we're passing around JSON, on the JS side we then have to find a way to turn the string `"window.CanvasRenderingContext2D.prototype"` into the object `window.CanvasRenderingContext2D.prototype`.


birdsarah:
@englehardt - how would you feel if "fingerprinting" instrumented a few more things e.g. whole of navigator and CanvasRenderingContext2D?

englehardt:
> @englehardt - how would you feel if "fingerprinting" instrumented a few more things e.g. whole of navigator and CanvasRenderingContext2D?

If it seems like it can be used for fingerprinting I think it's fair game. My only warning is that some APIs are incredibly noisy.

birdsarah:
>  My only warning is that some APIs are incredibly noisy.

Yeah, I think I need to leave a more flexible solution open.

birdsarah:
@englehardt - how do you feel about this as a starting point API for this PR. It will meet all my needs and the needs of the current fingerprinting instrumentation. We can follow-up with more implementation as needs arise (see "follow-on prs" above).

```js
      js_instrument:true,
      js_instrument_modules: [
        // Shortcut
        "fingerprinting",
        // APIs
        "Storage",
        {"XMLHttpRequest": ["send"]},
        // Specific instances on window
        {"window.document": ["cookie", "referrer"]},
        {"window": ["name", "localStorage", "sessionStorage"]}
      ],
      http_instrument:true,
      callstack_instrument:true,
```

englehardt:
> @englehardt - how do you feel about this as a starting point API for this PR. It will meet all my needs and the needs of the current fingerprinting instrumentation. We can follow-up with more implementation as needs arise (see "follow-on prs" above).
> 
> ```js
>       js_instrument:true,
>       js_instrument_modules: [
>         // Shortcut
>         "fingerprinting",
>         // APIs
>         "Storage",
>         {"XMLHttpRequest": ["send"]},
>         // Specific instances on window
>         {"window.document": ["cookie", "referrer"]},
>         {"window": ["name", "localStorage", "sessionStorage"]}
>       ],
>       http_instrument:true,
>       callstack_instrument:true,
> ```

This looks like a great starting point! Thanks!

birdsarah:
I've tried this on a crawl. I'm reviewing the results this week.

automation/default_browser_params.json
```diff
@@ -2,7 +2,7 @@
     "extension_enabled": true,
     "cookie_instrument": false,
     "js_instrument": false,
-    "js_instrument_modules": "fingerprinting",
+    "js_instrument_modules": ["fingerprinting"],
```
motin, line 5: Don't forget to retain support for crawler env vars (strings only) here

birdsarah: Indeed. Thanks for the reminder.I bumped into that when I was running crawls. [This](https://github.com/mozilla/OpenWPM/commit/361982e4d58c765621d87593444ddcc1199ab58b#diff-ee054b4b797f6c2a5743d2654764a8ed) is the change I made to `crawler.py`. I should add this to this PR and will need to add documentation about structuring the `crawl.yaml` probably to openwpm-crawler. I was actually please with how that ends up looking:

```yaml
        - name: NAVIGATION_INSTRUMENT
          value: '1'
        - name: JS_INSTRUMENT
          value: '1'
        - name: CALLSTACK_INSTRUMENT
          value: '1'
        - name: JS_INSTRUMENT_MODULES
          value: >
            [
                "AbortController",
                "window.navigator",
                "fingerprinting"
            ]
        - name: DWELL_TIME
          value: '15'
        - name: TIMEOUT
          value: '30'
```

englehardt:
Will fix #28

birdsarah:
@englehardt from riot:

> either way, the right way to instrument these going forward should be to just instrument access to `window.navigator.plugins` and `window.navigator.mimeTypes`, since the array will always be empty

Great. That's what this PR does (https://github.com/mozilla/OpenWPM/pull/642/files#diff-d95cdf4276d8381913a7a5466750094dR58). Although there's no need to break them out as separate.

For #28 - we can just update that JSON to be all window.navigator?

automation/default_browser_params.json
motin: Having an ability to use JSON to specify instrumentation configuration / options is a great addition in general. I'd love to see one more change here though - don't re-use JS_INSTRUMENT_MODULES. Add a separate config option called something like JS_INSTRUMENT_ATTRIBUTES or similar. Separation of concerns, allowing for JS_INSTRUMENT_MODULES and JS_INSTRUMENT_ATTRIBUTES to live separate lives in the codebase.

birdsarah: I wouldn't say this PR re-uses JS_INSTRUMENT_MODULES but completely redefines it. That may or may not be good.

Can you go into a bit more detail so I can understand what you're proposing: 
* JS_INSTRUMENT_MODULES and JS_INSTRUMENT_ATTRIBUTES are different concepts and where they would come back together for a uniifed

motin: If it completely redefines it and doesn't keep anything related to configuring which js "modules" are to be used, a rename may be more appropriate. Maybe just rename to _ATTRIBUTES, or _SPEC?

motin: This would also solve my bad choice of _MODULES in the first place, since it is confusing that OpenWPM docs talks about the different instruments as modules, and then there is a _MODULES config option for a module :)

birdsarah: Naming things is not my strong suit. @englehardt thoughts?

englehardt: I suggest renaming the browser param name as well.

automation/JSInstrumentation.py
```diff
@@ -0,0 +1,204 @@
...
+    {"XMLHttpRequest": {"badSetting": 1}},
+    {"XMLHttpRequest": {"excludedProperties": ["send"]}},
+    {"Prop1": ["hi"], "Prop2": ["hi2"]},
+    {"XMLHttpRequest": ["send"]},
+    "Storage",
+    // Specific instances on window
+    {"window.document": ["cookie", "referrer"]},
+    {"window": ["name", "localStorage", "sessionStorage"]}
+    """
+    assert isinstance(js_instrument_modules, list)
```
englehardt, line 193: You shouldn't use assert here. This method is potentially parsing user input and we'll always want to throw an error if there's an issue with the parameters (i.e.,`ValueError`). Assert can be [optimized out](https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement); it should only be used for debugging.

englehardt: It looks like this applies to all of the assert statements here; if a user is constructing their own settings they may cause any of these mistakes accidentally.

automation/JSInstrumentation.py
```diff
@@ -0,0 +1,204 @@
...
+        if propertiesToInstrument is not None:
+            propertiesToInstrument = set(propertiesToInstrument)
+            excludedProperties = set(
+                request['logSettings']['excludedProperties'])
+            assert len(propertiesToInstrument.intersection(
+                excludedProperties)) == 0
+    return True
+
+
+def merge_object_requests(python_list):
```
englehardt, line 69: Why support this at all? It seems safer to throw an error if there are duplicate configurations for the same object. Especially given that this isn't 100%.

englehardt: I guess maybe this might happen from recursive instrumentation?

automation/JSInstrumentation.py
```diff
@@ -0,0 +1,204 @@
...
+
+
+def convert_browser_params_to_js_string(js_instrument_modules):
+    """
+    We accept a list. From the list we need to parse each item.
+    Examples of the requests we accept.
+    // Shortcut
+    "fingerprinting",
+    // APIs
+    {"XMLHttpRequest": {"badSetting": 1}},
```
englehardt, line 184: What if I just want to instrument `XMLHttpRequest`?

automation/JSInstrumentation.py
```diff
@@ -0,0 +1,204 @@
...
+        'logSettings': logSettings
+    }
+
+
+def convert_browser_params_to_js_string(js_instrument_modules):
+    """
+    We accept a list. From the list we need to parse each item.
+    Examples of the requests we accept.
+    // Shortcut
+    "fingerprinting",
```
englehardt, line 182: I can imagine we may want to create a "storage" shortcut in the future that instruments localStorage, sessionStorage, indexedDB, Cache, and the like. Since this accepts both shortcut literals and API literals, this might get a bit confusing. E.g., `storage` and `Storage` would do very different things.

What do you think about making it explicit for the user by adding a prefix to the shortcuts? Perhaps something like `[shortcut]-fingerprinting`, `[shortcut]-storage`, `[shortcut]-canvas`, ...? That way we can check for prefix directly and not have to guess or worry about conflicts.

automation/JSInstrumentation.py
```diff
@@ -0,0 +1,204 @@
...
+    else:
+        raise RuntimeError('Invalid input')
+    return {
+        'object': obj,
+        'instrumentedName': instrumentedName,
+        'logSettings': logSettings
+    }
+
+
+def convert_browser_params_to_js_string(js_instrument_modules):
```
englehardt, line 177: This is basically the "user facing" function in the module, right? I suggest renaming it to something like `parse_settings(js_instrument_settings)`, and prefixing the rest of the helper functions with `_` to make that clear.

automation/default_browser_params.json
motin: Using _SETTINGS is confusing as well, since _MODULES (if kept) is also a js instrument configuration / setting. Ideally, something specific to what the setting does should also be its name, eg if it contains the spec for js instrument to use, _SPEC could be good.

englehardt: Sarah has removed the notion of a "module" and replaced it with a "shortcut" (i.e., a predefined set of APIs to instrument).

birdsarah: _spec, _settings, _config, ....

motin: > _spec, _settings, _config, ....

You are right. _spec is no better. What about _apis? JS_INSTRUMENT_APIS

englehardt: Let's use `_SETTINGS`. We don't have a distinction between modules and non-modules so it covers everything nicely.

birdsarah: Resolved in 8f593c8

automation/Extension/firefox/feature.js/index.js
```diff
@@ -20,7 +20,25 @@ async function main() {
       navigation_instrument:true,
       cookie_instrument:true,
       js_instrument:true,
-      js_instrument_modules:"fingerprinting",
+      js_instrument_modules: `
```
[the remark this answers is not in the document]
birdsarah: It's not possible in a compact way. The JS side no longer holds the concept of a shortcut and so now requires the full specification. Which in this case is:

```
[{"object": window['ScriptProcessorNode'].prototype, "instrumentedName": "ScriptProcessorNode", "logSettings": {"propertiesToInstrument": [], "nonExistingPropertiesToInstrument": [], "excludedProperties": [], "logCallStack": false, "logFunctionsAsStrings": false, "logFunctionGets": false, "preventSets": false, "recursive": false, "depth": 5}}, {"object": window['GainNode'].prototype, "instrumentedName": "GainNode", "logSettings": {"propertiesToInstrument": [], "nonExistingPropertiesToInstrument": [], "excludedProperties": [], "logCallStack": false, "logFunctionsAsStrings": false, "logFunctionGets": false, "preventSets": false, "recursive": false, "depth": 5}}, {"object": window['AnalyserNode'].prototype, "instrumentedName": "AnalyserNode", "logSettings": {"propertiesToInstrument": [], "nonExistingPropertiesToInstrument": [], "excludedProperties": [], "logCallStack": false, "logFunctionsAsStrings": false, "logFunctionGets": false, "preventSets": false, "recursive": false, "depth": 5}}, {"object": window['OscillatorNode'].prototype, "instrumentedName": "OscillatorNode", "logSettings": {"propertiesToInstrument": [], "nonExistingPropertiesToInstrument": [], "excludedProperties": [], "logCallStack": false, "logFunctionsAsStrings": false, "logFunctionGets": false, "preventSets": false, "recursive": false, "depth": 5}}, {"object": window['OfflineAudioContext'].prototype, "instrumentedName": "OfflineAudioContext", "logSettings": {"propertiesToInstrument": [], "nonExistingPropertiesToInstrument": [], "excludedProperties": [], "logCallStack": false, "logFunctionsAsStrings": false, "logFunctionGets": false, "preventSets": false, "recursive": false, "depth": 5}}, {"object": window['AudioContext'].prototype, "instrumentedName": "AudioContext", "logSettings": {"propertiesToInstrument": [], "nonExistingPropertiesToInstrument": [], "excludedProperties": [], "logCallStack": false, "logFunctionsAsStrings": false, "logFunctionGets": false, "preventSets": false, "recursive": false, "depth": 5}}, {"object": window['RTCPeerConnection'].prototype, "instrumentedName": "RTCPeerConnection", "logSettings": {"propertiesToInstrument": [], "nonExistingPropertiesToInstrument": [], "excludedProperties": [], "logCallStack": false, "logFunctionsAsStrings": false, "logFunctionGets": false, "preventSets": false, "recursive": false, "depth": 5}}, {"object": window['HTMLCanvasElement'].prototype, "instrumentedName": "HTMLCanvasElement", "logSettings": {"propertiesToInstrument": [], "nonExistingPropertiesToInstrument": [], "excludedProperties": [], "logCallStack": false, "logFunctionsAsStrings": false, "logFunctionGets": false, "preventSets": false, "recursive": false, "depth": 5}}, {"object": window['Storage'].prototype, "instrumentedName": "Storage", "logSettings": {"propertiesToInstrument": [], "nonExistingPropertiesToInstrument": [], "excludedProperties": [], "logCallStack": false, "logFunctionsAsStrings": false, "logFunctionGets": false, "preventSets": false, "recursive": false, "depth": 5}}, {"object": window.navigator, "instrumentedName": "window.navigator", "logSettings": {"propertiesToInstrument": [], "nonExistingPropertiesToInstrument": [], "excludedProperties": [], "logCallStack": false, "logFunctionsAsStrings": false, "logFunctionGets": false, "preventSets": false, "recursive": false, "depth": 5}}, {"object": window['CanvasRenderingContext2D'].prototype, "instrumentedName": "CanvasRenderingContext2D", "logSettings": {"propertiesToInstrument": [], "nonExistingPropertiesToInstrument": [], "excludedProperties": ["setTransform", "moveTo", "transform", "quadraticCurveTo", "globalAlpha", "drawImage", "canvas", "beginPath", "lineTo", "translate", "clearRect", "closePath"], "logCallStack": false, "logFunctionsAsStrings": false, "logFunctionGets": false, "preventSets": false, "recursive": false, "depth": 5}}, {"object": window, "instrumentedName": "window", "logSettings": {"propertiesToInstrument": ["localStorage", "name", "sessionStorage"], "nonExistingPropertiesToInstrument": [], "excludedProperties": [], "logCallStack": false, "logFunctionsAsStrings": false, "logFunctionGets": false, "preventSets": false, "recursive": false, "depth": 5}}, {"object": window.document, "instrumentedName": "window.document", "logSettings": {"propertiesToInstrument": ["cookie", "referrer"], "nonExistingPropertiesToInstrument": [], "excludedProperties": [], "logCallStack": true, "logFunctionsAsStrings": false, "logFunctionGets": false, "preventSets": false, "recursive": false, "depth": 5}}, {"object": window.screen, "instrumentedName": "window.screen", "logSettings": {"propertiesToInstrument": ["colorDepth", "pixelDepth"], "nonExistingPropertiesToInstrument": [], "excludedProperties": [], "logCallStack": false, "logFunctionsAsStrings": false, "logFunctionGets": false, "preventSets": false, "recursive": false, "depth": 5}}]

```

automation/Extension/webext-instrumentation/src/content/javascript-instrument-content-scope.ts
```diff
@@ -37,10 +46,10 @@ function emitMsg(type, msg) {
   });
 }
 
-const event_id = Math.random();
+const $event_id = Math.random().toString();
```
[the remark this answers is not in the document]
birdsarah: Resolved in 941c3cc

automation/Extension/webext-instrumentation/src/content/javascript-instrument-page-scope.ts
```diff
@@ -2,10 +2,7 @@
 // Also, no webpack/es6 imports may be used in this file since the script
 // is exported as a page script as a string
 
-export const pageScript = function({
-  jsInstruments,
-  instrumentFingerprintingApis,
-}) {
+export const pageScript = function($getInstrumentJS, $instrumentionRequests) {
```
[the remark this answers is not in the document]
birdsarah: Resolved in 98b4e37

automation/Extension/webext-instrumentation/src/content/javascript-instrument-content-scope.ts
```diff
@@ -1,29 +1,38 @@
...
-    jsInstruments +
-    "\n" +
-    instrumentFingerprintingApis +
-    "\n" +
-    "(" +
-    pageScript +
-    "({jsInstruments, instrumentFingerprintingApis}));"
-  );
+function getPageScriptAsString(
+  jsInstrumentationRequestsString: string,
```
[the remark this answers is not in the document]
birdsarah: Resolved in 101c41a

automation/JSInstrumentation.py
birdsarah: Resolved in 6756eaf

automation/JSInstrumentation.py
birdsarah: Resolved in 6756eaf

automation/JSInstrumentation.py
birdsarah: To make it easier to mix collections and API requests. I still think this is a good idea. We can remove it if it causes drama.

automation/JSInstrumentation.py
birdsarah: We decided on `collection-*`

automation/JSInstrumentation.py
birdsarah: Resolved in 520c782

automation/JSInstrumentation.py
birdsarah: Resolved in 8ac03dc

birdsarah:
@englehardt This PR is finished. I've incorporated all your suggestions / questions / concerns and the things we discussed in person the other day. The only thing I had to not do was "JSON everywhere". Half way through changing, I remembered why I don't do that. Here's why: 

(copied from front matter on this PR)
* Why not have it so settings are passed around as JSON and JSON is handled webext side.
  - Because by making a magic js string we are actually referring to the window object as opposed to having a string in JSON. It's' the difference between `'{object: window.CanvasRenderingContext2D.prototype, instrumentedName: "CanvasRenderingContext2D",...}'` and `{object: "window.CanvasRenderingContext2D.prototype", instrumentedName: "CanvasRenderingContext2D",...}`. In the latter case where we're passing around JSON, on the JS side we then have to find a way to turn the string `"window.CanvasRenderingContext2D.prototype"` into the object `window.CanvasRenderingContext2D.prototype`.

Whilst it may be possible to do it via JSON being passed to the webext, I don't think it should hold up this PR. Users still write a python list or JSON, it's just the transport between python and webext that's finecky and I don't think this PR needs to guarantee that that interface will be stable.

README.md
```diff
@@ -157,8 +204,36 @@ available [below](#output-format).
...
+    * The fingerprinting collection is specified by the json file
+      [fingerprinting.json](automation/js_instrumentation_collections/fingeprinting.json).
+      This file is also a nice reference example for specifying your own APIs using the other
+      shortcuts.
+    * Shortcuts:
+        * Specifying just a string will instrument
+          the whole API with the [default log settings](docs/schemas/js_instrument_settings-settings-objects-properties-log-settings.md)
+        * For just strings you can specify a [Web API](https://developer.mozilla.org/en-US/docs/Web/API) 
+          such as `XMLHttpRequest`. Or you can specify instances on window e.g. `window.document`.
+        * Alternatively, you can specify a dictionary with just one key. As with strings, the key
```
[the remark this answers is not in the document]
birdsarah: What I'm trying to say is that one key for the dictionary that may be an item in the list. Maybe https://github.com/mozilla/OpenWPM/blob/js-instrumentation/automation/js_instrumentation_collections/fingerprinting.json makes it clearer.

this is expected:
```json
"Storage",
"window.navigator",
{"window": {
        "propertiesToInstrument": ["name", "localStorage", "sessionStorage"]
    }
},
{"window.document": {
        "propertiesToInstrument": ["cookie", "referrer"],
        "logCallStack": true
    }
}
```

this is not allowed:
```
"Storage",
"window.navigator",
{"window": {
        "propertiesToInstrument": ["name", "localStorage", "sessionStorage"]
    },
 "window.document": {
        "propertiesToInstrument": ["cookie", "referrer"],
        "logCallStack": true
    }
}
```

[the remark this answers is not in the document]
birdsarah: I'm not quite sure how to make it clearer in text.

automation/Extension/webext-instrumentation/src/types/browser-web-navigation-event-details.ts
```diff
@@ -45,11 +45,13 @@ export interface WebNavigationOnCommittedEventDetails
   transitionQualifiers?: TransitionQualifier[];
 }
 
+/*
```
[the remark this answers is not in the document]
birdsarah: Whoops. Was planning to delete them. They're not used anywhere so was just cleaning up. That can be separate PR.

automation/js_instrumentation.py
```diff
@@ -5,7 +5,7 @@
 
 curdir = os.path.dirname(os.path.realpath(__file__))
 schema_path = os.path.join(
-    curdir, 'js_instrumentation', 'js_instrument_modules.schema'
+    curdir, os.pardir, 'schemas', 'js_instrument_settings.schema.json'
```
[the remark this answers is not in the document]
birdsarah: It's the naming pattern that the jsonschema2md uses by default. i'm not wed to it, I don't think it's a big change to jsonschema2md call to fix it.

package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "openwpm",
+  "version": "0.10.0",
```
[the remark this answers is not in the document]
birdsarah: I think we should remove it as long as npm doesn't complain to hard. But also remove it from the two other package.json files (which have not been updated in releases). I put it in because the other two package.json files had a version and 0.10 seemed like as good as any to pick.

test/openwpm_jstest.py
```diff
@@ -0,0 +1,37 @@
+import re
+
+from ..automation.utilities import db_utils
+from .openwpmtest import OpenWPMTest
+
+
+class OpenWPMJSTest(OpenWPMTest):
```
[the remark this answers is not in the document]
birdsarah: The get_config is a specific subclass unique to the js tests - I was just drying out that repeated block that was all over the place.

I think it's helpful to have it separate because my not secret longer term goal is to move this over to a pytest paradigm which doesn't have classes and uses fixtures and it helps to know what goes with what if we're going to make that change.

test/test_js_instrument.py
```diff
@@ -1,129 +1,286 @@
...
+class TestJSInstrumentExistingWindowPropertyByPython(TestJSInstrumentExistingWindowProperty):  # noqa
+    # This test is a repeat of above test, but as an integration test
+    # to test correct setting of instrumentaiton via browser params.
 
     def get_config(self, data_dir=""):
-        manager_params, browser_params = self.get_test_config(data_dir)
-        browser_params[0]['js_instrument'] = True
-        manager_params['testing'] = True
+        manager_params, browser_params = super().get_config(data_dir)
+        browser_params[0]['js_insrument_modules'] = [
```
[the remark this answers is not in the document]
birdsarah: It is running (https://travis-ci.org/github/mozilla/OpenWPM/jobs/703203803#L1153)

It is picking up the mis-named setting: https://travis-ci.org/github/mozilla/OpenWPM/jobs/703203803#L1205

And it's passing. :s

And it's passing because it's using the same test page which has the JS setting in! Great catch. This needs fixing. Thanks.

automation/js_instrumentation.py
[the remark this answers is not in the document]
englehardt: If it's the default that's fine. No need to change, I just wanted to make sure this wasn't a bug.

README.md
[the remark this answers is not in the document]
englehardt: Maybe "Alternatively you can specify a single-key dictionary that maps an API name one of the following:"?

README.md
```diff
@@ -80,20 +80,36 @@ After running the install script, activate your conda environment by running:
...
-### Troubleshooting
-
-`make` / `gcc` may need to be installed in order to build the web extension. On Ubuntu, 
-this is achieved with `apt-get install make` on OSX the necessary packages are part of xcode
-`xcode-select --install`
+In the rare instance that you need to create schema docs 
+(after updating or adding files to `schemas` folder), run `npm install`
+from OpenWPM top level. Then run `npm run render_schema_docs`. This will update the 
+`docs/schemas` folder. You may want to clean out the `docs/schemas` folder before doing this
+incase files have been renamed.
```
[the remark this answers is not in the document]
birdsarah: Resolved in 12141c5

README.md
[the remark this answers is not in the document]
birdsarah: Resolved in 76b3712

package.json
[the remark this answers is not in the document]
birdsarah: Resolved in 7dcebfc

test/test_js_instrument.py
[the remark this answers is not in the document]
birdsarah: Fixed in 9559ec9.

This actually was a bigger fix because you can't instrument properties that don't exist. So the other tests work because they build the object, then call the window.instrumentObject function. However, pyside the instrumentation loads before page load (which we want, so we don't miss anything) but it means that we can't instrument arbitrary objects, so I instrument real objects. Which had the nice side effect of me ending up getting a better understanding of how the instrumentation works, so I added a little extra to the README.

birdsarah:
@englehardt this is done. The bad test needed a bigger tweak (https://github.com/mozilla/OpenWPM/pull/642#discussion_r451301042) but nothing wild.

test/test_pages/js_instrument/instrument_pyside.html
```diff
@@ -0,0 +1,29 @@
...
+    <script type="text/javascript">
+        // Doubles as a test for the webdriver property
+        console.log(window.navigator.webdriver);
+        // Test set a cookie
+        window.document.cookie = 'a=COOKIE';
+        // Test get a cookie
+        console.log(window.document.cookie);
+        // Test both fetch() and window.fetch()
+        fetch(
+          "https://example1.com"
```
englehardt, line 22: These are real websites. We should avoid hitting them in our test suite.

I recommend using `example.com` and `example.org` (since these domains are official test domains by iana), but should still avoid making requests to the live servers. We can do that by setting the pref `network.dns.localDomains = example.com,example.org` which forces these to resolve to localhost.

birdsarah: Fixed in e2ef603.

PR #645, opened 2020-05-12
added codecov support

### LINKED ISSUE #640: Addition of codecov for analyisng test coverage
We can get a reporting dashboard of the tests being run in the repo during Travis, and then can have additional quality checks on the basis of coverage
@vringar

shubhank-saxena:
This fixes #640 
We still need to register the repo on codecov so that the test coverage can be picked up!
@vringar

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,11 +25,13 @@ before_install:
 install:
   - echo "y" | ./install.sh
   - pip3 install --upgrade -r requirements.txt -r requirements-dev.txt
+  - pip install codecov
 before_script:
   - pre-commit run --all
 script:
   - cd test
-  - python3 -m pytest -s -v --durations=10 $TESTS
+  - python3 -m pytest --cov --cov-report=xml -s -v --durations=10 $TESTS
+  - codecov -f coverage.xml
 
 jobs:
   include:
@@ -42,6 +44,7 @@ jobs:
       before_install:
       before_script:
       install:
+      - pip install codecov
       script:
         - docker build -f Dockerfile -t openwpm .
         - ./.deploy-to-dockerhub.sh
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -109,3 +117,4 @@ jobs:
       after_success:
         - # npx nyc report --reporter=lcov | npx codecov
```
vringar, line 118: Why is this line still here, did we use to have codecov enabled before?
And why should we do it differently now than we did before?

vringar:
Thanks for submitting this PR.
There are a couple of points that are still unclear to me only some of which can be addressed by you:
- Who do I need to talk to, to get this enabled for us? (That's probably a task for me)
- Shouldn't these tools be in `requirements-dev.in` and `package.json` respectively?
- Do we want to create policies around this tool (e.g. no lowering the code coverage) or is it just informative

shubhank-saxena:
- We can add them in the respective requirement files. But codecov only runs just on the Travis. So this way it won't clutter the packages. If you want, I can shift them.
- Currently, this is informative only. Can be toggled to a policy if and when required (like a threshold that every PR should at least maintain 40%+ coverage and shouldn't dip it

.travis.yml
shubhank-saxena: That's something I missed :open_mouth:

shubhank-saxena: I followed up using npm. Can be done this way too

birdsarah:
> We can add them in the respective requirement files. But codecov only runs just on the Travis. So this way it won't clutter the packages. If you want, I can shift them.

I think we should include in dev requirements. It's small and it makes sure everywhere's working with the same set of dependencies which makes it easier to track down problems.

The rebase now #648 is in will probably be a bit gross. Apologies. You may just want to start a new branch. The good news is the PR will be smaller as .travis.yml shrunk considerably.

shubhank-saxena:
@birdsarah sure! Will start a new PR!  #648 is literally a great thing! :rocket:

PR #647, opened 2020-05-13
Fix bulk of linting errors

birdsarah:
Was getting bothered by all the linting errors, so fixing the bulk.


birdsarah:
This is intended as basically a no-op PR.
* It leaves two warnings that I can't readily fix. 
* It removes the "no-console-log" lint rule.

Largely, removing non-arrow functions, and re-ordering lists of things.

birdsarah:
I'll pick this up later.

Issue #615, opened 2020-04-18
Manage openwpm dependencies with conda


birdsarah:
We can use conda instead of running scripts that install programs into people's environments.

We need to consider how to handle a clean integration with pip requirements. Do we keep a requirements file? How do we pin things and keep things up to date?

Original WIP PR: #429

vringar:
Whenever this gets addressed we should close #66

shubhank-saxena:
@birdsarah any specific reason to go for conda packages? It would again cause conda as a dependency. If we want more generic, we can consider poetry as one (Plus its methods comply to PEP 518).

birdsarah:
I'm afraid I've never heard of poetry before you mentioned it. Can you use poetry to install: firefox, node, redis, ....?

If yes, as best as I can tell from poetry it's still relatively young. 

Having worked for anaconda, I am not unbiased and it would be good to get additional input.

But conda / anaconda back AWS's EMR, Google's Dataproc, it's used at hundreds of very large companies along with over 19million users worldwide who've downloaded literally billion packages - it's really been put through the ringer and is a tried and tested cross-platform package-management solution.

shubhank-saxena:
> I'm afraid I've never heard of poetry before you mentioned it. Can you use poetry to install: firefox, node, redis, ....?
> 
> If yes, then poetry and conda would be on par in terms of the desired functionality. But I would still think that conda is a safer option because, as best as I can tell, poetry is still relatively young. Conda is very well established:
> 
> * conda / anaconda back AWS's EMR, Google's Dataproc,
> * it's used at hundreds of very large companies along with
> * over 19million users worldwide who've downloaded literally billions of packages
> * it's really been used a lot, and is a tried and tested cross-platform package-management solution.
> 
> All that said, if poetry is feature compatible, given that I used to work for anaconda, I am not unbiased, and it would be good to get additional input.
> 
> [edited for clarity was typing on mobile]

Wow, these are really good features! Thanks for letting me know! Will study more about condo packaging. It sounds way more better than poetry

Issue #381, opened 2019-07-01
Automatically manage Firefox version upgrades


englehardt:
We're currently pulling the latest Firefox binary from `https://index.taskcluster.net/v1/task/gecko.v2.mozilla-release.latest.firefox.linux64-add-on-devel/artifacts/public/build/target.tar.bz2`. This means that we'll get the latest version of the binary as soon as a new version release happens. This may completely break OpenWPM (as the latest upgrade to FF68 has).

We should use an alternate taskcluster URL and pull a static build from that. See https://tools.taskcluster.net/index/gecko.v2 for our options. Upgrades to the Firefox binary version should happen manually with a new PR updating the taskcluster link. That will let us run the full test suite and discover any breaking changes. We should also check the version of Firefox that the user has installed, and suggest that they upgrade when it's outdated. See also #66.

englehardt:
We could pull from `https://index.taskcluster.net/v1/task/gecko.v2.mozilla-release.pushdate.2019.07.01.20190701165736.firefox.linux64-add-on-devel/artifacts/public/build/target.tar.bz2` for example, which matches the build currently returned from `https://index.taskcluster.net/v1/task/gecko.v2.mozilla-release.latest.firefox.linux64-add-on-devel/artifacts/public/build/target.tar.bz2`.

We could pull in `https://index.taskcluster.net/v1/task/gecko.v2.mozilla-release.latest.firefox.linux64-add-on-devel/artifacts/public/build/target.json`, read the latest Firefox binary version from that and warn when an update is available.

motin:
Related: https://github.com/freaktechnik/get-firefox/issues/61

motin:
Coincidentally, I ran into this python script for downloading old versions of Firefox today: https://github.com/jrbenny35/experiment-test-automation-suite/blob/master/utilities/download_old_firefox.py

englehardt:
In #385 we hard code a specific unbranded Firefox build. My plan is to write a simple cron job to nudge me when builds with a new FF version are released, but this is just a stopgap solution to better crawl management. Ideally, we have some cloud job running that checks the latest version and stages a PR if a new build is available. That way we will run the new build against the test suite and upgrade if necessary.

motin:
[...]

Issue #66, opened 2016-04-08
Platform should track current Firefox version and warn user if different


englehardt:
Right now we only suggest a Firefox version in the install script. Since different versions will have incompatibilities, especially in the Javascript instrumentation, we should track the required version within the infrastructure and warn the user if the version they are launching with doesn't match the required version.

Examples include the renaming of `mozRTCPeerConnection` to `RTCPeerConnection` in the jump from 41 to 45.

englehardt:
We should do this an have the install script grab the newest version of the current ESR. The accepted way to get the latest version seems to be to use the following: `curl "https://download.mozilla.org/?product=firefox-esr-latest" | sed -e 's/.*releases\///g' | cut -d "/" -f1` (see the [README](https://ftp.mozilla.org/pub/firefox/releases/latest/README.txt)). However, that will start returning Firefox 59 during the 52-59 ESR transition period.

Instead, it looks like we can just grep the ftp listing: `curl "https://ftp.mozilla.org/pub/firefox/releases/" | grep "/pub/firefox/releases/52." | tail -n 1 | sed -e 's/.*releases\///g' | cut -d "/" -f1`.

englehardt:
454c25a30d0bf9ca74ba203df2514e607b12fc1f changes the install script to automatically grab the latest point release of the current ESR. We should consider warning the user if their current release is behind the latest point release and prompt them to re-run the install script. That requires two things:

1. The install script should check the currently installed version in the `firefox-bin` directory and update if outdated.
2. We want to check the latest version in such a way that we don't end up hammering the FTP server if users happen to run the platform in a way we don't anticipate.

englehardt:
```json
{
  "version": "Mozilla Firefox 80.0",
  "hash": "abc"
}
```

Have the openwpm startup function check this version file and compare the version number to the number via geckodriver. This is possible to check as we're deploying the browser. After we initialize the driver (https://github.com/mozilla/OpenWPM/blob/93ddf80c185de5f4298a2870eb04b05a0bc522e8/automation/DeployBrowsers/deploy_firefox.py#L175-L180) we can query its capabilities. If the versions don't match we should throw a critical error (https://github.com/mozilla/OpenWPM/blob/93ddf80c185de5f4298a2870eb04b05a0bc522e8/automation/Errors.py#L21).

PR #648, opened 2020-05-13
Conda install dependencies

### LINKED ISSUE #628: Running tests requires sudo
I was running tests locally and surprised to find them asking for my sudo password!

```
Starting local_http_server
Starting HTTP Server in a separate thread
...serving at port 8000
Starting edge router (http port 4566)...
Please enter your sudo password (required to configure local network):
[sudo] password for birdsarah:
```

I'm pretty sure this is related to localstack services as per: https://github.com/localstack/localstack/issues/2371

### LINKED ISSUE #638: Replace automation.utilities.domain_utils with library
domain_utils was extracted from openwpm-utils and made into a standalone package with tests and features that cover a lot more edge cases.

propose removing automation.utilities.domain_utils and using https://pypi.org/project/domain-utils/

birdsarah:
Fixes: #615, #638, #628

This PR will result in the following changes:

- [x] Installation of OpenWPM will have a completely new procedure.
  * [x] Documented in README
  * [x] Implemented in docker
  * [x] Implemented in .travisci
  * [x] Same installation procedure for OSX and Linux (OpenWPM remains windows incompatible)
- [x] Geckodriver will be conda installed and so no geckodriver finding, it's just assumed to be on path
- [x] Re-pinning of dependencies will be done via a script, documented in README, not handled through pip-tools
- [x] There will only be one environment that contains both development and production needs. See "one environment" note below.
- [x] While cleaning up dependencies also add in domain_utils and resolve #638
- [x] Should fix #628, as it was fixed upstream and this PR will pick up those changes
- [x] Will standardize on one version of python. Because python installation is managed via conda, and all installation flows from conda, we do not need to support and test against multiple versions of python.
- [x] Will standardize on one version of node, for the same reason as above. This will reduce complexity of .travis.yaml
- [x] The majority of scripts / dev utils have been moved to a scripts directory, to keep things a bit tidier.

Firefox installation will not be managed by conda installation for now. There will need to be a follow-on PR and possible work done in conda packaging. The firefox available via `conda install firefox -c conda-forge` is release, not unbranded. (So #66 and #381 will not be resolved here.) 

Note on one environment: With this PR we'll move to having only one environment instead of a requirements.txt and a requirements-dev.txt. Conda doesn't have the same tooling for managing dependencies as pip-tools. I'm happy with the `repin.sh` it's pretty simple. But I can't see a lightweight way to make a pinned `environment-dev.yaml` that has the same pinning `environment.yaml`. That same pinning is critical to ensure that the tests we run give the same result as someone who would just be using `environment.yaml`. That's not to say it isn't possible, we can write a bash script to do it, but I can't see that the extra maintenance and complexity is worth it, the only impact, as far as I can see, is that the conda environment grows a little. To measure this, I ran `conda pack` with and without the dev dependencies. The packed environments is 295M without dev dependencies and 322M with dev dependencies, a difference of <10%. I have, however, kept the input yaml files seperate so it's still clear what dependencies we have for dev and what for prod.

Package management desires / constraints:
- Installation should be encapsulated in an environment and not change user's machine outside that environment.
-  Pinned packages for deployments to facilitate replicability
  - Pinning / replicability is not so important for dev environments / toolsets.
- Run tests on OSX as well (future)

After merging this PR:
- [ ] Update wiki to reflect new installation changes, in particular OSX workflow
- [ ] Review need for `run-on-osx-via-docker.sh` script


birdsarah:
@englehardt - when you get a chance can you review the todo list and desires/constraints to make sure I've captured all the needs?

Dockerfile
```diff
@@ -1,60 +1,35 @@
...
-COPY automation/Extension/webext-instrumentation/. ../webext-instrumentation
-RUN npm install
-RUN npm run build
-RUN cp dist/openwpm-1.0.zip openwpm.xpi
-
-# Stage 2, build the main OpenWPM image
 FROM krallin/ubuntu-tini:bionic
 
-WORKDIR /opt/OpenWPM
+SHELL ["/bin/bash", "-c"]
```
vringar, line 3: We need to verify that our signal propagation still works as expected when we do this, since bash doesn't propagate a SIGTERM properly but instead just SIGKILLS all child processes.
This might be a non issue, just raising awareness

vringar:
As for documenting stuff in the wiki @englehardt  and I discussed that we want to move our documentation into a `docs` folder as this would allow us to get PRs against it.
I'm also hoping by moving this into the repository we'll be able to get our documentation onto readthedocs

Dockerfile
birdsarah: how would we verify this?

vringar: I'd build the docker image and start docker compose
When the crawler gets scaled down the containers are send a SIGTERM
If the crawler receives a SIGTERM ([which the crawler logs](https://github.com/mozilla/OpenWPM/blob/2178e4c0ea641b95e861e336cf320a95b81402b3/crawler.py#L137) ) then we'd know the signal gets propagated as expected.

vringar:
@nhnt11 do you have the time to try out `run-on-osx-via-docker.sh` because AFAIK you're the only one using MacOS that's involved with OpenWPM

Dockerfile
vringar: I just realized we can't test just on your branch as issue-443 isn't merged yet.
I'll do that locally and report back

Dockerfile
```diff
@@ -1,60 +1,35 @@
...
+
+ENV HOME /opt
+COPY scripts/install-miniconda.sh .
+RUN ./install-miniconda.sh
+ENV PATH $HOME/miniconda/bin:$PATH
+
+# Install OpenWPM
+WORKDIR /opt/OpenWPM
+COPY . .
+RUN ./install.sh
```
vringar, line 26: By having this run after COPY we need to rebuild the enviroment every time any file in the entire repo changes, which takes a significant amount of time and might break the workflow of people that develop in a docker image.
```suggestion
COPY enviroment.yaml scripts/conda_setup.sh .
RUN ./scripts/conda_setup.sh
COPY automation/extension .
RUN ./scripts/build-extension.sh
COPY . .
```
Where `scripts/conda-setup.sh` would be the first part of `install.sh`

Dockerfile
vringar: I just checked and it gets propagated as expected

birdsarah:
> @nhnt11 do you have the time to try out `run-on-osx-via-docker.sh` because AFAIK you're the only one using MacOS who's involved with OpenWPM

I was going to leave this for a follow-up discussion, but with this PR I would debate the usefullness/appropriateness of “run-on-osx-via-docker.sh”. 

I have run the install script on OSX on my husband’s machine and been able to tweak a few things, check it all runs. OSX and Linux should now be on equal footing when it comes to developer experience. My caveats in the docs are related to me not having fully explored, for example, xvfb and OSX. With that in mind, I don’t see the point of special casing OSX and docker. If we want general “run with docker” stuff - I think that’s well handled in the “openwpm-crawler” repo.  

(That’s the gist of my thoughts, pushback welcome)

Dockerfile
birdsarah: I agree to a certain extent. In particular, every time I changed the Dockerfile I needed to rerun a lot. I considered adding more to .dockerignore, but wanted to check in on that.

My reasoning behind what I have now is as little code duplication as I could possibly squeeze so we’re only updating one thing when we do maintenance.

Here’s what it would look like - it would mean some duplication from install.sh - but it’s not that bad. It would look like this:

```dockerfile
COPY scripts .  # We could copy scripts one by one but it’s probably not worth it, they won’t change often.
RUN ./scripts/install-miniconda.sh
ENV PATH $HOME/miniconda/bin:$PATH
COPY environment.yaml .
RUN conda env create --force -q -f environment.yaml
ENV PATH $HOME/miniconda/envs/openwpm/bin:$PATH
COPY automation/extension .
RUN ./scripts/build-extension.sh
RUN ./scripts/install-firefox.sh
... move firefox
COPY automation crawler.py demo.py . # Or `COPY . .`
```

I wanted to clarify what you mean by `scripts/conda-setup.sh` - there’s only one line - I don’t think it needs to be broken out into a script.

> Might break the workflow of people that develop in a docker image

It’s not that much code and not that much duplication, so I’m okay about making the change, but who’s actually developing in a docker image? (Or who would be now). When I was developing the docker file, and didn’t want to do all the waiting, I updated the docker file myself to include the lines from `install.sh`. We could also include instructions about this in the README and keep the main docker file clean of duplicated code.

vringar: > I wanted to clarify what you mean by scripts/conda-setup.sh - there’s only one line - I don’t think it needs to be broken out into a script.

Yeah, it's just a oneliner I didn't think that one through properly, sorry.

> but who’s actually developing in a docker image?

I was for a longer period of time that I'd like to admit because the pinned dependencies weren't available for Python3.8, but since that's no longer an issue I'm mostly thinking about people that develop on Windows but still want to run OpenWPM.

[Pycharm for example allows configuring a docker container as a remote interpreter](https://www.jetbrains.com/help/pycharm/using-docker-as-a-remote-interpreter.html) and [VSCode allows something similiar](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers)
That's also why I appreciate now having the dev dependencies be part of the docker image.

vringar:
First of all sorry for being unable to read I read 
> Review need for run-on-osx-via-docker.sh script
as 
> Review need_ed_ for run-on-osx-via-docker.sh script

> I was going to leave this for a follow-up discussion, but with this PR I would debate the usefullness/appropriateness of “run-on-osx-via-docker.sh”.

Having never looked into the script I expected it to do a lot more than it currently does.
We should/could just document the `xhost`command the same way we currently do for linux users.

And the actual docker exec is just the same thing as for the linux version, so yeah I agree with you, we don't need it anymore

Dockerfile
birdsarah: Please hold for windows update....

birdsarah: I've successfully run through the dev pipeline using WSL1 and 2, suggest we recommend that and don't recommend / support developing with docker.  I will open an issue to document developing on windows.

birdsarah:
Should be good to go @vringar. Thanks for testing it out. I have tested windows (via WSL), linux (ubuntu), the docker image, and osx. 

@englehardt i think the main thing you’ll mostly be fine with this. Our biggest discussion point was pinning and I worked hard to try and find a good solution for you, there are detailed notes in the PR doc. There are compromises, but I really think it’s good enough like this. It’s certainly working and we can iterate if needed.

englehardt:
> @englehardt i think you’ll mostly be fine with this. Our biggest discussion point was pinning and I worked hard to try and find a good solution for you, there are detailed notes in the PR doc. There are compromises, but I really think it’s good enough like this. It’s certainly working and we can iterate if needed.

I read through the readme changes and Comment 0 of this PR. I agree you found the best solution, and I really appreciate that we're able to continue to pin all dependencies + easily update them. Simplifying things by removing the dev/non-dev distinction seems like a worthwhile tradeoff. Thanks for putting in the work to figure everything out and for thoroughly documenting it.

r+ from me for the meta-review. I'll leave it to @vringar to do a more thorough pass.

PR #666, opened 2020-05-26
Consider non-prettier-formatted files a warning rather than an error in lint test

motin:
This makes hacking on the webext instrumentation more forgivable and clearer

### DIFF
--- a/automation/Extension/webext-instrumentation/package.json
+++ b/automation/Extension/webext-instrumentation/package.json
@@ -20,7 +20,7 @@
     "fix:prettier": "prettier \"src/**/*.ts\" --write",
     "fix:tslint": "tslint -t verbose --fix --project .",
     "test": "run-s build test:*",
-    "test:lint": "tslint -t verbose --project . && prettier \"src/**/*.ts\" --list-different",
+    "test:lint": "tslint -t verbose --project . && echo \",Files that needs to be prettier-formatted:\" && prettier \"src/**/*.ts\" --list-different || true",
     "test:unit": "nyc ava",
     "test:unit:verbose": "nyc ava --verbose --serial",
     "watch": "run-s clean build:main && run-p \"build:main -- -w\" \"test:unit -- --watch\"",


motin:
Number of the Beast!

birdsarah:
What about using the --write argument to auto rewrite?

motin:
Tests that are failing are not related to this PR. Intermittent test fails? Re-run tests?

motin:
> What about using the --write argument to auto rewrite?

That is basically the same, since there is no commit done afterwards by the lint commands.

Configuring a git hook to format (with --write) on commit would be a more robust approach, but those hooks needs to be installed on the root of the repository and so could potentially conflict with python-only workflows.

birdsarah:
Since the pre-commit hooks were added for python I'm now used to running them locally (I still do it manually). I think it would be great to hook in there.

Is your concern with "python only" that people won't have node? (which isn't a concern now we've switched to conda).

birdsarah:
> Tests that are failing are not related to this PR. Intermittent test fails? Re-run tests?

I'm not sure. @englehardt how do more of us get the ability to re-start travis runs.

birdsarah:
> That is basically the same, since there is no commit done afterwards by the lint commands.

Indeed!

motin:
> Since the pre-commit hooks were added for python I'm now used to running them locally (I still do it manually). I think it would be great to hook in there.
> 
> Is your concern with "python only" that people won't have node? (which isn't a concern now we've switched to conda).

Yes, but also the general idea of a python developer working on some python code, and have to have the npm dependencies of webext-instrumentation installed (prettier etc) even to commit any python code. This can be circumvented by letting the git hook check if any changes in the commit affects the webext-instrumentation code and only running in then. In either case, this particular PR is just a stop-gap measure to prevent irritating debugging of why tests/linting fail - for those that are not too familiar with the webext-instrumentation build process. I look forward to a PR for a non-obtrusive git hook :)

birdsarah:
> Yes, but also the general idea of a python developer working on some python code, and have to have the npm dependencies of webext-instrumentation installed (prettier etc) even to commit any python code.

I see your point, but could go either way.

>  In either case, this particular PR is just a stop-gap measure to prevent irritating debugging of why tests/linting fail - for those that are not too familiar with the webext-instrumentation build process.

I understand the pain as I experienced it myself working on the new JS PR recently.  I think I may have actually mis-understood the implications of this PR. Allowing the build to continue without enforcing the linting is +1 from me. But is there another step that means that a test will fail to prevent merging PRs that aren't properly formatted? My understanding of

englehardt:
> But is there another step that means that a test will fail to prevent merging PRs that aren't properly formatted? My understanding of `test:lint` is that it was the only place this got tested which is why I'm not keen on relaxing. But I'd be happy to hear I've misunderstood.

I've been bitten by these errors as well, but I think Sarah's point stands. If we don't fail when on Travis we are essentially dropping the formatting requirement since it can be silently ignored.

vringar:
We have dropped this requirements in dfa970f1, so we can close this PR.

PR #675, opened 2020-05-31
UI instrument

### LINKED ISSUE #669: ui-instrument: Instrument specific user interface properties / interactions
I propose a ui-instrument for instrumenting specific user interface properties and interactions, such as:
 - Clicks
 - Scrolls
 - Keyboard event metadata
 - Frame dimensions / Viewport dimensions
 - Document/Page visibility (indicates if the window/tab is active)
 - HTML5 Media states (is a sound/video playing etc)
 - etc

Track basic user interaction was desired in a previous study, comparing data collected during genuine user traffic vs crawls (https://github.com/mozilla/jestr-pioneer-shield-study/issues/5). One use-case is to detect violations of cookie policies where user consent / interaction is required before cookies are set.

Frame dimensions (useful to detect small/hidden iframes) has its own issue (https://github.com/mozilla/OpenWPM/issues/349), and is related to this issue as a possible home for its implementation.

Document/page visibility is useful to differentiate between activity from background vs foreground windows/tabs. 

HTML5 media states are useful to understand if there is video/audio (could be ads) playing / auto-playing during capture.

motin:
A first version of the proposed UI instrument. Implements:
 - Clicks
 - Document/Page visibility (indicates if the window/tab is active)
 - HTML5 Media states (is a sound/video playing etc)

Leaves the following for future PRs:
 - Scrolls
 - Keyboard event metadata
 - Frame dimensions / Viewport dimensions
 - etc

Fixes #669


englehardt:
I'm not comfortable merging such a large change without tests.

This instrumentation isn't very useful for web crawls, but I can see how you may want it for Pioneer. Is that what you've written this for?

I know we have the long-term goal to split the webextension instrumentation back out to a separate repo, but I'm also hesitant to merge things into it right now that (1) aren't exposed to the web crawler, and (2) aren't useful for web crawls. Achieving that long-term goal will require refactoring a lot of the tests, and I don't see us doing that anytime soon. My hesitation is rooted in the fact that OpenWPM has very little dedicated engineering support, and the support that we have is there for our web crawls. Adding this means that we have more code to support that we'll never end up using.

englehardt:
Closing this for the reasons given in https://github.com/mozilla/OpenWPM/pull/675#issuecomment-645683737

Issue #652, opened 2020-05-13
Remove dependency on localstack / refactor s3aggregator to parquet aggregator


birdsarah:
It's come up in a few issues so gathering the core issue here.

Related: #614, #628 

Localstack has caused us a number of headaches:
* dependency management (needing to pin older versions of libraries)
* upstream changes (#628)
* and generally adding complexity

It's purpose is to mock out s3, but that complexity is driven from the fact that the s3aggregator does two things for structured data:
* aggregates data into parquet
* pushes it to s3

In addition to that it saves out individual content files to s3.
We can separate these two functions allowing for openwpm to generate parquet and then to push it to a variety of locations: s3, gcs, or just local.

birdsarah:
@vringar @englehardt I recall from our conversations that the above description of the issue is not accurate and there's an additional piece of related to saving content. Can you update the above description and we can continue iterating the conversation?

englehardt:
This seems like a duplicate of #618? We should close that one or the other.

And yes, see https://github.com/mozilla/OpenWPM/pull/390 for context.

Right now we have:

1. S3Aggregator:
    1. structured data: parquet on S3
    2. unstructured data: individual gzipped files on S3 (very inefficient)
2. LocalAggregator:
    1. structured data: sqlite
    2. unstructured data: leveldb

it seems like we want an option to independently choose how to save the structured data (e.g., parquet / sql) from the unstructured data (e.g., leveldb when local and something else when remote).

birdsarah:
> it seems like we want an option to independently choose how to save the structured data (e.g., parquet / sql) from the unstructured data (e.g., leveldb when local and something else when remote).

+100

birdsarah:
Move from leveldb to rocksdb and use https://github.com/rockset/rocksdb-cloud to push to aws, google cloud etc?

birdsarah:
> Move from leveldb to rocksdb and use https://github.com/rockset/rocksdb-cloud to push to aws, google cloud etc?

I don't love this suggestion because I think the path to windows is easier with leveldb.

vringar:
One big thing I'd want to be mindful of is the discussion around caching and completion callbacks.
If we offload the caching into an external system we loose the knowledge whether something has been actually saved out to the remote storage or just cached locally.
At least for s3fs I couldn't find a way to manually flush, maybe this happens upon closing the FS?
[...]

PR #682, opened 2020-06-12
v0.10.0 / Firefox 77.0.1 release

englehardt:
I followed the steps in the new [release checklist](https://github.com/mozilla/OpenWPM/wiki/Release-Checklist) for this.

### DIFF
--- a/environment.yaml
+++ b/environment.yaml
@@ -10,7 +10,7 @@ dependencies:
 - geckodriver=0.26.0
 - ipython=7.15.0
 - leveldb=1.22
-- localstack=0.11.1
+- localstack=0.11.1.1
 - multiprocess=0.70.10
 - nodejs=14.4.0
 - pandas=1.0.5
@@ -25,16 +25,16 @@ dependencies:
 - redis-py=3.5.3
 - s3fs=0.4.0
 - selenium=3.141.0
-- sentry-sdk=0.15.0
+- sentry-sdk=0.15.1
 - tabulate=0.8.7
 - tblib=1.6.0
 - wget=1.20.1
 - pip:
   - amazon-kclpy==2.0.1
-  - crontab==0.22.8
+  - crontab==0.22.6
   - domain-utils==0.7.1
   - flask-cors==3.0.8
-  - moto-ext==1.3.15.21
+  - moto-ext==1.3.15.15
   - plyvel==1.2.0
   - subprocess32==3.5.4
 name: openwpm
--- a/environment.yaml
+++ b/environment.yaml
@@ -2,27 +2,27 @@ channels:
 - conda-forge
 - main
 dependencies:
-- autopep8=1.5.2
-- beautifulsoup4=4.9.0
+- autopep8=1.5.3
+- beautifulsoup4=4.9.1
 - dill=0.3.1.1
 - flake8-isort=3.0.0
-- flake8=3.8.1
+- flake8=3.8.3
 - geckodriver=0.26.0
-- ipython=7.14.0
+- ipython=7.15.0
 - leveldb=1.22
-- localstack=0.11.1.1
+- localstack=0.11.2
 - multiprocess=0.70.9
-- nodejs=13.13.0
-- pandas=1.0.3
+- nodejs=14.4.0
+- pandas=1.0.4
 - pillow=7.1.2
-- pip=20.1
-- pre-commit=2.4.0
+- pip=20.1.1
+- pre-commit=2.5.1
 - psutil=5.7.0
-- pyarrow=0.17.0
-- pytest=5.4.2
-- python=3.8.2
-- pyvirtualdisplay=0.2.5
-- redis-py=3.5.2
+- pyarrow=0.17.1
+- pytest=5.4.3
+- python=3.8.3
+- pyvirtualdisplay=1.3.2
+- redis-py=3.5.3
 - s3fs=0.4.0
 - selenium=3.141.0
 - sentry-sdk=0.14.4
@@ -31,10 +31,8 @@ dependencies:
 - wget=1.20.1
 - pip:
   - amazon-kclpy==2.0.1
-  - crontab==0.22.6
-  - domain-utils==0.7.1
+  - crontab==0.22.8
   - flask-cors==3.0.8
-  - moto-ext==1.3.15.15
-  - plyvel==1.2.0
+  - moto-ext==1.3.15.19
   - subprocess32==3.5.4
 name: openwpm
--- a/environment.yaml
+++ b/environment.yaml
@@ -4,16 +4,16 @@ channels:
 dependencies:
 - autopep8=1.5.3
 - beautifulsoup4=4.9.1
-- dill=0.3.1.1
+- dill=0.3.2
 - flake8-isort=3.0.0
 - flake8=3.8.3
 - geckodriver=0.26.0
 - ipython=7.15.0
 - leveldb=1.22
 - localstack=0.11.2
-- multiprocess=0.70.9
+- multiprocess=0.70.10
 - nodejs=14.4.0
-- pandas=1.0.4
+- pandas=1.0.5
 - pillow=7.1.2
 - pip=20.1.1
 - pre-commit=2.5.1
@@ -25,14 +25,16 @@ dependencies:
 - redis-py=3.5.3
 - s3fs=0.4.0
 - selenium=3.141.0
-- sentry-sdk=0.14.4
+- sentry-sdk=0.15.0
 - tabulate=0.8.7
 - tblib=1.6.0
 - wget=1.20.1
 - pip:
   - amazon-kclpy==2.0.1
   - crontab==0.22.8
+  - domain-utils==0.7.1
   - flask-cors==3.0.8
-  - moto-ext==1.3.15.19
+  - moto-ext==1.3.15.21
+  - plyvel==1.2.0
   - subprocess32==3.5.4
 name: openwpm
--- a/environment.yaml
+++ b/environment.yaml
@@ -21,7 +21,7 @@ dependencies:
 - pyarrow=0.17.1
 - pytest=5.4.3
 - python=3.8.3
-- pyvirtualdisplay=1.3.2
+- pyvirtualdisplay=0.2.5
 - redis-py=3.5.3
 - s3fs=0.4.0
 - selenium=3.141.0
--- a/environment.yaml
+++ b/environment.yaml
@@ -10,7 +10,7 @@ dependencies:
 - geckodriver=0.26.0
 - ipython=7.15.0
 - leveldb=1.22
-- localstack=0.11.2
+- localstack=0.11.1.1
 - multiprocess=0.70.10
 - nodejs=14.4.0
 - pandas=1.0.5
@@ -31,6 +31,7 @@ dependencies:
 - wget=1.20.1
 - pip:
   - amazon-kclpy==2.0.1
+  - crontab==0.22.8
   - domain-utils==0.7.1
   - flask-cors==3.0.8
   - moto-ext==1.3.15.15
--- a/environment.yaml
+++ b/environment.yaml
@@ -10,7 +10,7 @@ dependencies:
 - geckodriver=0.26.0
 - ipython=7.15.0
 - leveldb=1.22
-- localstack=0.11.1.1
+- localstack=0.11.2
 - multiprocess=0.70.10
 - nodejs=14.4.0
 - pandas=1.0.5
@@ -31,7 +31,6 @@ dependencies:
 - wget=1.20.1
 - pip:
   - amazon-kclpy==2.0.1
-  - crontab==0.22.6
   - domain-utils==0.7.1
   - flask-cors==3.0.8
   - moto-ext==1.3.15.15
--- a/environment.yaml
+++ b/environment.yaml
@@ -10,7 +10,7 @@ dependencies:
 - geckodriver=0.26.0
 - ipython=7.15.0
 - leveldb=1.22
-- localstack=0.11.2
+- localstack=0.11.1
 - multiprocess=0.70.10
 - nodejs=14.4.0
 - pandas=1.0.5
--- a/scripts/environment-unpinned-dev.yaml
+++ b/scripts/environment-unpinned-dev.yaml
@@ -5,14 +5,16 @@ dependencies:
     - flake8
     - flake8-isort
     - ipython
-    - localstack==0.11.1
+    - localstack==0.11.1.1
     - pip
     - pre-commit
     - pytest
     - pip:
         # Select depenedencies from localstack[full] that we need
-        - amazon-kclpy
-        - crontab
-        - flask-cors
-        - moto-ext
-        - subprocess32
+        # These should be pinned to compatible versions for the current
+        # localstack version
+        - amazon-kclpy==2.0.1
+        - crontab==0.22.6
+        - flask-cors==3.0.8
+        - moto-ext==1.3.15.15
+        - subprocess32==3.5.4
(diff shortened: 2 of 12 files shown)


environment.yaml
```diff
@@ -31,10 +31,8 @@ dependencies:
 - wget=1.20.1
 - pip:
   - amazon-kclpy==2.0.1
-  - crontab==0.22.6
-  - domain-utils==0.7.1
+  - crontab==0.22.8
   - flask-cors==3.0.8
-  - moto-ext==1.3.15.15
-  - plyvel==1.2.0
```
birdsarah, line 38: Assuming you didn't manually edit this yourself, something has gone wrong with the repin script. We've lost domain-utils and plyvel.

englehardt: Hmm maybe it's because my conda env wasn't active?

englehardt: I've updated conda and retried it with the openwpm env active. No change. Seems like the script may have a bug. Note that both of these are python dependencies, so it could be related to that.

birdsarah: activating openwpm env won't change anything. the script activates it for itself (https://github.com/mozilla/OpenWPM/blob/master/scripts/repin.sh#L10)

first debug step. run 
```
# Create openwpm env with unpinned yaml file
conda env create --force -q -f environment-unpinned.yaml

# Activate
conda activate openwpm
```

Then `conda list` and see if domain-utils and plyvel are in the list.

Then, next steps from repin

```
# Adding dev dependencies to environment
conda env update -f environment-unpinned-dev.yaml

# Export the environment including manually specify channels
conda env export --no-builds --override-channels -c conda-forge -c main -f ../environment.yaml
```

Then have a look at `environment.yaml`. If plyvel and domain-utils are in there (which I expect they will be) then the problem is with `prune-environment.py`.

I'm expecting them to be successfully in because it's much more likely that the custom script I wrote to clean up the environment is broken rather than conda being broken.

Can you post the output from `../environment.yaml` before pruning.

englehardt:
The missing packages discussed in comment https://github.com/mozilla/OpenWPM/pull/682#discussion_r439211411 are missing from `conda env create --force -q -f environment-unpinned.yaml`.  Unfortunately it looks like conda doesn't disable `~/.local` when generating the list of packages to install via pip, see https://github.com/conda/conda/issues/7173 and https://github.com/conda/conda/issues/448. This means that some requirements may be "already satisfied" when creating the environment if the user already has them installed in `~/.local`:

```
steven@apollo:~/research/OpenWPM/scripts$ conda env create --force -f environment-unpinned.yaml
Collecting package metadata (repodata.json): done
Solving environment: done
Preparing transaction: done
Verifying transaction: done
Executing transaction: done
Ran pip subprocess with arguments:
['/home/steven/miniconda3/envs/openwpm/bin/python', '-m', 'pip', 'install', '-U', '-r', '/home/steven/research/OpenWPM/scripts/condaenv.o4ueqpf1.requirements.txt']
Pip subprocess output:
Requirement already up-to-date: plyvel in /home/steven/.local/lib/python3.8/site-packages (from -r /home/steven/research/OpenWPM/scripts/condaenv.o4ueqpf1.requirements.txt (line 1)) (1.2.0)
Requirement already up-to-date: domain-utils in /home/steven/.local/lib/python3.8/site-packages (from -r /home/steven/research/OpenWPM/scripts/condaenv.o4ueqpf1.requirements.txt (line 2)) (0.7.1)
```
Note that we shouldn't assume this is unusual. Pip has long been recommended users install things in `~/.local` (via the `--user` flag) when not using a virtualenv, and I believe they've recently made that the default (https://github.com/pypa/pip/issues/1668). Thankfully there's a simple workaround for the repin script (see https://github.com/conda/conda/issues/448#issuecomment-195848539).

I suspect users who have site packages installed that are incompatible with the conda environment's packages may still see issues when trying to run OpenWPM (I believe @vringar experienced this before?). I wonder if we should write a simple `conda-activate.sh` that sets `PYTHONNOUSERSITE=True` and runs `conda activate openwpm`? My understanding is that `virtualenv` will do something like this for proper isolation (https://github.com/pypa/virtualenv/issues/24). @birdsarah wdyt?

environment.yaml
englehardt: To avoid this getting lost in a comment thread I followed up directly in the PR: https://github.com/mozilla/OpenWPM/pull/682#issuecomment-645648939

birdsarah:
> Note that we shouldn't assume this is unusual.  Pip has long been recommended users install things in ~/.local (via the --user flag) when not using a virtualenv,

....<removed unecessary rant>....I do agree that openwpm's user base are not software engineers typically so we should be mindful to make things robust.

> I suspect users who have site packages installed that are incompatible with the conda environment's packages may still see issues when trying to run OpenWPM (I believe @vringar experienced this before?)

hmmmm....

>  I wonder if we should write a simple conda-activate.sh that sets PYTHONNOUSERSITE=True and runs conda activate openwpm?

I really like the PYTHONNOUSERSITE fix.  I do not believe it will be possible to make a conda-activate script, otherwise I would have already as the wierd line that's at the top of all the sh scripts is activating the conda environment for that script. But I may be wrong.

scripts/environment-unpinned-dev.yaml
```diff
@@ -5,14 +5,16 @@ dependencies:
...
-        - flask-cors
-        - moto-ext
-        - subprocess32
+        # These should be pinned to compatible versions for the current
+        # localstack version
+        - amazon-kclpy==2.0.1
+        - crontab==0.22.6
+        - flask-cors==3.0.8
+        - moto-ext==1.3.15.15
+        - subprocess32==3.5.4
```
birdsarah, line 20: i don't believe it's necessary to pin all of these dependencies. but this is a pure gut instinct reaction.

scripts/repin.sh
```diff
@@ -10,7 +10,7 @@ set -e
 eval "$(conda shell.bash hook)"
 
 # Create openwpm env with unpinned yaml file
```
birdsarah, line 12: We should add a note about the PYTHONNOUSERSITE=True addition (can just point to the discussion on this PR) so that future devs can know why it's here.

scripts/environment-unpinned-dev.yaml
birdsarah: My concern is that overpinning creates future tech debt when packages should be updated to more secure / uptodate versions but we don't know / remember why packages were pinned.

birdsarah: But your comment is clear, so I'm "meh" about it in this case.

birdsarah:
@englehardt, I restarted the tests from just before you pinned localstack.

The a_e that were previously failing, now passed. The _s tests that are failing appear to be failing due to xvfb which I doubt has anything to do with localstack.

birdsarah:
I just noticed that pyvirtualdisplay is showing a big version bump in this PR " pyvirtualdisplay=0.2.5" -> "pyvirtualdisplay=1.3.2" which I would at least want to investigate as the source of the xvfb errors.

birdsarah:
The problem is accessing display_port. Which we probably can still do we just need to figure out the new API which is probably something like `display.display.??`. But do we need display_port?
As best as I can tell it's only used to remove a /tmp file. Why do we need this cleanup?

```
        if self.display_port is not None:  # xvfb diplay lock
            lockfile = "/tmp/.X%s-lock" % self.display_port
            try:
                os.remove(lockfile)
            except OSError:
                self.logger.debug("BROWSER %i: Screen lockfile (%s) already "
                                  "removed" % (self.crawl_id, lockfile))
                pass
```

scripts/environment-unpinned-dev.yaml
birdsarah: I take above comments back. I no longer think its necessary to pin localstack at all.

englehardt:
> > Note that we shouldn't assume this is unusual.  Pip has long been recommended users install things in ~/.local (via the --user flag) when not using a virtualenv,
> 
> ........I do agree that openwpm's user base are not software engineers typically so we should be mindful to make things robust.

I don't see how this distinction is important. We shouldn't make assumptions about the user's system setup.

> > I suspect users who have site packages installed that are incompatible with the conda environment's packages may still see issues when trying to run OpenWPM (I believe @vringar experienced this before?)
> 
> hmmmm....

Unfortunately this seems to be the case :-/. I filed https://github.com/mozilla/OpenWPM/issues/689 to track that.

> > I wonder if we should write a simple conda-activate.sh that sets PYTHONNOUSERSITE=True and runs conda activate openwpm?
> 
> I really like the PYTHONNOUSERSITE fix. I do not believe it will be possible to make a conda-activate script, otherwise I would have already as the wierd line that's at the top of all the sh scripts is activating the conda environment for that script. But I may be wrong.

Yeah I see what you mean. I didn't play around with it, but included this as an option in https://github.com/mozilla/OpenWPM/issues/689.

birdsarah:
> I don't see how this distinction is important. We shouldn't make assumptions about the user's system setup.

I deleted an unhelpful set of comments from myself. I've never run into this issue before and I attribute that fact, perhaps incorrectly, to the fact that I learned from day 1 as a software engineer that everything should be done in a virtualenv (or equivalent) otherwise you can run into all kind of weirdness.

But without the comments I deleted the second thought about being robust lost it's meaning.

scripts/environment-unpinned-dev.yaml
englehardt: I narrowed it down to `moto-ext` (at least when I run the tests locally)

I was also wondering whether to pin all of them or just that one. I chose to pin all of them because it took me a while for me to recognize that this was the issue. localstack didn't give a helpful error, and I didn't initially realize that we added some of localstack's dependencies separate from localstack itself. Let's say all of these packages were updated in the future, the user would have to test them 1 by 1 to find the responsible package (or set of packages).

The risk I see if the pinned versions are incompatible with other (non-pinned) python packages. And of course security vulnerabilities, but hopefully GitHub will warn us there and we can update as needed.

I'm secretly hoping we don't have to keep this around too long and can just get rid of localstack (https://github.com/mozilla/OpenWPM/issues/652).

wdyt?

scripts/repin.sh
englehardt: Will do!

englehardt:
> The problem is accessing display_port. Which we probably can still do we just need to figure out the new API which is probably something like `display.display.??`. But do we need display_port?
> As best as I can tell it's only used to remove a /tmp file. Why do we need this cleanup?
> 
> ```
>         if self.display_port is not None:  # xvfb diplay lock
>             lockfile = "/tmp/.X%s-lock" % self.display_port
>             try:
>                 os.remove(lockfile)
>             except OSError:
>                 self.logger.debug("BROWSER %i: Screen lockfile (%s) already "
>                                   "removed" % (self.crawl_id, lockfile))
>                 pass
> ```

Yes, at least for the time being. This was initially added because long-running crawls would fill the tmp directory with too many lock files and would eventually lead XVFB to fail because it would be assigned a pid that already had a lock file.

This can probably be filed by actually shutting down xvfb properly in https://github.com/mozilla/OpenWPM/blob/43922bc18cc458eab3aaea618c1f4bb4b14ccf8f/automation/BrowserManager.py#L503-L512 and not just killing the process like we do right now.

scripts/environment-unpinned-dev.yaml
birdsarah: I don't see a reason to pin localstack, so then pinning moto-ext doesn't seem necessary.

birdsarah:
> > The problem is accessing display_port. Which we probably can still do we just need to figure out the new API which is probably something like `display.display.??`. But do we need display_port?
> > As best as I can tell it's only used to remove a /tmp file. Why do we need this cleanup?
> > ```
> >         if self.display_port is not None:  # xvfb diplay lock
> >             lockfile = "/tmp/.X%s-lock" % self.display_port
> >             try:
> >                 os.remove(lockfile)
> >             except OSError:
> >                 self.logger.debug("BROWSER %i: Screen lockfile (%s) already "
> >                                   "removed" % (self.crawl_id, lockfile))
> >                 pass
> > ```
> 
> Yes, at least for the time being. This was initially added because long-running crawls would fill the tmp directory with too many lock files and would eventually lead XVFB to fail because it would be assigned a pid that already had a lock file.
> 
> This can probably be fixed by actually shutting down xvfb properly in
> 
> https://github.com/mozilla/OpenWPM/blob/43922bc18cc458eab3aaea618c1f4bb4b14ccf8f/automation/BrowserManager.py#L503-L512
> and not just killing the process like we do right now.

This sounds good. If we properly shutdown xvfb then even if we have some rogue lock files due to a bad shutdown we certainly wouldn't expect the tmp directory to fill up because if every shutdown isn't shutting down properly then you have other problems. So this seems like a very satisfactory solution.

I'm also fine with pinning pyvirtualdisplay to `<1` for this PR and then having a seperate issue to tackle this.l

scripts/environment-unpinned-dev.yaml
englehardt: @birdsarah Why not? Pinning these dependencies fixed the test timeouts we were seeing here: https://travis-ci.org/github/mozilla/OpenWPM/jobs/699484251#L1132-L1158.

It's not clear to me _why_ pinning fixed the timeout. I had assumed it was a dependency mismatch, but the newer version of `moto-ext` is valid according to localstack's [requirement.txt](https://github.com/localstack/localstack/blob/6dcffb35fb685b4297599c9ee1064b6433ab64c9/requirements.txt#L36).

birdsarah: Pinning didn't fix the timeout. The timeout was a fluke. I re-ran the tests that failed before you pinned and they passed.

englehardt: Oh it looks like I missed your comment here: https://github.com/mozilla/OpenWPM/pull/682#issuecomment-646220583. It seems like this build passed https://travis-ci.org/github/mozilla/OpenWPM/builds/699484265. Interesting! I could reproduce the timeouts locally as well. Let me investigate.

birdsarah: Hmmmmm....They're timing out here too: https://travis-ci.org/github/mozilla/OpenWPM/jobs/699484251

birdsarah: There's two runs associated with each commit - I only successfully re-ran the tests on one. I don't know whether there's a meaningful difference between the two runs.

birdsarah: OK. My apologies. I now see the difference between the two runs. Please ignore all my comments regarding pinning. Glad you've pinned it down to moto-ext. Does this mean we can advance localstack and just pin moto-ext?

englehardt: > Glad you've pinned it down to moto-ext. Does this mean we can advance localstack and just pin moto-ext?

Yes looks like it. I just tried it locally and that worked. I'll push a new commit.

englehardt: okay I spoke too soon. I was bitten by my local packages. There are still failures on travis https://travis-ci.org/github/mozilla/OpenWPM/jobs/701020698. I've now re-pinned localstack.

scripts/repin.sh
```diff
@@ -10,7 +10,9 @@ set -e
...
-conda env create --force -q -f environment-unpinned.yaml
+# `PYTHONNOUSERSITE` set so python ignores local user site libraries when building the env
+# See: https://github.com/mozilla/OpenWPM/pull/682#issuecomment-645648939
+PYTHONNOUSERSITE=True conda env create --force -q -f environment-unpinned.yaml
```
birdsarah, line 15: I think we should also update `./install.sh` to match.

englehardt: Good catch! I added it there as well as during the update command within repin.sh

birdsarah:
sorry pressed the wrong button

PR #690, opened 2020-06-18
New url for conda

spirillen:
As conda have changed there url for `Miniconda3-latest-Linux-x86_64.sh` I've changed this to avoid at least one redirect and currently assured the future, in they disable the old url... which probably isn't soon, but why not do it right up :)


englehardt:
Thank you :)

spirillen:
> Fabulous thank you. I started working at anaconda when it was continuum, so my gists (which I copied from) are really old!

It is also only cos I just made another script using conda I noticed the change in the url :smile:

Issue #701, opened 2020-06-26
Crawl ID -> Browser ID, and matching data structures


birdsarah:
Following up from discussion on chat.

TIL that Crawl ID is actually meant to uniquely identify browser instances, rather than to give a "crawl" a global ID.

In the LocalAggregator, the concept exists of a Task ID, which identifies a set of browsers and visits as all belonging to one "task".

The "task" is what we would normally think of as a "crawl", in the sense of "I'm going to run an alexa 10k crawl".

The crawl id is actually identifying browser instances.

Looking at the LocalAggregator sql schema https://github.com/mozilla/OpenWPM/blob/1655dedc954733b9d67fa02e2547beb3654ad5cc/automation/DataAggregator/schema.sql#L6-L18 you can also see that storage of manager params against a task id and browser params against a crawl id.

This issue proposes:
* Rename `crawl_id` to `browser_id`
* Rename `task_id` to `crawl_id`
* Ensure that s3 aggregator also saves a task table

@englehardt asked on chat 

>> Makes sense. But I still think task_id -> crawl_id, crawl_id -> browser_id makes sense 

> I'm not sure why you'd want both (I have to head out now, but we can continue tomorrow / in an issue, sorry about that). 

Moving forward in this discussion, I am going to use the proposed nomenclature. That is, `browser_id` refers to what is currently `crawl_id` in the code, and `crawl_id` refers to `task_id`.

@englehardt also noted that browser_id *was* used in the S3 aggregator to infer that a browser instance has moved to a new site by observing that data with the same browser_id has moved from one visit_id to another. But, Stefan recently removed that with good reason.

So to take a stab at @englehardt's question: `browser_id` and `visit_id` both go to the extension, are used for the reporting back of data. But `visit_id` is highly unique. If we are not using `browser_id` for anything any more then it seems that we could get rid of it.
Alternatively, we could have a transition period where `crawl_id` is renamed to `browser_id` so the data structure stays the same, and the extension stays the same (save for some renaming), and then at a future date we can remove it as obsolete. This may make the transition easier and may make it easier to identify if there's use cases for both a `crawl_id` and a `browser_id`.

Regardless, I think the renaming will be clarifying and S3Aggregator data should match LocalAggregator. 

(Aside: The discovery of this mismatch is additional support for splitting S3Aggregator up, having one (sqlalchemy) base for all structured data that can then be saved to multiple backends - sqlite, postgres, bigquery, parquet etc.)

[...]

Issue #561, opened 2020-02-14
Rewrite DataAggregators


vringar:
If we split structured and unstructured storage we'd currently have the following 4 Aggregators:
- Structured
   - SQLLite
   - Parquet on S3
- Unstructured
   - LevelDB
   - S3

In a second step we then could move the SQLLite Aggregator to a [SQLAlchemy](https://www.sqlalchemy.org/) based one, so we could then write to a Cloud SQL DB. This would give us atomic commits with no batching and enable us to have an exactly once guarantee.
Also we should reconsider #230 as this would remove the need for a bunch of sockets and possibly a whole process

birdsarah:
I made a plan after doing a spike in #704. We discussed in person (@englehardt, @vringar, and @birdsarah) and agree this is a reasonable way forward.

Open questions:
- should you be able to send structured data and unstructured data to separate destinations? Can we live with not being able to do this for now.
- Why not have more listeners - one running process_record (listening to data coming up from the socket and parsing it out) and then one each for structured and unstructured data? (just a thought - this could pave the way for one process_record listener per browser if that was useful)

Notes:
- Don't see the need for s3_bucket and s3_directory
- Don't see the need for user's to be able to specify the database name - they can just specify the folder - like they do with s3
- I think this can work gracefully if we unify some concepts - like handling batches - which are in both.
- The tricky bit is that sometimes there's interplay between what you're storing and where you're storing it, so we need to make a graceful interface that allows for a matrix / mixin approach even if not all elements of matrix will be supported. In particular, blob output types (sqlite, leveldb) are not going to get cloud support. But there's no reason for parquet and gzip not to have local support. And postgres and rocksdb, for example, should work as local or cloud....I think I have a plan.

```
DataAggregator - should be scoped down to controlling the listener process and passing messages to it
- launch
- shutdown
- get_status
- get_most_recent_status
- get_new_completed_visits

It does not need to be responsible for making a visit_id.
It would perhaps be better named as a DataProcessBus

Listener - consider renaming to DataProcess 
- it has top level methods for setup and teardown, draining the queue, processing records, managing the
concept of a batch
- it sets up a "structured" and an "unstructured" class (unstructured only if needed - save_content=True) - examples below
- The (un)structured class that is used then decides how it wants to work with batches e.g. pretty diff between
S3Parquet and LocalSqlite. 
- The storage destination becomes the lowest level of implementation detail (see class structure below) instead of being
wrapped up at the top level.

- Structured
[...]

PR #704, opened 2020-06-26
Structured aggregator

birdsarah:
This PR is functionally a no-op, but it starts us down the path of splitting up the aggregators (see #701, #652, #561).

This PR dynamically builds a class DataAggregator pulling in either former LocalAggregator methods or former S3Aggregator methods.

We should be able to build on this relatively simply to start adding, for example, parquet saving locally. There maybe some combinations we never want to support - e.g. ldb in an s3 context (seems like we could support that though, even though it would be a little odd for distributed crawls).

I'm thinking that once we have unpacked the classes so that the type of structured/content data and how you save it are separate we can work on the pieces to unify the structured data processing with something like sqlalchemy and a single schema.

Open questions:
- should you be able to send structured data and unstructured data to separate destinations? Can we live with not being able to do this for now.
- Why not have more listeners - one running process_record (listening to data coming up from the socket and parsing it out) and then one each for structured and unstructured data? (just a thought - this could pave the way for one process_record listener per browser if that was useful)

Notes:
- Don't see the need for s3_bucket and s3_directory
- Don't see the need for user's to be able to specify the database name - they can just specify the folder - like they do with s3
- I think this can work gracefully if we unify some concepts - like handling batches - which are in both.
- The tricky bit is that sometimes there's interplay between what you're storing and where you're storing it, so we need to make a graceful interface that allows for a matrix / mixin approach even if not all elements of matrix will be supported. In particular, blob output types (sqlite, leveldb) are not going to get cloud support. But there's no reason for parquet and gzip not to have local support. And postgres and rocksdb, for example, should work as local or cloud....I think I have a plan.

```
DataAggregator - should be scoped down to controlling the listener process and passing messages to it
- launch
- shutdown
- get_status
- get_most_recent_status
- get_new_completed_visits

It does not need to be responsible for making a visit_id.
It would perhaps be better named as a DataProcessBus

Listener - consider renaming to DataProcess 
- it has top level methods for setup and teardown, draining the queue, processing records, managing the
concept of a batch
- it sets up a "structured" and an "unstructured" class (unstructured only if needed - save_content=True) - examples below
- The (un)structured class that is used then decides how it wants to work with batches e.g. pretty diff between
S3Parquet and LocalSqlite. 
- The storage destination becomes the lowest level of implementation detail (see class structure below) instead of being
wrapped up at the top level.

- Structured
-- Parquet
--- S3Parquet
--- LocalParquet
-- Sql
--- Postrgresql
--- LocalSqlite

- Unstructured
-- LocalLevelDB
-- Gzip
--- LocalGzip
--- S3Gzip
```

The goal of a first PR would be to:
* Only change TaskManager in two ways: 
  - Simplified launch of "DataAggregator" https://github.com/mozilla/OpenWPM/pull/704/commits/7c8538d171a36ff537bedb76b3ffe1340e210791#diff-df517bce823800d647ba110df4679453L261-L265
  - Remove visit_id responsibility out of DataAggregator
  - Rest of TaskManager interface remains unchanged but guts of DataAggregator gets pulled out and put back together with:
     - new DataProcess / Listener that has taken on new responsibilities as outline above
     - only implement: LocalGZip, LocalSQlite, S3Parquet, S3Gzip

Follow-on PR:
* Implement new manager params to allow for:
   - specific specification of structured / unstructured class desired
   - unify some params that can be

Follow-on PRs:
* Implement additional classes e.g. LocalParquet, LocalGzip
* Unify structured processing to have one schema backing it (so everyone gets task id for example)


automation/TaskManager.py
```diff
@@ -15,9 +15,8 @@
 from .BrowserManager import Browser
 from .Commands.utils.webdriver_utils import parse_neterror
 from .CommandSequence import CommandSequence
-from .DataAggregator import BaseAggregator, LocalAggregator, S3Aggregator
-from .DataAggregator.BaseAggregator import (ACTION_TYPE_FINALIZE,
-                                            RECORD_TYPE_SPECIAL)
+from .data_aggregator import BaseAggregator, build_data_aggregator_class
+from .data_aggregator.base import ACTION_TYPE_FINALIZE, RECORD_TYPE_SPECIAL
```
birdsarah, line 19: As an aside. @englehardt and I discussed incrementally moving the codebase to more conventional python naming (https://www.python.org/dev/peps/pep-0008/#package-and-module-names)

vringar: Does this mean, we'll also rename the base module from automation to openwpm?

vringar:
Thank you for spearheading this issue.
I really like this design but don't quite understand how it would work for the listener process.
Do you want to show how you would go about doing that here or would you want to file a separate PR for that?

automation/TaskManager.py
birdsarah: Not any time soon. But that seems logical to me.

birdsarah:
> Thank you for spearheading this issue.
> I really like this design but don't quite understand how it would work for the listener process.
> Do you want to show how you would go about doing that here or would you want to file a separate PR for that?

I'm not sure yet, but I don't see why it couldn't be coerced into working.

For what it's worth I'm not sure that in the end we would end up with a solution that looks like this - it's a little unconventional and I think it probably makes it harder to read the code base and have a clear sense of what's going on quickly. That said, it's a path to untangle the situtation we're in and get into a new situation.

birdsarah:
>  I really like this design but don't quite understand how it would work for the listener process.

If you're being polite, and think it wouldn't work, then it's fine to ask to add the listener process to this PR.

automation/data_aggregator/builder.py
```diff
@@ -2,9 +2,57 @@
 
 
 def build_data_aggregator_class(manager_params):
-    output_format = manager_params["output_format"]
+    """
+    Suggest renaming data_directory and s3_bucket
+    to a consistent thing like "output_folder"
```
vringar, line 7: How about `output_url` this way we handle any [pyFilesytem](https://www.pyfilesystem.org/) as long as it's installed.

birdsarah:
Closing this in favor of a new PR based on the plan outlined above and discussed with @englehardt and @vringar in person.

PR #715, opened 2020-07-23
Update hash function - use await (and generally clean up)

### LINKED ISSUE #711: How does the content hash saved to levelDB evaluated?
When running OpenWPM over a set of URLs, I observed that it computes different content hashes (md5) for the same script (https://connect.facebook.net/en_US/fbevents.js) used in different first party URLs (e.g., http://jacquescartierchamplain.ca/?lang=en and http://jacquescartierchamplain.ca/circulation-travaux/reseau-en-direct/pont-honore-mercier/). Although the content hash (MD5) of the script URL (https://connect.facebook.net/en_US/fbevents.js) is different for the first party URLs, the source content of the script is identical. So, does this mean that the content hash is computed from the rendered content of the script (i.e., not from the script source)?

birdsarah:
This code change updates the sha256 function to match the latest code on MDN.

It's my attempt to fix #711. 

I don't think we could say we definitively know it's worked to fix #711 without running it with a large crawl and repeating the analysis I did.

That said, we have tests in `test_http_instrumentation.py` that explicitly test for expected hashes. 

This code is simpler than the previous code. So if we think the tests are sufficient, then we could merge this on the grounds it's simpler and have an outstanding todo to rerun an analysis when a crawl's been run with this fix in.

### DIFF
--- a/automation/Extension/webext-instrumentation/package-lock.json
+++ b/automation/Extension/webext-instrumentation/package-lock.json
@@ -1408,25 +1408,6 @@
         "xdg-basedir": "^4.0.0"
       }
     },
-    "conventional-changelog": {
-      "version": "3.1.18",
-      "resolved": "https://registry.npmjs.org/conventional-changelog/-/conventional-changelog-3.1.18.tgz",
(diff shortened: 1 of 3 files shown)


vringar:
> This code is simpler than the previous code. So if we think the tests are sufficient, then we could merge this on the grounds it's simpler and have an outstanding todo to rerun an analysis when a crawl's been run with this fix in.
Seems good to me
Where would we track such a todo?

PR #726, opened 2020-08-07
Added scripts to set PYTHONUSERSITE

### LINKED ISSUE #689: Isolate a user's conda environment from python user packages in ~/.local
#### The problem

In #682 we discovered that conda doesn't provide a fully isolated execution environment, and instead allows python dependencies installed in `~/.local` to be picked up. Unfortunately these seems to be preferred to conda's own packages. Here's a simple test script:

```python
import crontab
from pprint import pprint as print
import sys

print(sys.path)
print(crontab.__file__)
```

When a conda environment isn't active, run `pip install --user crontab`. Then `conda activate openwpm`. The environment variable `PYTHONNOUSERSITE=True` will control whether python will pick up the users local packages.

```
(openwpm) steven@apollo:~/research/OpenWPM/test$ python test.py
['/home/steven/research/OpenWPM/test',
 '/home/steven/miniconda3/envs/openwpm/lib/python38.zip',
 '/home/steven/miniconda3/envs/openwpm/lib/python3.8',
 '/home/steven/miniconda3/envs/openwpm/lib/python3.8/lib-dynload',
 '/home/steven/.local/lib/python3.8/site-packages',
 '/home/steven/miniconda3/envs/openwpm/lib/python3.8/site-packages']
'/home/steven/.local/lib/python3.8/site-packages/crontab/__init__.py'
```

```
(openwpm) steven@apollo:~/research/OpenWPM/test$ PYTHONNOUSERSITE=True python test.py
['/home/steven/research/OpenWPM/test',
 '/home/steven/miniconda3/envs/openwpm/lib/python38.zip',
 '/home/steven/miniconda3/envs/openwpm/lib/python3.8',
 '/home/steven/miniconda3/envs/openwpm/lib/python3.8/lib-dynload',
 '/home/steven/miniconda3/envs/openwpm/lib/python3.8/site-packages']
'/home/steven/miniconda3/envs/openwpm/lib/python3.8/site-packages/crontab/__init__.py'
```

I don't know the exact internal logic python uses to order locations in `sys.path`. From [PEP 370](https://www.python.org/dev/peps/pep-0370/) it sounds like this is the intended order. This happens both with packages that are installed directly from conda's package repository and those installed via pip within conda.

#### Why fix

We shouldn't assume that users don't have local packages installed. It's the default flow for pip when global packages aren't writable (i.e., most linux users?). While virtual environments are best practice, I'm sure many users will pip install a few packages to run a simple script every once in a while (or are human and don't always follow best practices). In fact, OpenWPM's previous install script would intentionally install packages into `~/.local` (which in retrospect was a mistake) so this may impact a lot of our long-term users.

These issues aren't easy to debug either. As @vringar experienced when he hit the issue in: https://github.com/mozilla/OpenWPM/issues/614#issuecomment-630479738, `conda list` shows the conda packages so it's not immediately clear to the user.

When python dependencies are listed in a `requirements.txt` file, users can choose to run in a virtualenv or not, and if they choose not to they're taking things into their own hands. But that's not possible with `environment.yaml`, so it's on us to make sure things work.

#### What can we do

This is less clear. This feels like something that should be fixed in conda, but reading through the discussions in https://github.com/conda-forge/python-feedstock/issues/171, https://github.com/conda/conda/issues/7173, and https://github.com/conda/conda/issues/448 it seems like this is intended design despite the conda folks also billing conda as an [alternate to pip and venv](https://www.anaconda.com/blog/understanding-conda-and-pip).

We could try to export `PYTHONNOUSERSITE=True` in a conda helper script, but as Sarah points out this may have some issues https://github.com/mozilla/OpenWPM/pull/682#issuecomment-646203708.

We could also try to use both `virtualenv` and `conda`, which is suggested by a conda developer here https://github.com/conda/conda/issues/9788#issuecomment-604599393, but we'd have to refactor how we manage dependencies. Maybe this is the best long-term option.

vringar:
This also puts in place a more general infrastructure
should we ever become more opinionated and
automatically hook up things like pre-commit

Closes #689


birdsarah:
I don't get this, we already set PYTHONUSERSITE where it's needed to solve the issue: https://github.com/mozilla/OpenWPM/blob/master/scripts/repin.sh#L12-L15

Why is the existing solution insufficient.

birdsarah:
I've just seen the tread on matrix so am caught up on your motivations.

scripts/deploy_env_scripts.py
```diff
@@ -0,0 +1,35 @@
...
+    ```
+    conda activate openwpm
+    conda info --json | python deploy_env_scripts.py
+    conda activate openwpm
+    ```
+
+    This script aims to be plattform independent and idempotent
+"""
+
+CONDA_SCRIPT_DIR = "./etc/conda"
```
birdsarah, line 22: I still have a path in my head of how we can support windows. I think it would be good if we didn't add things that inhibit that possibility.

faith (changes requested):
Tests seem to fail too

vringar:
Yeah, this PR is pretty broken. Do you want to pick it up and play around with it?

PR #728, opened 2020-08-07
Renamed browser_pid to geckodriver_pid

vringar:


automation/TaskManager.py
```diff
@@ -243,20 +243,20 @@ def _manager_watchdog(self) -> None:
...
                     if browser.display_pid is not None:
                         display_pids.add(browser.display_pid)
                 for process in psutil.process_iter():
                     if process.create_time() + 300 < check_time and (
                             (process.name() == 'firefox' and (
-                                process.pid not in browser_pids)) or (
+                                process.pid not in geckodriver_pids)) or (
                             process.name() == 'Xvfb' and (
-                                process.pid not in browser_pids))):
+                                process.pid not in geckodriver_pids))):
```
vringar, line 259:
```suggestion
                                process.pid not in display_pids))):
```
Isn't this the thing we want to check here @birdsarah

automation/TaskManager.py
```diff
@@ -243,20 +243,20 @@ def _manager_watchdog(self) -> None:
...
                 for browser in self.browsers:
-                    if browser.browser_pid is not None:
-                        browser_pids.add(browser.browser_pid)
+                    if browser.geckodriver_pid is not None:
+                        geckodriver_pids.add(browser.geckodriver_pid)
                     if browser.display_pid is not None:
                         display_pids.add(browser.display_pid)
                 for process in psutil.process_iter():
                     if process.create_time() + 300 < check_time and (
                             (process.name() == 'firefox' and (
```
vringar, line 256: @englehardt seeing as this is the geckodriver process I don't think it'll ever be named `firefox`
```suggestion
                            (process.name() == 'geckodriver' and (
```

englehardt: @vringar There definitely could be processes named Firefox that aren't closed properly. I'm wondering whether we should add an additional check for geckodriver processes instead. Note that the geckodriver process spawns the firefox process, so we may want to do something like [this](https://github.com/mozilla/OpenWPM/blob/e42e79fcb3eb2cc6d08030208c66ccab3e3c0c2a/automation/BrowserManager.py#L367-L384).

automation/TaskManager.py
englehardt: Not @birdsarah, but yes. We want to use display_pids here.

birdsarah (approved):
Cool!

englehardt (approved):
Very smart refactoring. Thanks @vringar

PR #750, opened 2020-09-17
Command refactoring

### LINKED ISSUE #743: Refactor commands to actually follow the Command pattern
The [command pattern](https://en.wikipedia.org/wiki/Command_pattern)  describes how to encapsulate behaviour and the necessary state for this behaviour in a class, that is opaque to the caller.
My suggestion is as follows:
https://github.com/mozilla/OpenWPM/blob/1ad7d5a41f6c0e87173bcfa8838e894d373b5b48/automation/BrowserManager.py#L502-L504
gets replaced by
```python
 command.execute(driver, browser_settings, browser_params, manager_params, extension_socket)
```
This is enabled by removing the split between declaration of actions and their implementation.
For example the GetCommand turns into:
```python 
class GetCommand(BaseCommand):
    def __init__(self, url, sleep):
        self.url = url
        self.sleep = sleep

    def __repr__(self):
        return "GetCommand({},{})".format(self.url, self.sleep)

    def execute(webdriver, browser_settings, browser_params, manager_params, extension_socket: clientsocket):
    """
    goes to <url> using the given <webdriver> instance
    """

    tab_restart_browser(webdriver)

    if extension_socket is not None:
        extension_socket.send(this.visit_id)

    # Execute a get through selenium
    try:
        webdriver.get(this.url)
    except TimeoutException:
        pass

    # Sleep after get returns
    time.sleep(this.sleep)

    # Close modal dialog if exists
    try:
        WebDriverWait(webdriver, .5).until(EC.alert_is_present())
        alert = webdriver.switch_to_alert()
        alert.dismiss()
        time.sleep(1)
    except (TimeoutException, WebDriverException):
        pass

    close_other_windows(webdriver)

    if browser_params['bot_mitigation']:
        bot_mitigation(webdriver)

```
We could then also shrink the CommandSequence by removing 
https://github.com/mozilla/OpenWPM/blob/1ad7d5a41f6c0e87173bcfa8838e894d373b5b48/automation/CommandSequence.py#L64-L69
and replacing it with a generalized `add_command(self, command)`.
We'd also add a validate method, to check that we contain a get or browse before every other command.
This method would be called when a CommandSequence gets submitted to the TaskManager, where we would also add the IntitializeCommand and the FinalizeCommand.

This would simplify adding a new command to:
1. Create Class `MyCommand` that derives from `BaseCommand`.
2. Implement  `execute()` (and maybe `__repr__` and [eventually `name()`](https://github.com/mozilla/OpenWPM/issues/684))
3. Add `command_sequence.add_command(MyCommand())`

And we could remove the `RunCustomFunctionCommand` as it would be easier to just tell the user to create a new class instead.

vringar:
Closes #743


vringar:
TODO: Write Documentation on how to create a new command

docs/Using_OpenWPM.md
```diff
@@ -34,80 +34,42 @@ for _ in range(num_browsers):
...
-
-Finally, the command sequence given to the Task Manager to visit a site, sleep for 10 seconds, jiggle the mouse 10 times, and takes a screenshot would look like this:
-
+If you'd like to create a custom command, you must create a new file with the details
+of your command. Suppose we want to make a command to take a screenshot of the page we are viewing.
+In `custom_command.py` we would define a class, `SaveScreenshotCommand` that derives from `BaseCommand`
+```python
+class SaveScreenshotCommand(BaseCommand)
+```
+Next, in our newly derived class, we would define the `___init___`, `execute` and `__repr__` functions.
```
englehardt, line 43: Do these have a set interface? Do they need to accept a specific set of arguments? What's a `__repr__` function? 

These are the kinds of questions I'd expect someone new to have -- can we update this to point folks to the abstract base class (for a list of expected arguments) and also to provide guidance on the repr function, even if just a link out to some python docs.

englehardt (approved):
This is fantastic work. Great job @cyruskarsan and @vringar. I'm so glad that we now have a clean way to define commands.

@vringar what do you think about adding a small custom function to `demo.py` just to showcase how easy it now is to write one?

docs/Using_OpenWPM.md
vringar: We have the `BaseCommand` as an interface that defines the `execute` function. Defining `__repr__` is just a python best practice (I suppose we could like to a document saying that). And an `__init__` is just required in this specific case as we want to store some data in the command. I'm unsure how to document that.

englehardt: I would add something like `see the Base class (link) for required arguments` and link `__repr__` to python docs. Try to make it as easy as possible for the reader :)

vringar:
> @vringar what do you think about adding a small custom function to `demo.py` just to showcase how easy it now is to write one?

Somehow if the command gets created in a script (in contrast to getting created in a different module) it can't be sent through the `command_queue`. So I created an extra top-level file and made a comment to encourage people to take a look at it.

englehardt:
> > @vringar what do you think about adding a small custom function to `demo.py` just to showcase how easy it now is to write one?
> 
> Somehow if the command gets created in a script (in contrast to getting created in a different module) it can't be sent through the `command_queue`. So I created an extra top-level file and made a comment to encourage people to take a look at it.

Do you know why?

vringar:
> > Somehow if the command gets created in a script (in contrast to getting created in a different module) it can't be sent through the `command_queue`. So I created an extra top-level file and made a comment to encourage people to take a look at it.
> 
> Do you know why?

No, I haven't been able to figure out why. It's very weird and I could find nobody else mentioning anything similar.

I have also written some more documentation. Could you please re-review it.

englehardt:
Change simple_commands to custom_commands.

custom_command.py
```diff
@@ -0,0 +1,56 @@
...
+1. Create a class that derives from BaseCommand
+2. Implement the execute method
+3. Append it to the CommandSequence
+4. Execute the CommandSequence
+
+The following parameters get passed to the execute function:
+1. webdriver: WebDriver is a Selenium class used to control
+   browser. You can simulate arbitrary interactions and extract almost all browser state
+   with the tools that Selenium gives you
+2. browser_params: Configuration that might differ per browser
```
vringar, line 14: Move this into either the `BaseCommand.execute()` and `BrowserParams` and `ManagerParams`

docs/Using_OpenWPM.md
```diff
@@ -18,58 +18,25 @@ browser_params = [BrowserParams() for _ in range(manager_params.num_browsers)]
...
-            self.suffix = "-" + self.suffix
-
-        urlhash = md5(webdriver.current_url.encode("utf-8")).hexdigest()
-        outname = os.path.join(
-            manager_params["screenshot_path"],
-            "%i-%s%s.png" % (self.visit_id, urlhash, self.suffix),
-        )
-        webdriver.save_screenshot(outname)
-```
+Please have a look at [`custom_command.py`](../custom_command.py)
```
englehardt, line 37:
```suggestion
Please have a look at [`custom_command.py`](../custom_command.py). Note that custom commands must be
defined in a separate module and imported. They can't be defined within the main crawl script.
See [#837](https://github.com/mozilla/OpenWPM/issues/837).
```

englehardt (approved):

Issue #232, opened 2018-11-09
Add support for saving screenshots, page source, and other arbitrary files to data aggregators


englehardt:
Screenshots, page source, and other files collected in the browser manager process are currently written directly to disk. This worked when OpenWPM only saved data locally, but will not work for the S3Aggregator. Instead, `BaseAggregator` should include a `save_file` method. In `LocalAggregator` we can implement that to save to disk, and in `S3Aggregator` we can upload to S3.

vringar:
Updating this comment as #753 removed everything mentioned in the original issue.
Observations:
- UnstructuredStorageProviders already have an interface suitable for storing a bunch of bytes under a user-defined name
- The base path for storing is specified at time of object instantiation
- => There is no more need for a `data_directory` in the manager params similiar to the `database_name` name being removed in #753

Paths forward:
1. Add a second UnstructuredStorageProvider to the StorageController that is responsible for saving unstructured platform data
2. Expand the  UnstructuredStorageProvider interface with a second method that is responsible for saving unstructured platform data

I prefer option 1 as it is inherently more flexible, e.g. this way screenshots can get saved into the cloud while web content just gets saved to disk.

PR #753, opened 2020-09-29
Data Aggregator Rewrite

### LINKED ISSUE #561: Rewrite DataAggregators

### LINKED ISSUE #652: Remove dependency on localstack / refactor s3aggregator to parquet aggregator

### LINKED ISSUE #684: Command table should contain short command names
It shouldn't be the full type name in the table but so instead of automation.commands...GetCommand

vringar:
This rewrite modularizes the storage provider setup to make it more flexible and easier to extend.
It also aims to reduce the coupling between how the extension outputs data and how it's passed to storage
Closes #561, Closes #652, Closes #684


vringar:
Closes #684

automation/data_aggregator/storage_providers.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+class UnstructuredStorageProvider(StorageProvider):
+    @abstractmethod
+    def store_blob(
+        self,
+        filename: str,
+        blob: bytes,
+        compressed: bool = True,
+        skip_if_exists: bool = True,
```
vringar, line 75: We need to discuss this paramter further

automation/data_aggregator/storage_providers.py
```diff
@@ -0,0 +1,89 @@
...
+        records for this visit_id will be submitted
+        """
+        pass
+
+
+class UnstructuredStorageProvider(StorageProvider):
+    @abstractmethod
+    def store_blob(
+        self,
+        filename: str,
```
vringar, line 72:
```suggestion
        hash: str,
```
Maybe we should verify that hash (optionally)...

automation/data_aggregator/storage_providers.py
vringar:
```suggestion
        override: bool = False,
```

automation/data_aggregator/storage_providers.py
```diff
@@ -0,0 +1,89 @@
...
+        blob: bytes,
+        compressed: bool = True,
+        skip_if_exists: bool = True,
+    ) -> None:
+        """Stores the given bytes under the provided filename"""
+        pass
+
```
vringar, line 79:
```suggestion
@abstractmethod
def get_blob_locator(hash:str):
    pass
```

automation/data_aggregator/__init__.py
```diff
@@ -0,0 +1,348 @@
...
+RECORD_TYPE_CREATE = "create_table"
+STATUS_TIMEOUT = 120  # seconds
+SHUTDOWN_SIGNAL = "SHUTDOWN"
+BATCH_COMMIT_TIMEOUT = 30  # commit a batch if no new records for N seconds
+
+
+STATUS_UPDATE_INTERVAL = 5  # seconds
+
+
+class DataAggregator:
```
vringar, line 30:
```suggestion
class StorageController:
```

automation/data_aggregator/__init__.py
```diff
@@ -0,0 +1,348 @@
...
+        aggregator.update_status_queue()
+        aggregator.update_completion_queue()
+        aggregator.save_batch_if_past_timeout()
+        aggregator.poll_queue()
+
+    aggregator.drain_queue()
+    aggregator.shutdown()
+
+
+class DataAggregatorHandle:
```
vringar, line 221:
```suggestion
class StorageControllerHandle:
```

test/test_profile.py
```diff
@@ -83,40 +88,43 @@ def test_profile_saved_when_launch_crashes(self):
...
-            "SELECT * FROM crawl_history;",
+            var prefs = Components
+                        .classes["@mozilla.org/preferences-service;1"]
+                        .getService(Components.interfaces.nsIPrefBranch);
+            try {
+                return prefs.getBoolPref("test_pref")
+            } catch (e) {
+                return false;
+            }
+        """
```
englehardt, line 106:
```suggestion
            """
```

test/test_simple_commands.py
```diff
@@ -78,404 +87,405 @@
...
-    This entire test class is parametrized to run against
-    both headless and xvfb modes to ensure both are exercized
-    during the test suite and there are no obvious problems.
+scenarios = [
+    pytest.param("headless", id="headless"),
+    pytest.param("xvfb", id="xvfb"),
+]
+
+
+@pytest.mark.parametrize("display_mode", scenarios)
```
englehardt, line 96: Is there a reason we can't use this same parameterize shortcut in `test_memory_storage_provider.py`?

englehardt (approved):
Again, great work on this!

crawler.py
```diff
@@ -74,29 +85,41 @@
         browser_params[i].prefs = json.loads(PREFS)
 
 # Manager configuration
-manager_params.data_directory = "~/Desktop/%s/" % CRAWL_DIRECTORY
-manager_params.log_directory = "~/Desktop/%s/" % CRAWL_DIRECTORY
+manager_params.data_directory = Path("~/Desktop/") / CRAWL_DIRECTORY
+manager_params.log_directory = Path("~/Desktop/") / CRAWL_DIRECTORY
 manager_params.output_format = "s3"
```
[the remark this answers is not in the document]
vringar: Yes, unfortunately it still is. We want to address that as part of #232

openwpm/socket_interface.py
```diff
@@ -166,15 +162,46 @@ def close(self):
...
+    b = b""
+    while True:
+        try:
+            return await reader.readexactly(n)
+        except IncompleteReadError as e:
+            b += e.partial
+            n -= len(e.partial)
+            if reader.at_eof():
+                raise IOError("Socket Connection closed")
+
```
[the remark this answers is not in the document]
vringar: Yes, in the storage controller.

openwpm/storage/cloud_storage/s3_storage.py
```diff
@@ -0,0 +1,6 @@
+from ..arrow_storage import ArrowProvider
+from ..storage_providers import UnstructuredStorageProvider
+
+
+class S3StorageProvider(ArrowProvider):
+    pass
```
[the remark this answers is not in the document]
vringar: This class can't be instantiated since it's missing the abstract method, since it doesn't implement the `write_table` method.

test/openwpmtest.py
```diff
@@ -1,62 +1,71 @@
...
 
 import pytest
 
 from openwpm import task_manager
 from openwpm.config import BrowserParams, ManagerParams
+from openwpm.storage.sql_provider import SqlLiteStorageProvider
 
 from . import utilities
 
+NUM_BROWSERS = 2
```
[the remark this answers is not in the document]
vringar: Yes, I wanted to make sure our tests need to explicitly limit themselves to one browser.
I was hoping this would uncover more errors.

test/storage/test_gcp.py
```diff
@@ -0,0 +1,34 @@
...
+from openwpm.storage.cloud_storage.gcp_storage import GcsStructuredProvider
+from openwpm.storage.storage_providers import TableName
+from openwpm.types import VisitId
+from test.storage.test_values import TEST_VALUES, TEST_VISIT_IDS
+
+
+@pytest.mark.skip
+@pytest.mark.asyncio
+async def test_gcp_structured(mp_logger):
+    project = "senglehardt-openwpm-test-1"
```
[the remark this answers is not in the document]
vringar: Yeah, this was mostly so I could keep the implementation around.

.travis.yml
```diff
@@ -36,15 +36,11 @@ after_success:
...
       env:
         - TESTS="Docker"
       services:
         - docker
-      before_install:
-      before_script:
-      install:
       script:
         - docker build -f Dockerfile -t openwpm .
         - ./scripts/deploy-to-dockerhub.sh
```
boolean5, line 46: I think this is causing the failure in building the Docker image on Travis. Removing the `before_install:`, `install:`, and `before_script:` sections from the Docker job makes Travis use these sections as they were specified [above](https://github.com/mozilla/OpenWPM/blob/master/.travis.yml#L17L28) and unecessarily installs everything before building the Docker image like it does for the rest of the test jobs.

If we put these back it should be alright.

vringar: Thank you for pointing this out! I'll revert the changes and check.

vringar: This worked! Thanks for pointing this out.

test/test_custom_function_command.py
```diff
@@ -22,66 +26,66 @@
...
+                    "browser_id": browser_id,
+                },
             )
-            sock.send(("create_table", query))
+            sock.send(query)
+        sock.close()
+
+    manager_params, browser_params = default_params
+
+    db = sqlite3.connect(manager_params.database_name)
```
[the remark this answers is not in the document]
vringar: I have linked to it in the error message when they try to create a table the old fashioned way.

openwpm/storage/cloud_storage/s3_storage.py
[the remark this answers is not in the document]
vringar: Also #823 should be addresses before the next release so I don't think we need to many precautions around this. (I just added it to the v0.14.0 milestone)

test/test_simple_commands.py
vringar: We can. And it's a lot cleaner that way. Thanks for pointing this out.

vringar:
TODO: Figure out why failed visits don't get reported as such.

vringar:
TODO Need to fix path encoding in marshmellow / dataclass-json

openwpm/storage/storage_controller.py
```diff
@@ -0,0 +1,454 @@
...
+            await self.unstructured_storage.flush_cache()
+            await self.unstructured_storage.shutdown()
+
+    async def should_shutdown(self) -> None:
+        """Returns when we should shut down"""
+
+        while self.shutdown_queue.empty():
+            await asyncio.sleep(STATUS_UPDATE_INTERVAL)
+        _, relaxed = self.shutdown_queue.get()
+        self._relaxed = relaxed
```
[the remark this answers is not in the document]
vringar: #847 addresses this issue

test/utilities.py
```diff
@@ -5,7 +5,6 @@
 from os.path import dirname, realpath
 from urllib.parse import parse_qs, urlparse
 
-import boto3
 import pyarrow.parquet as pq
 import s3fs
 from botocore.credentials import Credentials
```
boolean5, line 10: We don't need these imports anymore. Nor the `S3FSWrapper` import right below

boolean5 (commented):
Impressive work!  :rocket: I left some minor suggestions and might add a couple more tomorrow.

We should also update the ["Persistence Types"](https://github.com/mozilla/OpenWPM/blob/master/README.md#persistence-types) section of the `README` and the ["Data Aggregator"](https://github.com/mozilla/OpenWPM/blob/master/docs/Platform-Architecture.md#data-aggregator) section of `docs/Platform-Architecture.md` as they still reference S3 and `DataAggregator`.

And there are a few more references to the data aggregator in comments: [1](https://github.com/mozilla/OpenWPM/blob/6527179d11ef4f96eeed5ee752a02455d64be0bd/openwpm/Extension/firefox/feature.js/loggingdb.js#L197), [2](https://github.com/mozilla/OpenWPM/blob/6527179d11ef4f96eeed5ee752a02455d64be0bd/openwpm/Extension/webext-instrumentation/src/background/dns-instrument.ts#L83), [3](https://github.com/mozilla/OpenWPM/blob/6527179d11ef4f96eeed5ee752a02455d64be0bd/test/test_profile.py#L77)

environment.yaml
```diff
@@ -7,37 +7,34 @@ dependencies:
...
 - pyarrow=2.0.0
+- pytest-asyncio=0.14.0
 - pytest-cov=2.10.1
 - pytest=6.2.1
-- python=3.8.6
+- python=3.9.1
 - pyvirtualdisplay=1.3.2
 - redis-py=3.5.3
-- s3fs=0.4.0
+- s3fs=0.5.2
```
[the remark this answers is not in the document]
vringar: Per #823 we want to (re-)implement the ability to save out to S3. So while you're correct and we don't need it in this PR anymore, we still need the dependency in general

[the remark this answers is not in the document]
vringar: I have removed the unused imports in `test/utilities.py` .

openwpm/commands/profile_commands.py
```diff
@@ -19,69 +24,56 @@
...
-    def __init__(self, dump_folder, close_webdriver, compress):
-        self.dump_folder = dump_folder
+    def __init__(self, tar_path: Path, close_webdriver: bool, compress: bool) -> None:
+        self.tar_path = tar_path
         self.close_webdriver = close_webdriver
         self.compress = compress
+        raise ConfigError(
+            "BROWSER %i: Profile dumping is currently unsupported. "
+            "See: https://github.com/mozilla/OpenWPM/projects/2." % self.browser_id
+        )
```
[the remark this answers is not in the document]
vringar: You're right we can't access the `browser_id` here and the `NotImplementedError` is more appropriate.
However I want to raise this error in `__init__` to make sure the object creation fails, since that usually happens in the main thread.
This way the user is immediately informed that this command won't run instead of having to wait until the browser tries to execute it.

openwpm/utilities/platform_utils.py
```diff
@@ -95,8 +97,20 @@ def get_configuration_string(manager_params, browser_params, versions):
...
     config_str += "\n========== Manager Configuration ==========\n"
+
+    def serialize_paths(obj: Any) -> Any:
+        if isinstance(obj, Path):
+            return str(obj)
+        raise TypeError(
+            f"Object of type {obj.__class__.__name__} is not JSON serializable"
+        )
+
     config_str += json.dumps(
```
vringar, line 108: TODO use ConfigEncoder

test/conftest.py
```diff
@@ -12,29 +21,97 @@
...
-        server_thread.join()
 
-    request.addfinalizer(local_http_server_stop)
+@pytest.fixture()
+def mp_logger(tmp_path: Path) -> Generator[MPLogger, Any, None]:
+    log_path = tmp_path / "openwpm.log"
+    logger = MPLogger(log_path, log_level_console=logging.DEBUG)
+    yield logger
+    logger.close()
+    # The performance hit for this might be unacceptable but it might help us discover bugs
```
vringar, line 114: This costs us 15s per test. Is that too much?

test/storage/__init__.py
```diff
@@ -0,0 +1,3 @@
+"""TODO: Write tests for a correctly set interrupt status
```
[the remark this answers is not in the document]
vringar: Tracked in #852

openwpm/commands/profile_commands.py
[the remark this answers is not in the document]
boolean5: That makes sense. How about we just don't include the `browser_id` in the error message then?

openwpm/socket_interface.py
```diff
@@ -166,15 +162,46 @@ def close(self):
...
+async def get_n_bytes_from_reader(reader: asyncio.StreamReader, n: int) -> bytes:
+    b = b""
+    while True:
+        try:
+            return await reader.readexactly(n)
+        except IncompleteReadError as e:
+            b += e.partial
+            n -= len(e.partial)
+            if reader.at_eof():
+                raise IOError("Socket Connection closed")
```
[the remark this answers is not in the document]
vringar: I don't think so. We always ask for the minimum amount of bytes we need to take the next step in parsing the input.
So if the connection gets broken without having transmitted the n bytes, then all received data is garbage data.

openwpm/config.py
```diff
@@ -68,65 +81,74 @@ class BrowserParams:
...
     It can be used to control storage locations or which watchdogs should
     run
     """
 
-    data_directory: str = "~/openwpm/"
-    log_directory: str = "~/openwpm/"
-    screenshot_path: Optional[str] = None
-    source_dump_path: Optional[str] = None
-    output_format: str = "local"
-    database_name: str = "crawl-data.sqlite"
```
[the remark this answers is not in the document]
vringar: I removed all references

test/conftest.py
vringar: https://github.com/mozilla/OpenWPM/pull/753/checks?check_run_id=1838971116#step:12:2705

openwpm/socket_interface.py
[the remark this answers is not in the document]
boolean5: I see. I'm still a bit confused though, because variable `b` is updated but doesn't seem to be used anywhere. Am I missing something?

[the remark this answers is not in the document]
vringar: You're right. I completely missed that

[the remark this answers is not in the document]
vringar: TODO: Remove this function

openwpm/storage/cloud_storage/gcp_storage.py
```diff
@@ -0,0 +1,108 @@
...
+    ) -> None:
+        assert self.file_system is not None
+        target_path = self.base_path.format(filename=filename)
+        if not overwrite and (
+            filename in self.file_name_cache or self.file_system.exists(target_path)
+        ):
+            self.logger.info("Not saving out file %s as it already exists", filename)
+            return
+
+        with self.file_system.open(target_path, mode="wb") as f:
```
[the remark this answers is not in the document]
englehardt: I assume this is resolved since we now have transactions? @vringar

openwpm/storage/arrow_storage.py
```diff
@@ -97,63 +96,76 @@ def _is_cache_full(self) -> bool:
...
         to pass us the locked storing_lock
         """
-        assert self.storing_lock is not None
-        assert self.flush_event is not None
-        got_cond = cond is not None
-        if not got_cond:
-            cond = self.storing_lock
-            await cond.acquire()
-        assert cond is not None and cond.locked()
+        got_lock = lock is not None
```
englehardt, line 145:
```suggestion
        has_lock_arg = lock is not None
```

openwpm/storage/arrow_storage.py
```diff
@@ -97,63 +96,76 @@ def _is_cache_full(self) -> bool:
...
+        if not got_lock:
```
englehardt, line 146:
```suggestion
        if not has_lock_arg:
```

openwpm/storage/arrow_storage.py
```diff
@@ -97,63 +96,76 @@ def _is_cache_full(self) -> bool:
...
-        self.flush_event.set()
-        if not got_cond:
-            cond.release()
+        self._batches.clear()
+
+        for event in self.flush_events:
+            event.set()
+        self.flush_events.clear()
+
+        if not got_lock:
```
englehardt, line 161:
```suggestion
        if not has_lock_arg:
```

I've updated the arg name here because I was a little confused by `got_lock` (i.e. after we locked locally I assumed we'd expect this to be true).

openwpm/storage/parquet_schema.py
```diff
@@ -10,6 +10,25 @@
 
 PQ_SCHEMAS = dict()
 
+fields = [
+    pa.field("task_id", pa.int64(), nullable=False),
+    pa.field("manager_params", pa.string(), nullable=False),
+    pa.field("openwpm_version", pa.string(), nullable=False),
+    pa.field("browser_version", pa.string(), nullable=False),
+    pa.field("instance_id", pa.uint32(), nullable=False),
+    # Omitting start_time as I couldn't figure out how to implement it
```
englehardt, line 19: We should remove this comment and file a bug if we think it's important.

I don't think it is important, and I'm happy to just drop it from `schema.sql` as well to keep the schemas consistent. The start time is self-evident by the timestamps on any of the collected data.

openwpm/task_manager.py
```diff
@@ -173,6 +165,18 @@ def __init__(
         self.callback_thread.name = "OpenWPM-completion_handler"
         self.callback_thread.start()
 
+    def __enter__(self):
```
englehardt, line 168: Why are these needed?

openwpm/task_manager.py
```diff
@@ -620,16 +603,17 @@ def execute_command_sequence(
         int  -> index of browser to send command to
         """
 
-        # Block if the aggregator queue is too large
-        agg_queue_size = self.storage_controler_handle.get_most_recent_status()
-        if agg_queue_size >= AGGREGATOR_QUEUE_LIMIT:
-            while agg_queue_size >= AGGREGATOR_QUEUE_LIMIT:
+        # Block if the storage controller has too many unfinished records
+        # TODO create tests that these numbers are still meaningful
```
englehardt, line 607: I suggest we remove this comment and create an issue. Otherwise it will be lost.

test/conftest.py
englehardt: What's the mean in terms of total test suite runtime? In general I think it's okay to trade off some test time if it means that we'll catch errors that we'd otherwise miss.

test/storage/test_gcp.py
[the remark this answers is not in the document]
englehardt: Do you plan to remove this? We shouldn't land it with test values.

englehardt (changes requested):

openwpm/socket_interface.py
```diff
@@ -163,24 +163,12 @@ def close(self):
 
 
 async def get_message_from_reader(reader: asyncio.StreamReader) -> Any:
-    msg = await get_n_bytes_from_reader(reader, 5)
+    msg = await reader.readexactly(5)
```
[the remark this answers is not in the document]
englehardt: I see that we rely on the caller to check if at eof before reading. I wonder w

(Also you should remove the `IncompleteReadError` import)

openwpm/mp_logger.py
```diff
@@ -207,16 +207,6 @@ def _initialize_sentry(self):
         )
         self._event_handler = EventHandler(level=self._log_level_sentry_event)
         sentry_sdk.init(dsn=self._sentry_dsn, before_send=self._sentry_before_send)
-        with sentry_sdk.configure_scope() as scope:
```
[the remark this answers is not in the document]
vringar: Because we no longer have easy access to the the storage config (as it's contained in the storage providers) but I think I found a way around it by just using `__str__`  which should be a "nicely printable string representation of an object".

openwpm/socket_interface.py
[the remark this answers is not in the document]
vringar: I removed the check for EOF and instead am wrapping `get_message_from_reader` in  a try-except-block.

docs/Configuration.md
```diff
@@ -53,19 +58,21 @@ validate_manager_params(manager_params)
 # Platform Configuration Options
 
 * `data_directory`
-  * The directory in which to output the crawl database and related files. The
-    directory given will be created if it does not exist.
+  * The directory into which screenshots and page dumps will be saved
+  * [Intended to be removed by #232](https://github.com/mozilla/OpenWPM/issues/232)
```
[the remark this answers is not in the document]
vringar: I think we shouldn't have a path for this in the BrowserParams. Instead we should either add a second `UnstructuredStorageProvider` to the `StorageController` or adding another method to the `UnstructuredStorageProvider` (I prefer option 1 as this way screenshots can get saved into the cloud while web content just gets saved to disk)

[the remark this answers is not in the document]
vringar: I commented some more in the issue

openwpm/task_manager.py
vringar: These turn the TaskManager into a context manager, making it harder to forget to shut it down.

docs/Configuration.md
```diff
@@ -11,20 +11,25 @@ To load the default configuration create instances of `config::ManagerParams` an
...
+from openwpm.config import (
+  validate_browser_params, 
+  validate_manager_params,
+  validate_crawl_configs,
+)
 
 for bp in browser_params:
   validate_browser_params(bp)
 validate_manager_params(manager_params)
+validate_crawl_configs(manager_params, browser_params)
```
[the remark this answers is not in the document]
vringar: No, and that's a lot nicer. I moved it there

openwpm/task_manager.py
vringar: I filed https://github.com/mozilla/OpenWPM/issues/861 and removed the comment.

test/conftest.py
vringar: Hmm, running a super unprofessional benchmark, the test suite running without this check was 20s slower :sweat_smile: 
29:03 (with this check) vs 29:30 (with this check commented out)

openwpm/storage/cloud_storage/gcp_storage.py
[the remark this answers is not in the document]
vringar: Yes

openwpm/storage/parquet_schema.py
vringar: I think it's convinient to have this information. E.g. if you want to measure how websites change over time, being able to group by day the crawl ran, seems like something you'd want to have accessible.

vringar: https://github.com/mozilla/OpenWPM/issues/862

vringar:
I think I have addressed all outstanding comments.
The merge request UI is slowy but surely breaking under the amount of commits and files changed.
So I'm proposing to merge this and fix all followup issues in new PRs.
With two :+1: on this comment I'll merge it.

englehardt (approved):

PR #757, opened 2020-10-05
Command refactoring

cyruskarsan:
@vringar I spent a while trying to understand your code and here's what I came up with (it will certainly need revision)
It looks like for the other classes you created, i.e. `GetCommand` and `BrowseCommand`, the only variables to be printed by repr method were to be instantiated of the class. In the `SaveScreenShotCommand` , when running `demo.py` I did not see it called anywhere and hence, did not know how to which variables to instaitate.

This leads me to a major issue I encountered. After unsuccessfully attempting to debug the program, I read that you must build `./scripts/build-extension.sh` and then run `python -m test.manual_test`
I did this and the program stopped (according to the terminal) but I did not know how to advance it. 
I've got a working copy of the repo in VScode and was hoping to set breakpoints throughout the program to see where certain methods (`saveScreenshot`) were being called. Do you have any advice? Thank you.

### DIFF
--- a/automation/Commands/browser_commands.py
+++ b/automation/Commands/browser_commands.py
@@ -227,7 +227,13 @@ class BrowseCommand(BaseCommand):
                 pass
 
 
-class SaveScreenShotCommand(BaseCommand):
+class SaveScreenshotCommand(BaseCommand):
+    def __init__(self, suffix): 
+         self.suffix = suffix 
+  
+    def __repr__(self): 
+        return "SaveScreenshotCommand({})".format(self.suffix)
+
     def execute(
         self,
         webdriver,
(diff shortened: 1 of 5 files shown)


automation/Commands/browser_commands.py
```diff
@@ -227,16 +227,35 @@ def execute(
...
+    def execute(
+        self,
+        webdriver,
+        browser_settings,
+        browser_params,
+        manager_params,
+        extension_socket,
+    ):
+        if suffix != "":
+            suffix = "-" + suffix
```
vringar, line 240:
```suggestion
        if self.suffix != "":
            self.suffix = "-" + suffix
```
since you don't get passed the suffix as a parameter but instead have to save it at construction time.

vringar:
First of, thanks for following through on your submission.
You can see how the `SaveScreenshotCommand` is instantiated here:
https://github.com/mozilla/OpenWPM/blob/a811af1c9c1f8efa4b9d2a008c6cbb3e5be6d1e8/automation/Commands/Types.py#L57-L62
or here
https://github.com/mozilla/OpenWPM/blob/a811af1c9c1f8efa4b9d2a008c6cbb3e5be6d1e8/automation/CommandSequence.py#L103-L112

When it comes to debugging I'd suggest you'd modify `demo.py` by changing
https://github.com/mozilla/OpenWPM/blob/a811af1c9c1f8efa4b9d2a008c6cbb3e5be6d1e8/demo.py#L53-L54
to
```python
command_sequence.get(sleep=3, timeout=60)
command_sequence.save_screenshot()
```
and try to run `demo.py` and see what happens.

As for using breakpoints I'm afraid that OpenWPM is very hard to debug this way as it uses multiple process and threads which most debuggers can't handle. I usually debug by running the code, seeing how it crashes, and if that isn't enough adding some more logging just before it crashes.

vringar:
@cyruskarsan could you please use [black](https://github.com/psf/black) to format your code. Otherwise our CI will reject it and we can't see if the tests are passing or not.

cyruskarsan:
@vringar Thank you! reformatted the code with black. Just wanted to see if I did the first function correct before moving onto the others.

vringar:
@cyruskarsan can I just merge this and you'll create a new PR for each new Command?
This way we can also split this between multiple contributors unless you want to port every command by yourself.

cyruskarsan:
@vringar Sounds good, I can create a PR for each new command

Issue #759, opened 2020-10-06
Running install.sh gives ResolvePackageNotFound error in Ubuntu 18.04


Prajwal7842:
I tried installing the OpenWPM repo on my OS : Ubuntu 18.04. When I run the file ./install.sh, the terminal gave me this error given in the picture
![2020-10-06-1602000034_screenshot_1920x1080](https://user-images.githubusercontent.com/34996624/95227042-40d87980-081b-11eb-8569-159c8968bffc.jpg)

Also I tried installing it in gitpod. I got the same error. 

My temporary solution : I removed the version from environment.yaml file. (geckodriver=0.27.0  -->  geckodriver). This worked and the repo was install successfully. 
I already tried to find any existing issue regarding this but could not find it(if it existed).

vringar:
This is super weird given that we test against bionic (18.04) on Travis 
https://github.com/mozilla/OpenWPM/blob/a811af1c9c1f8efa4b9d2a008c6cbb3e5be6d1e8/.travis.yml#L2
but I have had the same problem before under OpenSuSe (see #742)
Would you mind creating a branch where you run `repin.sh` in the `scripts` folder to update all of our dependencies?
If the test pass on there, I'll merge it and nobody should have this problem anymore.

Prajwal7842:
okay, I can create a new branch and run repin.sh file. I just wanted to know is there any guidelines that are followed like some naming conventions that should be used to name a branch ?

vringar:
Closing this as #760 fixed the problem.

PR #760, opened 2020-10-07
Update environment.yaml

Prajwal7842:
@vringar 
This is related to the issue #759 . I ran repin.sh in a new branch. After running this, I tested the installation on my OS : Ubuntu 18.04 and also on gitpod. Now both seems to be working fine. Installation, now does not cause the error mentioned in the issue.

### DIFF
--- a/environment.yaml
+++ b/environment.yaml
@@ -3,32 +3,32 @@ channels:
 - main
 dependencies:
 - autopep8=1.5.4
-- beautifulsoup4=4.9.1
+- beautifulsoup4=4.9.3
 - click=7.1.2
 - codecov=2.1.9
 - dill=0.3.2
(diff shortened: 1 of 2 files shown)


environment.yaml
```diff
@@ -3,40 +3,40 @@ channels:
...
+- beautifulsoup4=4.9.3
 - click=7.1.2
 - codecov=2.1.9
 - dill=0.3.2
 - flake8-isort=4.0.0
-- flake8=3.8.3
-- geckodriver=0.27.0
-- ipython=7.17.0
+- flake8=3.8.4
+- geckodriver=0.26.0
```
vringar, line 12: @birdsarah do you have any idea why repining would have us switch to an older version of geckodriver?

Issue #665, opened 2020-05-22
Use new fullscreen screenshot command


birdsarah:
Geckodriver now supports full page screenshots (https://github.com/SeleniumHQ/selenium/pull/7182)

So https://github.com/mozilla/OpenWPM/blob/aa5e2a6fe628e67953016f79fddd6e9c9096ac7f/automation/CommandSequence.py#L106-L130 can be updated to work with moden geckodriver.

vringar:
This feature is not yet available to us, because it's currently only in the alpha branches of the 4.0 version of selenium.
We are currently on version 3.141

PR #762, opened 2020-10-08
Update geckodriver version numbers

### LINKED ISSUE #651: Update references to geckodriver 0.15
There are a couple of places that document the limitations of geckodriver 0.15. As we're now on 0.26, we should check these references and update them or update the code if the limitation is no longer present.

MLJBrackett:
Update the references of geckodriver 0.15 -> 0.26 & Check if new implementation is needed

Fixes #651

### DIFF
--- a/automation/CommandSequence.py
+++ b/automation/CommandSequence.py
@@ -130,7 +130,7 @@ class CommandSequence:
         open) seems to break element-only screenshots. So using this command
         will cause any future element-only screenshots to be mis-aligned
 
-        NOTE: In geckodriver v0.24 implemented a full screenshot feature, this
+        NOTE: In geckodriver v0.24 the full screenshot feature was implemented, this
         code is still using the above method.
         """
         self.total_timeout += timeout
--- a/automation/CommandSequence.py
+++ b/automation/CommandSequence.py
@@ -114,7 +114,7 @@ class CommandSequence:
     def screenshot_full_page(self, suffix="", timeout=30):
         """Save a screenshot of the entire page.
 
-        NOTE: geckodriver v0.26 only supports viewport screenshots. To
+        NOTE: geckodriver v0.15 only supports viewport screenshots. To
         screenshot the entire page we scroll the page using javascript and take
         a viewport screenshot at each location. This method will save the
         parts and a stitched version in the `screenshot_path`. We only scroll
@@ -126,9 +126,12 @@ class CommandSequence:
         the page is scrolled (i.e. infinite scroll) will only go as far as the
         original height.
 
-        NOTE: In geckodriver v0.26 doing any scrolling (or having devtools
+        NOTE: In geckodriver v0.15 doing any scrolling (or having devtools
         open) seems to break element-only screenshots. So using this command
         will cause any future element-only screenshots to be mis-aligned
+
+        NOTE: In geckodriver v0.24 implemented a full screenshot feature, this
+        code is still using the above method.
         """
         self.total_timeout += timeout
         if not self.contains_get_or_browse:
--- a/automation/CommandSequence.py
+++ b/automation/CommandSequence.py
@@ -114,7 +114,7 @@ class CommandSequence:
     def screenshot_full_page(self, suffix="", timeout=30):
         """Save a screenshot of the entire page.
 
-        NOTE: geckodriver v0.15 only supports viewport screenshots. To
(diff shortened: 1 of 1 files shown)


MLJBrackett:
As of right now there are 9 references of "geckodriver" and 4 references of "geckodriver 0.15"

### Geckodriver references:

1. selenium_firefox.py
2. BrowserManager.py
3. CHANGELOG.md
4. automation/TaskManager.py
5. automation/CommandSequence.py
6. scripts/environment-unpinned.yaml
7. environment.yaml
8. README.md
9. docs/OpenWPM.svg

**Out of the 9 there are only two files that will need to be updated**

5. automation/CommandSequence.py
8. README.md

### CommandSequence.py

``` python
def screenshot_full_page(self, suffix="", timeout=30):
        """Save a screenshot of the entire page.

        NOTE: geckodriver v0.15 only supports viewport screenshots. To
        screenshot the entire page we scroll the page using javascript and take
        a viewport screenshot at each location. This method will save the
        parts and a stitched version in the `screenshot_path`. We only scroll
        vertically, so pages that are wider than the viewport will be clipped.
        See: https://github.com/mozilla/geckodriver/issues/570

        The screenshot produced will only include the area originally
        loaded at the start of the command. Sites which dynamically expand as
        the page is scrolled (i.e. infinite scroll) will only go as far as the
        original height.

        NOTE: In geckodriver v0.15 doing any scrolling (or having devtools
        open) seems to break element-only screenshots. So using this command
        will cause any future element-only screenshots to be mis-aligned
        """
        self.total_timeout += timeout
        if not self.contains_get_or_browse:
            raise CommandExecutionError(
                "No get or browse request preceding " "the dump page source command",
                self,
            )
        command = ScreenshotFullPageCommand(suffix)
        self._commands_with_timeout.append((command, timeout))
```
I don't know as of right now if there needs to be changes to anything with the above code, but with the updated version of geckodriver there is now a command to get the entire screen.

The original implementation PR here: https://github.com/mozilla/OpenWPM/pull/156
The release that added the new function: https://github.com/mozilla/geckodriver/releases/tag/v0.24.0

It would also seem that there is already an open issue for updating this code.
Issue here: https://github.com/mozilla/OpenWPM/issues/665

So either the code needs to be updated or we can remove the two 'NOTE:' comments in the code

### README.md

``` markdown
* Screenshots
    * Selenium 3 can be used to screenshot an individual element. None of the
        built-in commands offer this functionality, but you can use it when
        [writing your own](https://github.com/citp/OpenWPM/wiki/Platform-Demo#adding-a-new-command). See the [Selenium documentation](https://seleniumhq.github.io/selenium/docs/api/py/webdriver_remote/selenium.webdriver.remote.webelement.html?highlight=element#selenium.webdriver.remote.webelement.WebElement.screenshot).
    * Viewport screenshots (i.e. a screenshot of the portion of the website
        visible in the browser's window) are available with the
        `CommandSequence::save_screenshot` command.
    * Full-page screenshots (i.e. a screenshot of the entire rendered DOM) are
        available with the `CommandSequence::screenshot_full_page` command.
        * This functionality is not yet supported by Selenium/geckodriver,
          though [it is planned](https://github.com/mozilla/geckodriver/issues/570).
          We produce screenshots by using JS to scroll the page and take a
          viewport screenshot at each location. This method will save the parts
          and a stitched version in the `screenshot_path`.
        * Since the screenshots are stitched they have some limitations:
            * On the area of the page present when the command is called will
              be captured. Sites which dynamically expand when scrolled (i.e.,
              infinite scroll) will only go as far as the original height.
            * We only scroll vertically, so pages that are wider than the
              viewport will be clipped.
            * In geckodriver v0.15 doing any scrolling (or having devtools
              open) seems to break element-only screenshots. So using this
              command will cause any future element-only screenshots to be
              misaligned.
```

The above quote references how Full-page screenshots are not currently supported.

@birdsarah Seeing that you are the poster of the issue #651 & #665 could you direct me on how to move forward with my findings.

MLJBrackett:
I have just taken a look at #665 and contributor vringar has stated that we don't have the functionality yet to implement the code for a full screenshot yet since we are on an outdated version of selenium https://github.com/mozilla/OpenWPM/issues/665#issuecomment-706963399. For now I've added another note stating that the functionality is now included in geckodriver 0.24 but the code is still using the stitching method.

PR #752, opened 2020-09-26
Update all request_id types to integer.

richardgreg:
Fixes: #748

Signed-off-by: Richard O. Gregory <richardgrecoson@gmail.com>

PR #765, opened 2020-10-12
Change request_id to integer

### LINKED ISSUE #748: Different types for `request_id` in parquet_schema.py
https://github.com/mozilla/OpenWPM/blob/181920fb229b708ec6b4010b9f85dab66d9baff5/automation/DataAggregator/parquet_schema.py#L47
https://github.com/mozilla/OpenWPM/blob/181920fb229b708ec6b4010b9f85dab66d9baff5/automation/DataAggregator/parquet_schema.py#L194
The only place where we use int is in the callstacks instrument. But it also should be an int, because it's obviously a number.
Like 25 or smt.

vringar:
Fixes #748
Subsumes #752


englehardt (approved):
LGTM. Thanks for fixing!

PR #770, opened 2020-10-19
Command refactoring

cyruskarsan:
refactored finalize and initialize commands
@vringar Please review. 

Are `DumpProfCommand` , `RunCustomFunctionCommand`, and `ShutdownCommand` the only commands left to refactor? If so, these didn't appear in `browser_commands.py` so would I just add the execute function to each of these classes in `browser_commands.py`?


vringar:
I'll review this on Monday as I'm off on vacation for the rest of the week.
As for where to place the last few commands:
- RunCustomFunction can be removed as users should now be able to just implement their own commands
- The others I'd suggest putting into a new file called `internal_commands.py`, I'd also like to see the initialize and finalize command there

PR #775, opened 2020-10-31
Remove support for random_attributes, ua_strings, screen_res #745

### LINKED ISSUE #745: Remove support for random_attributes and browser_settings
These are a vestige of an unmaintained attempt to allow users to randomize some of the browser's properties. It's incomplete (#17), unsupported (#620), and untested.

Here are some code pointers:

https://github.com/mozilla/OpenWPM/blob/e989ce5f74c1737185c89a33bde4015b55df066a/automation/Commands/profile_commands.py#L14-L35

https://github.com/mozilla/OpenWPM/blob/master/automation/DeployBrowsers/user_agent_strings.txt

https://github.com/mozilla/OpenWPM/blob/master/automation/DeployBrowsers/screen_resolutions.txt

https://github.com/mozilla/OpenWPM/blob/e989ce5f74c1737185c89a33bde4015b55df066a/automation/DeployBrowsers/deploy_firefox.py#L80-L97

https://github.com/mozilla/OpenWPM/blob/e989ce5f74c1737185c89a33bde4015b55df066a/automation/DeployBrowsers/deploy_firefox.py#L105-L110

Any future anti-bot circumvention work should be tracked in #448

ankushduacodes:
Running tests to fix #745

### DIFF
--- a/automation/BrowserManager.py
+++ b/automation/BrowserManager.py
@@ -110,7 +110,7 @@ class Browser:
             )
             # make sure browser loads crashed profile
             self.browser_params['recovery_tar'] = tempdir
-            # don't re-randomize attributes
+            
             crash_recovery = True
         else:
         """
--- a/automation/BrowserManager.py
+++ b/automation/BrowserManager.py
@@ -75,7 +75,6 @@ class Browser:
         self.restart_required = False
 
         self.current_timeout: Optional[int] = None  # timeout of the current command
-        self.browser_settings = None
         self.browser_manager = None  # process that controls browser
 
         self.logger = logging.getLogger("openwpm")
@@ -105,7 +104,6 @@ class Browser:
                 self.browser_params,
                 tempdir,
                 close_webdriver=False,
-                browser_settings=self.browser_settings
             )
             # make sure browser loads crashed profile
             self.browser_params['recovery_tar'] = tempdir
@@ -164,9 +162,7 @@ class Browser:
                 # 4. Browser launch attempted
                 check_queue(launch_status)
                 # 5. Browser launched
-                (self.geckodriver_pid, self.browser_settings) = check_queue(
-                    launch_status
-                )
+                (self.geckodriver_pid) = check_queue(launch_status)
 
                 (driver_profile_path, ready) = check_queue(launch_status)
                 if ready != "READY":
@@ -415,7 +411,6 @@ class Browser:
                 self.browser_params,
                 self.browser_params['profile_archive_dir'],
                 close_webdriver=False,
-                browser_settings=self.browser_settings,
                 compress=True
             )
         """
@@ -438,7 +433,7 @@ def BrowserManager(
     logger = logging.getLogger("openwpm")
     try:
         # Start Xvfb (if necessary), webdriver, and browser
-        driver, prof_folder, browser_settings = deploy_browser.deploy_browser(
+        driver, prof_folder = deploy_browser.deploy_browser(
             status_queue, browser_params, manager_params, crash_recovery
         )
         if prof_folder[-1] != "/":
@@ -520,7 +515,6 @@ def BrowserManager(
                 command_executor.execute_command(
                     command,
                     driver,
-                    browser_settings,
                     browser_params,
                     manager_params,
                     extension_socket,
--- a/automation/BrowserManager.py
+++ b/automation/BrowserManager.py
@@ -162,7 +162,7 @@ class Browser:
                 # 4. Browser launch attempted
                 check_queue(launch_status)
                 # 5. Browser launched
-                (self.geckodriver_pid) = check_queue(launch_status)
+                self.geckodriver_pid = check_queue(launch_status)
 
                 (driver_profile_path, ready) = check_queue(launch_status)
                 if ready != "READY":
--- a/automation/BrowserManager.py
+++ b/automation/BrowserManager.py
@@ -111,7 +111,6 @@ class Browser:
             # make sure browser loads crashed profile
             self.browser_params['recovery_tar'] = tempdir
             # don't re-randomize attributes
-            self.browser_params['random_attributes'] = False
             crash_recovery = True
         else:
         """
--- a/automation/BrowserManager.py
+++ b/automation/BrowserManager.py
@@ -75,7 +75,6 @@ class Browser:
         self.restart_required = False
 
         self.current_timeout: Optional[int] = None  # timeout of the current command
-        # dict of additional browser profile settings (e.g. screen_res)
         self.browser_settings = None
         self.browser_manager = None  # process that controls browser
 
--- a/automation/DeployBrowsers/deploy_firefox.py
+++ b/automation/DeployBrowsers/deploy_firefox.py
@@ -91,7 +91,7 @@ def deploy_firefox(status_queue, browser_params, manager_params, crash_recovery)
         fo.add_argument("--height={}".format(DEFAULT_SCREEN_RES[1]))
     if display_mode == "xvfb":
         try:
-            display = Display(visible=0)
+            display = Display(visible=0, size=DEFAULT_SCREEN_RES)
             display.start()
             display_pid, display_port = display.pid, display.cmd_param[-1][1:]
         except EasyProcessError:
@@ -192,6 +192,8 @@ def deploy_firefox(status_queue, browser_params, manager_params, crash_recovery)
             % browser_params["browser_id"]
         )
 
+    # set window size
+    driver.set_window_size(*DEFAULT_SCREEN_RES)
 
     # Get browser process pid
     if hasattr(driver, "service") and hasattr(driver.service, "process"):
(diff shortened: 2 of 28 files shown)


automation/BrowserManager.py
```diff
@@ -111,7 +111,6 @@ def launch_browser_manager(self):
             # make sure browser loads crashed profile
             self.browser_params['recovery_tar'] = tempdir
             # don't re-randomize attributes
```
vringar, line 113:
```suggestion
            
```
Please also delete the associated comment

automation/default_browser_params.json
```diff
@@ -8,6 +8,7 @@
     "save_content": false,
     "callstack_instrument": false,
     "dns_instrument": false,
+    "bot_mitigation": false,
```
vringar, line 11:
```suggestion
    
```
The key already exists in line 18. Consider either moving it up here or removing this entry.

vringar: Also please remove the "random_attributes" in like 22.

vringar (changes requested):
This PR is looking very good. But there are some little changes to be made.

vringar:
You should also remove 
https://github.com/mozilla/OpenWPM/blob/e989ce5f74c1737185c89a33bde4015b55df066a/automation/Commands/profile_commands.py#L13-L35
And the places where this code is called further down in the file.

automation/default_browser_params.json
ankushduacodes: Could please explain what do you mean by `remove the "random_attributes" in like 22.`

ankushduacodes:
> You should also remove
> https://github.com/mozilla/OpenWPM/blob/e989ce5f74c1737185c89a33bde4015b55df066a/automation/Commands/profile_commands.py#L13-L35
> 
> 
> And the places where this code is called further down in the file.

Do you mean remove the comments? or the whole code snippet and its calls down the line?
if I need to remove the whole code snippet then, do I need to remove only these two functions and its calls? or is there other functions that I need to remove?

vringar:
I mean the whole snippet and it's calls down the line as well as any references to browser_settings. browser_settings were only used to store ua_strings and screen_res and since we want to remove both, we don't need them anymore.

automation/BrowserManager.py
```diff
@@ -75,7 +75,6 @@ def __init__(self, manager_params, browser_params) -> None:
         self.restart_required = False
 
         self.current_timeout: Optional[int] = None  # timeout of the current command
-        # dict of additional browser profile settings (e.g. screen_res)
         self.browser_settings = None
```
vringar, line 78: Meaning we can remove this line as well.

ankushduacodes:
> I mean the whole snippet and it's calls down the line as well as any references to browser_settings. browser_settings were only used to store ua_strings and screen_res and since we want to remove both, we don't need them anymore.
> I don't think you need to remove any other functions but I might have missed something.

What do you suggest I do to remove browser_settings as in this case https://github.com/mozilla/OpenWPM/blob/e989ce5f74c1737185c89a33bde4015b55df066a/automation/Commands/profile_commands.py#L143-L190
browser_settings are returned here https://github.com/mozilla/OpenWPM/blob/e989ce5f74c1737185c89a33bde4015b55df066a/automation/DeployBrowsers/deploy_firefox.py#L66-L82

vringar:
> What do you suggest I do to remove browser_settings as in this case
> https://github.com/mozilla/OpenWPM/blob/e989ce5f74c1737185c89a33bde4015b55df066a/automation/Commands/profile_commands.py#L143-L190
> 
Here I'd suggest deleting Lines 182 and 190
> 
> https://github.com/mozilla/OpenWPM/blob/e989ce5f74c1737185c89a33bde4015b55df066a/automation/DeployBrowsers/deploy_firefox.py#L66-L71
> 
> And here
> 
> https://github.com/mozilla/OpenWPM/blob/e989ce5f74c1737185c89a33bde4015b55df066a/automation/DeployBrowsers/deploy_firefox.py#L77-L82

Delete the assignment and keep the call. 
So change it to
```python
load_profile( 
     browser_profile_path, 
     manager_params, 
     browser_params, 
     browser_params["recovery_tar"], 
 ) 
```
And then go down in those functions to remove everything that mentions `profile_settings`

ankushduacodes:
https://github.com/mozilla/OpenWPM/blob/e989ce5f74c1737185c89a33bde4015b55df066a/automation/Commands/profile_commands.py#L59-L140
this code snippet  is unreachable due to the return statement above.
Do you want to also remove this code snippet or do you want to keep it?

ankushduacodes:
@vringar Can you please briefly explain what exactly load_profile function is doing? and is 

As it is wrapped around with try-except block, we can only remove line 182 and line 190.

Is that try-except block needed after we delete the load_browser_settings function? at what point exactly are we expecting where an error may be needed to be handled?

To my understanding we are using load_profile function to load browser_settings and return the browser_settings and assigning it to profile_settings which we want to get rid of. and we are also extracting the tar.gz file into browser_profile_folder directory.

what exactly does full_profile.tar.gz or profile.tar.gz includes, I mean what kind of information?

vringar:
So this is off the top of my head and might not be completely accurate, but as far as I remember:
> what exactly does full_profile.tar.gz or profile.tar.gz includes, I mean what kind of information?

These files contain a tarred Firefox Profile. This is where your cookies, history and settings are stored in one place.

We use Selenium to control Firefox. Selenium creates an empty profile folder in a temporary directory when we tell it to start a new Firefox instance.
But we want Firefox to use the profile we have had created somewhere else, so we ask it where that empty profile is and put the contents of the tar in there. This is what `load_profile` does.

Somehow we decided that we also needed to save some extra information for each browser which wasn't in the `browser_params` and not in the profile so we created `browser_settings` which contained things like screen size and just saved that right next to the tar.
But now we have changed our minds on this and have decided that the benefits of this aren't worth all of the complexity. So now we try to purge any usage of `browser_settings` to allow us to simplify `load_profile` even further in the future.

automation/DeployBrowsers/deploy_firefox.py
```diff
@@ -237,6 +202,6 @@ def deploy_firefox(status_queue, browser_params, manager_params, crash_recovery)
     else:
         raise RuntimeError("Unable to identify Firefox process ID.")
 
-    status_queue.put(("STATUS", "Browser Launched", (int(pid), profile_settings)))
+    status_queue.put(("STATUS", "Browser Launched", (int(pid))))
```
vringar, line 205:
```suggestion
    status_queue.put(("STATUS", "Browser Launched", int(pid)))
```
As you only have one element here and no comma no tuple gets created.

automation/BrowserManager.py
```diff
@@ -166,9 +162,7 @@ def check_queue(launch_status):
                 # 4. Browser launch attempted
                 check_queue(launch_status)
                 # 5. Browser launched
-                (self.geckodriver_pid, self.browser_settings) = check_queue(
-                    launch_status
-                )
+                (self.geckodriver_pid) = check_queue(launch_status)
```
vringar, line 165:
```suggestion
                self.geckodriver_pid = check_queue(launch_status)
```

vringar (approved):
Two small nitpicks but otherwise this looks good to me!

vringar:
> https://github.com/mozilla/OpenWPM/blob/e989ce5f74c1737185c89a33bde4015b55df066a/automation/Commands/profile_commands.py#L59-L140
> 
> this code snippet is unreachable due to the return statement above.
> Do you want to also remove this code snippet or do you want to keep it?

Yes please keep this code around. We eventually want to re-enable profile dumping but haven't gotten around to it yet.

ankushduacodes:
> So this is off the top of my head and might not be completely accurate, but as far as I remember:
> 
> > what exactly does full_profile.tar.gz or profile.tar.gz includes, I mean what kind of information?
> 
> These files contain a tarred Firefox Profile. This is where your cookies, history and settings are stored in one place.
> 
> We use Selenium to control Firefox. Selenium creates an empty profile folder in a temporary directory when we tell it to start a new Firefox instance.
> But we want Firefox to use the profile we have had created somewhere else, so we ask it where that empty profile is and put the contents of the tar in there. This is what `load_profile` does.
> 
> Somehow we decided that we also needed to save some extra information for each browser which wasn't in the `browser_params` and not in the profile so we created `browser_settings` which contained things like screen size and just saved that right next to the tar.
> But now we have changed our minds on this and have decided that the benefits of this aren't worth all of the complexity. So now we try to purge any usage of `browser_settings` to allow us to simplify `load_profile` even further in the future.
> 
> If any of this makes sense to you and you feel like you can shorten this a bit, please write this as a docstring on `load_profile`

Thank you for explaining, I think now have a big picture of what goes on when we save cookies in the browser.
Does this apply when a real user is using firefox, Is that how the cookies are stored in the system?

vringar:
> Thank you for explaining, I think now have a big picture of what goes on when we save cookies in the browser.
> Does this apply when a real user is using firefox, Is that how the cookies are stored in the system?

Cookies are always stored in your profile. However usually your Profile just stays in one place on the system and doesn't get tarred or copied somewhere else.

vringar:
Do you feel like there is anything else you'd want to do as part of this change?
Otherwise I'd be willing to merge it rn.

ankushduacodes:
> Do you feel like there is anything else you'd want to do as part of this change?
> Otherwise I'd be willing to merge it rn.

I think I have made all the necessary changes. I think its good to go!...

vringar:
Then thank you very much for following through on this!

PR #791, opened 2020-11-12
Add performace timer

### LINKED ISSUE #679: Add column to crawl_history table to capture how long command took
I'd like to get some kind of idea of how long different commands are taking. 
My motivation is to be able to debug "performance" when instrumenting lots of APIs. I'm thinking about a very crude python timer added around here: 
https://github.com/mozilla/OpenWPM/blob/a811af1c9c1f8efa4b9d2a008c6cbb3e5be6d1e8/automation/TaskManager.py#L469
Obviously this number will need to be understood carefully, but I still think it'll be useful.

The other place I think it could be useful in the future is in quantifying the amount of time we spend in "finalize". Now that we can introspect this part of the crawl, it feels like an area we may want to tune in the future in order to optimize cost of crawling.

For this issue to be closed the following things need to be done:
1. Add a  `duration` column of type `int` to the `crawl_history` table both for the [SQL](https://github.com/mozilla/OpenWPM/blob/a811af1c9c1f8efa4b9d2a008c6cbb3e5be6d1e8/automation/DataAggregator/schema.sql#L41) as well as for the [Parquet](https://github.com/mozilla/OpenWPM/blob/a811af1c9c1f8efa4b9d2a008c6cbb3e5be6d1e8/automation/DataAggregator/parquet_schema.py#L26) schema
2. Add 
    ```python
    t1 = time.time()
    ``` 
    before
    https://github.com/mozilla/OpenWPM/blob/a811af1c9c1f8efa4b9d2a008c6cbb3e5be6d1e8/automation/TaskManager.py#L469
3. Add another line here
https://github.com/mozilla/OpenWPM/blob/a811af1c9c1f8efa4b9d2a008c6cbb3e5be6d1e8/automation/TaskManager.py#L521-L537
saying
```python
"duration": time.time()-t1
```

ankushduacodes:
fixes #679


ankushduacodes:
@vringar I have question about core functionality of OpenWPM, if we go to a website lets say `www.youtube.com`, I see that its logging all the files that are being loaded, and all the requests it making and responses its giving.
But I was wondering about what happens  to redirects, Does OpenWPM goes to a redirect if a website wants to send a user to that redirect? or does it block that redirect.
And if it does go to that redirect, then how does it keep track of orginal website that is refering us to that redirect(Using multiple processes?)

vringar (changes requested):
Could you please also add this field to the [Schema-Documentation](https://github.com/mozilla/OpenWPM/blob/master/docs/Schema-Documentation.md)
The rest looks good :)

vringar:
> @vringar I have question about core functionality of OpenWPM, if we go to a website lets say `www.youtube.com`, I see that its logging all the files that are being loaded, and all the requests it making and responses its giving.
> But I was wondering about what happens to redirects, Does OpenWPM goes to a redirect if a website wants to send a user to that redirect? or does it block that redirect.
> And if it does go to that redirect, then how does it keep track of orginal website that is refering us to that redirect(Using multiple processes?)
Each CommandSequence needs to specify a `site_url` it wants to visit. Stored here:
https://github.com/mozilla/OpenWPM/blob/65337e0c19dc857f63bbca1b3aa3773a2c49fd42/automation/DataAggregator/schema.sql#L23-L28
In your example this would be `youtube.com`. Once a CommandSequence gets executed it gets a visit_id and a browser_id. All data collected from the execution of that CommandSequence get stored with the same visit_id and browser_id. This includes redirects and anything else the site might do.

automation/TaskManager.py
```diff
@@ -536,6 +540,7 @@ def _issue_command(
                         "command_status": command_status,
                         "error": error_text,
                         "traceback": tb,
+                        "duration": time.time_ns() - t1,
```
vringar, line 543: Reading this again I noticed, that we discussed saving the time in millisecond, so this needs to be
```suggestion
                        "duration": int((time.time_ns() - t1)/1000000),
```

docs/Schema-Documentation.md
```diff
@@ -38,7 +38,7 @@ update the description of the field here.
 | command_status | string |          |             |
 | error          | string |          |             |
 | traceback      | string |          |
-
+| duration       | int64  |          | A timer that logs how long a command took|
```
vringar, line 41: Please document that the duration is in milliseconds.

vringar (changes requested):

ankushduacodes:
> In your example this would be `youtube.com`. Once a CommandSequence gets executed it gets a visit_id and a browser_id. All data collected from the execution of that CommandSequence get stored with the same visit_id and browser_id. This includes redirects and anything else the site might do.

But does it actually visit that redirect or does it just logs that this website requested that redirects?

vringar:
> But does it actually visit that redirect or does it just logs that this website requested that redirects?

It follows the redirects :) Overall OpenWPM tries to behave as much as a "normal" browser as possible.
To see that for yourself, change demo.py to visit `http://github.com/citp/openwpm` and then have a look into the database what gets logged there.

ankushduacodes:
@vringar Is there a better way to combat these conflicts? I know why these conflicts are here, It's because of the package rename.
Do I just have to abandon this pull request and start a new one?

vringar:
What I did in my PR was to first move the `automation` folder to `openwpm`, commit that change and then merge the master branch into the PR branch.
Seeing as your changes are quite small I think this should merge without any problems.

ankushduacodes:
> What I did in my PR was to first move the `automation` folder to `openwpm`, commit that change and then merge the master branch into the PR branch.
> Seeing as your changes are quite small I think this should merge without any problems.

Just so I am understanding it correctly, you are suggesting that I rename my automation folder to openwpm in `add_performance timer` branch and then commit?

vringar:
Yes, that's my suggestion. This way git can match up the corresponding files more easily which should prevent any merge conflicts.

PR #794, opened 2020-11-16
Updated all JS/TS dependencies

vringar:
I just went through and updated everything to the most up to date version of each package.
I should note that tslint [is being deprecated in favour of eslint](https://github.com/palantir/tslint/issues/4534) with TS configs, so we should eventually remove that as well.
But for now this seems good enough :)


openwpm/Extension/webext-instrumentation/src/background/http-instrument.ts
```diff
@@ -43,7 +43,7 @@ const allTypes: ResourceType[] = [
       "sub_frame",
       "web_manifest",
       "websocket",
-      "xbl",
+      //"xbl",
```
vringar, line 46:
```suggestion
```

Issue #800, opened 2020-11-18
Unnecessary List Unpacking


ankushduacodes:
In the code snippet: https://github.com/mozilla/OpenWPM/blob/512018f852a2b554b5b08a72936f835adf6b7110/test/test_webdriver_utils.py#L29-L33 
List unpacking is very unnecessary here, I am not sure why `_` is used here, and what the use of it is.

Explanation:
Now, when I tried to debug what the `SELECT` query is returning is a list of two tuples, Now, to my understanding  one is the tuples is for `<class 'openwpm.Commands.Types.InitializeCommand'>` for initialization command and the second one is for `<class 'openwpm.Commands.Types.GetCommand'>`. 
To my understanding, for this test, we are only interested in `command_status` and `error` for `<class 'openwpm.Commands.Types.GetCommand'>`.

Potential fixes:
1. Remove the underscore and assign the result of SELECT command into get_command and in the assertion we can use appropriate indices. Something like:
`assert get_command[1][0] == "neterror"`
`assert get_command[1][1] == "dnsNotFound"`
2. Remove the underscore and assign indicised result to `get_command`. Something like:
```python
get_command = db_utils.query_db(
            manager_params["db"],
            "SELECT command_status, error " "FROM crawl_history ",
            as_tuple=True,
        )[1]
```

vringar:
I'd prefer option 2 over 1. Thanks for pointing this out.
Please also remove the double quotes in the middle of the query string.

ankushduacodes:
@vringar We could also potentially use `WHERE` clause here as well as pointed out by you [here](https://github.com/mozilla/OpenWPM/pull/799#discussion_r525895829)

LordReigns:
I would like to work on this issue. @ankushduacodes .

ankushduacodes:
@LordReigns Have to set up the development environment? If not please do so using [this guide](https://github.com/mozilla/OpenWPM/blob/master/README.md#installation), You should also set up pre-commit configurations using [this guide](https://github.com/mozilla/OpenWPM/blob/master/CONTRIBUTING.md#setting-up-a-dev-enviroment)

[...]

PR #804, opened 2020-11-19
Updated stored commands in test_webdriver_utils.py

LordReigns:
Fixes updated change issue in #800


vringar:
Thanks for the quick fix!

vringar (approved):

PR #807, opened 2020-11-23
Refactoring browser and manager params into dataclasses

ankushduacodes:
This pull request/draft aims to fix #495 

### TODOS
- [x] Refactor use of params dicts to use BrowserParams and ManagerPramans everywhere
- [x] Write Tests
- [x] Write tutorial/Documentation for the use of custom browser and manager params
- [x] Write documentation for dataclasses use
- [x] Resolve all the #TODOS that I added in this pull request

PR #817, opened 2020-12-03
Fixing import in manual_test.py broken after #807

MollieBakal:
Minor fix for import no longer found. Quick question--given the recent work converting module names to abide by snake_case, should BrowserParams be browser_params?

### DIFF
--- a/test/manual_test.py
+++ b/test/manual_test.py
@@ -9,8 +9,8 @@ from selenium import webdriver
 from selenium.webdriver.firefox.firefox_binary import FirefoxBinary
 
 from openwpm import js_instrumentation as jsi
-from openwpm.deploy_browsers import configure_firefox
 from openwpm.config import BrowserParams
+from openwpm.deploy_browsers import configure_firefox
 from openwpm.utilities.platform_utils import get_firefox_binary_path
 
 from .conftest import create_xpi
@@ -108,7 +108,7 @@ def start_webdriver(
             print("Cleanup before shutdown...")
             server.shutdown()
             thread.join()
-            print("...sever shutdown")
+            print("...server shutdown")
             driver.quit()
             print("...webdriver closed")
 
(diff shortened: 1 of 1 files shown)


MollieBakal:
Just noticed more changes which need to be made in the area--sorry. Also realized I forgot to add in pre-commit, which has since been rectified.

vringar:
Thank you for this PR!
As BrowserParams is a class and not a module I think CamelCase is the appropriate formatting.

MollieBakal:
You're right--I just looked at PEP8, which I hadn't before. Interesting :).

test/manual_test.py
```diff
@@ -127,12 +127,12 @@ def cleanup_server():
...
 
-        browser_params = load_default_params()[1][0]
+        browser_params = BrowserParams()
         if browser_params_file is not None:
             with open(browser_params_file, "r") as f:
-                browser_params.update(json.loads(f.read()))
+                browser_params.from_json(f.read())
         js_request = browser_params.js_instrument_settings
-        js_request_as_string = jsi.convert_browser_params_to_js_string(js_request)
+        js_request_as_string = json.dumps(js_request)
```
vringar, line 135: Please revert the change in this line. The js_request isn't a valid json unfortunately, so we need to call out to that method.

vringar (changes requested):

test/manual_test.py
MollieBakal: Sorry about that--but I don't see the function `convert_browser_params_to_js_string`, and that was throwing an error at me. Is the correct function `clean_js_instrumentation_settings`?

vringar: Yes, that would be it. Sorry for letting this file get so out of date.

MollieBakal: Not a problem at all. Just committed with new name.

vringar (approved):

PR #831, opened 2020-12-14
Update pyvirtualdisplay

### LINKED ISSUE #694: Update pyvirtualdisplay to >1.0.0 version
In #682 we pinned `pyvirtualdisplay` to 0.2.5 due to an incompatible interface change. See the excerpt from this [test run](https://travis-ci.org/github/mozilla/OpenWPM/jobs/699798694#L2176-L2183):

```

  File "/home/travis/build/mozilla/OpenWPM/automation/BrowserManager.py", line 450, in BrowserManager

    driver, prof_folder, browser_settings = deploy_browser.deploy_browser(

  File "/home/travis/build/mozilla/OpenWPM/automation/DeployBrowsers/deploy_browser.py", line 13, in deploy_browser

    return deploy_firefox.deploy_firefox(status_queue, browser_params,

  File "/home/travis/build/mozilla/OpenWPM/automation/DeployBrowsers/deploy_firefox.py", line 126, in deploy_firefox

    display_pid, display_port = display.pid, display.cmd_param[-1][1:]

AttributeError: 'Display' object has no attribute 'cmd_param'
```

We use `cmd_param` to retrieve the "display port", which is then used to identify a temporary lockfile placed in the users tmp folder. See:

```python
        if self.display_port is not None:  # xvfb diplay lock
            lockfile = "/tmp/.X%s-lock" % self.display_port
            try:
                os.remove(lockfile)
            except OSError:
                self.logger.debug("BROWSER %i: Screen lockfile (%s) already "
                                  "removed" % (self.crawl_id, lockfile))
                pass
```

This was initially added because long-running crawls would fill the tmp directory with too many lock files and would eventually lead XVFB to fail because it would be assigned a port that already had a lock file.

This can probably be fixed by actually shutting down xvfb properly in https://github.com/mozilla/OpenWPM/blob/43922bc18cc458eab3aaea618c1f4bb4b14ccf8f/automation/BrowserManager.py#L503-L512 and not just killing the process like we do right now.

I think the new >1.0.0 interface still exposes the same info via the `display` property. See the lockfile [creation code](https://github.com/ponty/PyVirtualDisplay/blob/8fd1db158518fb67b461a5acdcf3934c0679a503/pyvirtualdisplay/abstractdisplay.py#L42-L67). But as Sarah points out in https://github.com/mozilla/OpenWPM/pull/682#issuecomment-647680950 this might not even be necessary, since a proper shutdown should clear the lockfile and loads of improper shutdowns point to another issue. We should verify that's the case (i.e., a proper shutdown removes the lockfile).

MollieBakal:
Closes #694


MollieBakal:
Linking the issue #694 -- also, please note the first 5 commits here were already merged

environment.yaml
```diff
@@ -22,7 +22,7 @@ dependencies:
 - pytest-cov=2.10.1
 - pytest=6.1.2
 - python=3.8.6
-- pyvirtualdisplay=0.2.5
+- pyvirtualdisplay=1.3.2
```
vringar, line 25: Please don't edit the enviroment.yaml directly.
Instead have a look at [this](https://github.com/mozilla/OpenWPM/blob/master/CONTRIBUTING.md#managing-requirements)

openwpm/deploy_browsers/deploy_firefox.py
```diff
@@ -157,4 +158,4 @@ def deploy_firefox(status_queue, browser_params, manager_params, crash_recovery)
 
     status_queue.put(("STATUS", "Browser Launched", int(pid)))
 
-    return driver, driver.capabilities["moz:profile"]
+    return driver, driver.capabilities["moz:profile"], display
```
vringar, line 161: Could you please give this function a type annotation.

test/test_mp_logger.py
```diff
@@ -176,7 +176,9 @@ def test_child_process_logging(self, tmpdir):
         child_process = Process(target=child_proc_logging_exception())
         child_process.daemon = True
         child_process.start()
-        openwpm_logger.close()
+        time.sleep(2)
         child_process.join()
+        print("Child processes joined...")
```
vringar, line 181: This seems to fix this test. Could you please remove the `skipif` so it runs in CI again.

test/test_xvfb_browser.py
```diff
@@ -0,0 +1,26 @@
...
+    def test_display_shutdown(self):
+        manager_params, browser_params = self.get_config()
+        TEST_SITE = BASE_TEST_URL + "/test_pages/simple_a.html"
+        manager = TaskManager(manager_params, browser_params)
+
+        sequence = CommandSequence(TEST_SITE)
+        sequence.get()
+        sequence.save_screenshot()
+        manager.execute_command_sequence(sequence)
+        manager.close()
```
vringar, line 26: I don't think that this test tests anything that isn't tested by [`test_simple_commands.py`](https://github.com/mozilla/OpenWPM/blob/master/test/test_simple_commands.py)

vringar (changes requested):
Thank you for creating this PR! I've left a couple of minor things and will comment some more on the issue.

vringar:
When writing [this comment](https://github.com/mozilla/OpenWPM/issues/694#issuecomment-744448099)  I noticed one thing.
If we could guarantee that the Display object always gets  `stop`ed then we could just remove all of the other code.
And since the entire function is already wrapped in a try/except-block we could add a [finally](https://docs.python.org/3/reference/compound_stmts.html#the-try-statement) that contains the changes you made in the shutdown block.
This way, no matter how we exit the function we'd `stop` the Display properly.

MollieBakal:
These commits aren't done--I'm having some trouble with the environment

environment.yaml
MollieBakal: Sorry--done.

openwpm/deploy_browsers/deploy_firefox.py
MollieBakal: I think I annotated this properly--unsure, though

test/test_mp_logger.py
MollieBakal: I don't think this was me? But I removed the skipif

test/test_xvfb_browser.py
MollieBakal: As per request later, added different command. Had not seen that displaymode=xvfb was tested, sorry.

MollieBakal:
By rebasing based on CommandRefactoring, it seems to have created a number of unintended commits--I don't know if that was unintentional, or if I should file another pull request to that branch instead of master.

openwpm/deploy_browsers/deploy_firefox.py
```diff
@@ -16,7 +19,12 @@
 logger = logging.getLogger("openwpm")
 
 
-def deploy_firefox(status_queue, browser_params, manager_params, crash_recovery):
+def deploy_firefox(
+    status_queue: Optional[Queue],
```
vringar, line 23: The `status_queue` is required, so we don't need optional here.
Thanks for annotating this!
```suggestion
    status_queue: Queue,
```

test/test_mp_logger.py
vringar: Oh then that must have been me ^^ Thanks anyways

test/test_xvfb_browser.py
```diff
@@ -0,0 +1,32 @@
...
+    def test_display_shutdown(self):
+        manager_params, browser_params = self.get_config()
+        TEST_SITE = BASE_TEST_URL + "/test_pages/simple_a.html"
+        manager = TaskManager(manager_params, browser_params)
+
+        sequence = CommandSequence(TEST_SITE)
+        sequence.get()
+        sequence.append_command(ExceptionCommand)
+        manager.execute_command_sequence(sequence)
+        manager.close()
```
vringar, line 32: Could you now assert that the lock file has been removed?
So we can verify that the Display got stopped properly.

vringar (changes requested):

vringar:
> By rebasing based on CommandRefactoring, it seems to have created a number of unintended commits--I don't know if that was unintentional, or if I should file another pull request to that branch instead of master.

I have changed this PR to target the CommandRefactoring branch. Unfortunately this is always a bit messy with GitHub, but I'll ignore all files that I know you didn't touch.

vringar:
> These commits aren't done--I'm having some trouble with the environment

If you elaborate on what these troubled are, I'd be happy to help you.

MollieBakal:
Thanks, but I think it was just an issue with cached files? I reran install.sh and that seems to have cleared it up.

test/test_xvfb_browser.py
MollieBakal: Yep--check out the last commit.

test/test_xvfb_browser.py
```diff
@@ -0,0 +1,38 @@
...
+from openwpm.commands.types import BaseCommand
+from openwpm.task_manager import TaskManager
+
+from .openwpmtest import OpenWPMTest
+from .utilities import BASE_TEST_URL
+
+
+class ExceptionCommand(BaseCommand):
+    def __init__():
+        None
```
vringar, line 15:
```suggestion
```
Why declare the `__init__` if you don't need it?

vringar:
Thanks for fixing this issue :) 
I'll wait for @englehardt to merge the CommandRefactoring into master before I merge this.
Do you want remove the `__init__` in the meantime?

test/test_xvfb_browser.py
MollieBakal: It was a placeholder--sorry.

MollieBakal:
Yep! Thank you

vringar (approved):

PR #832, opened 2020-12-15
Removing use of deprecated methods

ankushduacodes:
Removing use of deprecated method `webdriver.switch_to_window(window)` in favor of `webdriver.switch_to.window(window)` to avoid unnecessary warnings
`webdriver.switch_to_window(window)` calls `webdriver.switch_to.window(window)` internally.
Also removed some more methods like above that were deprecated by selenium.

`webdriver.switch_to.<method>` also seems to be the way selenium will be going moving forward as they are added few more new exclusive methods to `SwitchTo` class to be released with selenium4

### DIFF
--- a/openwpm/commands/browser_commands.py
+++ b/openwpm/commands/browser_commands.py
@@ -78,9 +78,9 @@ def close_other_windows(webdriver):
     if len(windows) > 1:
         for window in windows:
             if window != main_handle:
-                webdriver.switch_to_window(window)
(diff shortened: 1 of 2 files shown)


vringar:
Thank you very much for cleaning this up. But could you please make these changes to the `CommandRefactoring` branch as any changes to the current commands will introduce merge conflicts in #750 which will get merged soon.

Issue #858, opened 2021-02-15
Localstack doesn't work with python 3.9


vringar:
This issue just appeared for me a couple of days ago.
You might need to switch to docker based workflow/an older version of conda for the time being.
https://github.com/conda/conda/issues/9653

boolean5:
I Just ran `repin.sh` with conda version 4.8.3 and I'm getting some errors about specifications being incompatible with each other and with the python installation in my environment. Here's the output: [repin_output.txt](https://github.com/mozilla/OpenWPM/files/5988304/repin_output.txt)

vringar:
Paging @birdsarah as the conda expert in residence:
As you can see some of the conflicts are complete bogus:
```
Package pathlib2 conflicts for:
ipython -> pathlib2
pytest -> pathlib2[version='>=2.2.0']
pytest-cov -> pytest[version='>=4.6'] -> pathlib2[version='>=2.2.0']
pre-commit -> importlib_metadata -> pathlib2
```
The issue I have referenced conda v4.8.1 so we might need to go back another version, but that doesn't seem like a good way forward.

Any ideas? Do we say that #753 removes the need for solving from a partially pinned environment and so we just merge the evironment.yamls and then solve from there?

vringar:
Looking through conda/conda#9653 again someone mentions `--update-deps` as a workaround however neither `conda env create` nor `conda env update` support this flag

birdsarah:
The latest version of conda is 4.9.2. If you're not on that, start there.

vringar:
I have the same issue on 4.9.2

vringar:
It seems to be broken starting in 4.8.1 and has been broken ever since.

birdsarah:
OK. I've run this locally. I don't see the problem. I'm not getting "frozen." When it tries to add the dev dependencies it hits conflicts. Here's the full output:

```
UnsatisfiableError: The following specifications were found to be incompatible with each other:

Output in format: Requested package -> Available versions

Package toml conflicts for:
black -> toml[version='>=0.10.1|>=0.9.4']
pytest-cov -> pytest[version='>=4.6'] -> toml
pytest -> toml
pre-commit -> toml

Package pytest conflicts for:
pytest-cov -> pytest[version='>=3.6|>=4.6']
pytest

Package ca-certificates conflicts for:
ipython -> python[version='>=2.7,<2.8.0a0'] -> ca-certificates
pip -> python -> ca-certificates
[...]

PR #860, opened 2021-02-16
Now handling all constraint notations in unpinned enviroment.yamls

vringar:
@birdsarah mentioned in https://github.com/mozilla/OpenWPM/issues/858#issuecomment-779931118 that we don't pin dependencies specified by `<` or `>`.

This PR fixes that inconsistency and guarantees that the last `not_pip` dependency will get pinned if there are no pip dependencies


scripts/prune-environment.py
```diff
@@ -1,3 +1,5 @@
+from typing import Iterable, List
```
birdsarah, line 1: how to spot a PR from @vringar

birdsarah (commented):
This looks fine from a cursory look. If repin works without changing the pinned env file with the new < then I'd say we're good to go.

What this is highlighting for me is how gross this is.

I built it to match the infrastructure we were replacing. I can't remember the logic for wanting a pruned environment file.

We could switch to [this kind of thing](https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#building-identical-conda-environments) which would let us have a similar setup but without this  prune step which just feels like a maintenance headache.

vringar:
I can't remember either why we didn't want full pinning
Paging @englehardt for context

scripts/prune-environment.py
vringar: Sorry :grin: 
But I got https://github.com/mozilla/OpenWPM/blob/master/CONTRIBUTING.md#types-annotations-in-python into the repo I feel like I have a justification. ^^
It also helped me understand the code.

vringar:
> This looks fine from a cursory look. If repin works without changing the pinned env file with the new < then I'd say we're good to go.
> 
> What this is highlighting for me is how gross this is.
> 
> I built it to match the infrastructure we were replacing. I can't remember the logic for wanting a pruned environment file.
> 
> We could switch to [this kind of thing](https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#building-identical-conda-environments) which would let us have a similar setup but without this prune step which just feels like a maintenance headache.

These conda environments seem to be specific to the OS and Processor used. So I'm hesitant to use this.
I think we can just not do the dependency pruning with `conda env export`

scripts/prune-environment.py
```diff
@@ -19,24 +21,50 @@
...
+    (
+        env_unpinned["dependencies"][:-1]
+        if env_unpinned_contains_pip
+        else env_unpinned["dependencies"]
+    )
+    + (
+        env_unpinned_dev["dependencies"][:-1]
+        if env_unpinned_dev_contains_pip
+        else env_unpinned_dev["dependencies"]
+    ),
```
boolean5, line 51: Shouldn't we also check if `env_pinned` contains `pip` in order to guarantee that the last `not_pip` dependency will get pinned if there are no pip dependencies?

scripts/prune-environment.py
```diff
@@ -19,24 +21,50 @@
...
+
+# Checking if there are any pip dependencies
+try:
+    deps_pip_unpinned = env_unpinned["dependencies"][-1]["pip"]
+except:
+    deps_pip_unpinned = []
+try:
+    deps_pip_unpinned_dev = env_unpinned_dev["dependencies"][-1]["pip"]
+except:
+    deps_pip_unpinned_dev = []
```
boolean5, line 63: Just a suggestion, maybe we could reuse `env_unpinned_contains_pip` and `env_unpinned_dev_contains_pip` here instead of checking again in a different way

scripts/prune-environment.py
vringar: Right! Good catch!

birdsarah:
> > This looks fine from a cursory look. If repin works without changing the pinned env file with the new < then I'd say we're good to go.
> 
> I ran `./repin.sh` and nothing changed.
> 
> > We could switch to [this kind of thing](https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#building-identical-conda-environments) which would let us have a similar setup but without this prune step which just feels like a maintenance headache.
> 
> These conda environments seem to be specific to the OS and Processor used. So I'm hesitant to use this.
> I think we can just not do the dependency pruning after running `conda env export`

YES! This is exactly what I ran into before. Thanks for jogging my memory.

scripts/prune-environment.py
vringar: I changed the check to `env_unpinned_contains_pip or env_unpinned_dev_contains_pip` because if neither of the unpinned envs mentions pip dependencies, the resulting env should also only contain conda packages.

englehardt:
> I can't remember either why we didn't want full pinning
> Paging @englehardt for context

I prefer full pinning. IIRC not pruning would lead to issues. Maybe @birdsarah can speak to the motivation for implementing pruning?

vringar:
Seeing as she said:

> I built it to match the infrastructure we were replacing. I can't remember the logic for wanting a pruned environment file.

I don't think she can speak to why the pruning was implemented.
Do we just want to drop the pruning and see if we encounter any issues? Or should I try to find the PR and the associated issue for introducing conda and try to figure out why we were pruning?

englehardt:
> Seeing as she said:
> 
> > I built it to match the infrastructure we were replacing. I can't remember the logic for wanting a pruned environment file.
> 
> I don't think she can speak to why the pruning was implemented.
> Do we just want to drop the pruning and see if we encounter any issues? Or should I try to find the PR and the associated issue for introducing conda and try to figure out why we were pruning?

Check out the May 14th commits here: https://github.com/mozilla/OpenWPM/pull/648

It sounds like it might have been OS differences in dependencies? As in maybe plyvel requires one dependency on Linux and a different one on MacOS.

vringar:
Also dependencies on XServer are probably not satisfiable under MacOS

boolean5 (approved):
This is ready to merge, right?

vringar:
Yes

Issue #680, opened 2020-06-08
Enable dump_profile


vringar:
@MphDevelopment needs stateful crawling for an experiment they wanted to run. This issue outlines the steps that need to be taken to get to a version of stateful crawling that works to satisfy [the given requirements](https://view.matrix.org/room/!pFJihVSEWzcMCcOzSH:mozilla.org/?anchor=$jitB6DiHAZJwyYeu08yE8E51v_VfUejEQApxjNGs6zw&offset=-10&highlight=$jitB6DiHAZJwyYeu08yE8E51v_VfUejEQApxjNGs6zw)
Since we are aiming for a minimum viable product we only care about the happy path and ignore all error handling.
As such the only thing needed is restoring the functionality of [dump_profile](https://github.com/mozilla/OpenWPM/blob/5ffde00ecd5ecaa9105b74935490e5e267596eb7/automation/Commands/profile_commands.py#L38).

The envisioned workflow would look something like:
- Visit the site at 8am with no given profile and afterwards dump out a profile to `8am_profile.tar.gz`
- At 1PM you'd set `browser_params[0]['profile_tar'] = absolute path of 8am_profile.tar.gz`,
visit the site and then dump_profile to `1pm_profile.tar.gz`
- At 8 PM you'd do the same as at 1PM

There is an [old issue](https://github.com/mozilla/OpenWPM/issues/253) discussing this, but quite a lot has changed since then so anything in that issue might be out of date.

MphDevelopment:
Thanks for the information, i'll look into it !

englehardt:
This issue is a bit confusing without context from matrix. @vringar would you mind to add context here? Not sure if this is just for helping MphDevelopment?

vringar:
@MphDevelopment just checking in. Do you need guidance or assistance on anything?

MphDevelopment:
Thanks for checking !
We haven't started on the issue yet, we're still in the research part of the study..
I'll let you know if i need assistance on this ! 

For now i've only tried running OpenWPM with docker, and met some bugs related to npm that seems to magically not be triggered sometimes, i'm trying to understand the issue with it.

Issue #423, opened 2019-07-24
Restore support for profile saving aka stateful crawl support


motin:
#382 disables support for stateful crawling and profile management. This should be restored.

Related to https://github.com/mozilla/OpenWPM/issues/383 and https://github.com/mozilla/openwpm-crawler/issues/13, and there is [a project](https://github.com/mozilla/OpenWPM/projects/2) but no specific issue to track restoration of stateful crawl support, hence this.

englehardt:
[This project](https://github.com/mozilla/OpenWPM/projects/2) collects the issues related to stateful crawling. I suspect the problem will be pretty simple to fix, but will require a lot of testing to discover inevitable corner-case failures. Unfortunately, these corner cases are quite important for stateful crawling, as missing a site leads to an incomplete/inconsistent profile.

The instability was introduced by geckodriver (back during the conversion from [Selenium 2 to 3](https://github.com/mozilla/OpenWPM/pull/152)). Geckodriver manages browser profiles separately from Selenium, and I suspect this upgrade introduced a race condition between our profile management code and geckodriver's. E.g., sometimes the geckodriver process would delete the profile directory before we had a chance to back it up.

Some specific pointers:
* Most of the profile management commands are in [`profile_commands.py`](https://github.com/mozilla/OpenWPM/blob/e96d97c7f2463f1d33ec1da6c328087f70d6f92c/automation/Commands/profile_commands.py). This includes the code to dump and load the current profile.
* When launching Firefox with Selenium we use the [`FirefoxProfile` interface](https://seleniumhq.github.io/selenium/docs/api/py/webdriver_firefox/selenium.webdriver.firefox.firefox_profile.html#module-selenium.webdriver.firefox.firefox_profile). This creates a profile in `/tmp/`. Prior to launching the browser we [copy in a profile](https://github.com/mozilla/OpenWPM/blob/e96d97c7f2463f1d33ec1da6c328087f70d6f92c/automation/DeployBrowsers/deploy_firefox.py#L41-L58) from a specified archive (via the `profile_tar` browser parameter). We also [write](https://github.com/mozilla/OpenWPM/blob/e96d97c7f2463f1d33ec1da6c328087f70d6f92c/automation/DeployBrowsers/deploy_firefox.py#L98-L115) the instrumentation extension configuration file to this directory.
[...]

Issue #384, opened 2019-07-01
Firefox 68 breaks `dump_profile`


englehardt:
For examples see: https://travis-ci.org/mozilla/OpenWPM/jobs/552843910#L2413

Issue #383, opened 2019-07-01
Enable profile tests once stateful crawling support is added back


englehardt:
#382 disables support for stateful crawling and profile management. It also disables a number of tests for these. This issue is a reminder to enable those tests once support for profile saving is restored.

PR #864, opened 2021-03-03
Restore stateful crawling support

boolean5:
This PR restores support for stateful crawling by addressing the core issue that broke it, i.e., the fact that geckodriver would delete the browser profile when closing or crashing before we had a chance to archive it. It does so by using a [custom profile](https://firefox-source-docs.mozilla.org/testing/geckodriver/CrashReports.html) instead of creating the profile via the `FirefoxProfile` Selenium class. The custom profile is set via the arguments of the `Options` class. 

**Note:**
Geckodriver has a bug that makes it write the browser preferences we set, as well as its own default browser preferences, to a `user.js` file in the wrong profile directory when using a custom profile: https://github.com/mozilla/geckodriver/issues/1844. As a temporary workaround until this issue gets fixed, we create the `user.js` file ourselves. In order to do this, we keep a copy of geckodriver's default preferences in `configure_firefox.py`.

This PR also:
* Fixes a bug in `PatchedGeckoDriverService` class that caused geckodriver not to receive the `service_args` passed when starting the browser.
* Removes the [code that intercepts the location of the browser profile from the geckodriver logs](https://github.com/mozilla/OpenWPM/blob/cb95ecc05fb4618e4275faecbc62b84123ebf1b0/openwpm/deploy_browsers/selenium_firefox.py#L71L74), as we no longer need it.
* Updates `manual_test.py`

Closes https://github.com/mozilla/OpenWPM/issues/423, Closes https://github.com/mozilla/OpenWPM/issues/383, Closes https://github.com/mozilla/OpenWPM/issues/161, Closes https://github.com/mozilla/OpenWPM/issues/253, Closes https://github.com/mozilla/OpenWPM/issues/384, Closes https://github.com/mozilla/OpenWPM/issues/680

### DIFF
--- a/openwpm/browser_manager.py
+++ b/openwpm/browser_manager.py
@@ -166,8 +166,8 @@ class Browser:
             # Read success status of browser manager
             launch_status = dict()
             try:
-                # 1. Selenium profile created
-                spawned_profile_path = check_queue(launch_status)
+                # 1. Browser profile created
+                browser_profile_path = check_queue(launch_status)
                 # 2. Profile tar loaded (if necessary)
                 check_queue(launch_status)
                 # 3. Display launched (if necessary)
@@ -177,7 +177,7 @@ class Browser:
                 # 5. Browser launched
                 self.geckodriver_pid = check_queue(launch_status)
 
-                (driver_profile_path, ready) = check_queue(launch_status)
+                ready = check_queue(launch_status)
                 if ready != "READY":
                     self.logger.error(
                         "BROWSER %i: Mismatch of status queue return values, "
@@ -209,7 +209,7 @@ class Browser:
                 )
                 self.close_browser_manager()
                 if "Profile Created" in launch_status:
-                    shutil.rmtree(spawned_profile_path, ignore_errors=True)
+                    shutil.rmtree(browser_profile_path, ignore_errors=True)
 
         # If the browser spawned successfully, we should update the
         # current profile path class variable and clean up the tempdir
@@ -217,9 +217,7 @@ class Browser:
         if success:
             self.logger.debug("BROWSER %i: Browser spawn sucessful!" % self.browser_id)
             previous_profile_path = self.current_profile_path
-            self.current_profile_path = driver_profile_path
-            if driver_profile_path != spawned_profile_path:
-                shutil.rmtree(spawned_profile_path, ignore_errors=True)
+            self.current_profile_path = browser_profile_path
             if previous_profile_path is not None:
                 shutil.rmtree(previous_profile_path, ignore_errors=True)
             if tempdir is not None:
@@ -447,7 +445,7 @@ def BrowserManager(
     display = None
     try:
         # Start Xvfb (if necessary), webdriver, and browser
-        driver, prof_folder, display = deploy_firefox.deploy_firefox(
+        driver, browser_profile_path, display = deploy_firefox.deploy_firefox(
             status_queue, browser_params, manager_params, crash_recovery
         )
 
@@ -456,11 +454,11 @@ def BrowserManager(
         if browser_params.extension_enabled:
             logger.debug(
                 "BROWSER %i: Looking for extension port information "
-                "in %s" % (browser_params.browser_id, prof_folder)
+                "in %s" % (browser_params.browser_id, browser_profile_path)
             )
             elapsed = 0
             port = None
-            ep_filename = prof_folder / "extension_port.txt"
+            ep_filename = browser_profile_path / "extension_port.txt"
             while elapsed < 5:
                 try:
                     with open(ep_filename, "rt") as f:
@@ -487,10 +485,9 @@ def BrowserManager(
 
         logger.debug("BROWSER %i: BrowserManager ready." % browser_params.browser_id)
 
-        # passes the profile folder back to the
-        # TaskManager to signal a successful startup
-        status_queue.put(("STATUS", "Browser Ready", (prof_folder, "READY")))
-        browser_params.profile_path = prof_folder
+        # passes "READY" to the TaskManager to signal a successful startup
+        status_queue.put(("STATUS", "Browser Ready", "READY"))
+        browser_params.profile_path = browser_profile_path
 
         # starts accepting arguments until told to die
         while True:
--- a/openwpm/browser_manager.py
+++ b/openwpm/browser_manager.py
@@ -104,7 +104,7 @@ class Browser:
         # to be a tar of the crashed browser's history
         if self.current_profile_path is not None:
             # tar contents of crashed profile to a temp dir
-            tempdir = tempfile.mkdtemp(prefix="owpm_profile_archive_")
+            tempdir = tempfile.mkdtemp(prefix="openwpm_profile_archive_")
             tar_path = Path(tempdir) / "profile.tar.gz"
 
             self.browser_params.profile_path = self.current_profile_path
--- a/openwpm/browser_manager.py
+++ b/openwpm/browser_manager.py
@@ -217,7 +217,7 @@ class Browser:
         if success:
             self.logger.debug("BROWSER %i: Browser spawn sucessful!" % self.browser_id)
             previous_profile_path = self.current_profile_path
-            self.current_profile_path = Path(driver_profile_path)
+            self.current_profile_path = driver_profile_path
             if driver_profile_path != spawned_profile_path:
                 shutil.rmtree(spawned_profile_path, ignore_errors=True)
             if previous_profile_path is not None:
@@ -414,10 +414,10 @@ class Browser:
                 "BROWSER %i: Archiving browser profile directory to %s"
                 % (self.browser_id, self.browser_params.profile_archive_dir)
             )
-
+            tar_path = self.browser_params.profile_archive_dir / "profile.tar.gz"
             self.browser_params.profile_path = self.current_profile_path
             dump_profile_command = DumpProfileCommand(
-                tar_path=self.browser_params.profile_archive_dir,
+                tar_path=tar_path,
                 close_webdriver=False,
                 compress=True,
             )
@@ -450,8 +450,6 @@ def BrowserManager(
         driver, prof_folder, display = deploy_firefox.deploy_firefox(
             status_queue, browser_params, manager_params, crash_recovery
         )
-        if prof_folder[-1] != "/":
-            prof_folder += "/"
 
         # Read the extension port -- if extension is enabled
         # TODO: Initial communication from extension to TM should use sockets
@@ -462,7 +460,7 @@ def BrowserManager(
             )
             elapsed = 0
             port = None
-            ep_filename = os.path.join(prof_folder, "extension_port.txt")
+            ep_filename = prof_folder / "extension_port.txt"
             while elapsed < 5:
                 try:
                     with open(ep_filename, "rt") as f:
@@ -504,12 +502,6 @@ def BrowserManager(
             command: Union[ShutdownSignal, BaseCommand] = command_queue.get()
 
             if type(command) is ShutdownSignal:
-                # Geckodriver creates a copy of the profile (and the original
-                # temp file created by FirefoxProfile() is deleted).
-                # We clear the profile attribute here to prevent prints from:
-                # https://github.com/SeleniumHQ/selenium/blob/4e4160dd3d2f93757cafb87e2a1c20d6266f5554/py/selenium/webdriver/firefox/webdriver.py#L193-L199
-                if driver.profile and not os.path.isdir(driver.profile.path):
-                    driver.profile = None
                 driver.quit()
                 status_queue.put("OK")
                 return
--- a/openwpm/browser_manager.py
+++ b/openwpm/browser_manager.py
@@ -65,7 +65,7 @@ class Browser:
 
         # Queues and process IDs for BrowserManager
 
-        # thread to run commands issues from TaskManager
+        # thread to run commands issued from TaskManager
         self.command_thread: Optional[threading.Thread] = None
         # queue for passing command tuples to BrowserManager
         self.command_queue: Optional[Queue] = None
@@ -78,7 +78,7 @@ class Browser:
         # the port of the display for the Xvfb display (if it exists)
         self.display_port: Optional[int] = None
 
-        # boolean that says if the BrowserManager new (to optimize restarts)
+        # boolean that says if the BrowserManager is new (to optimize restarts)
         self.is_fresh = True
         # boolean indicating if the browser should be restarted
         self.restart_required = False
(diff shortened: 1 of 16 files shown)


openwpm/browser_manager.py
```diff
@@ -394,33 +401,32 @@ def shutdown_browser(self, during_init: bool, force: bool = False) -> None:
...
                 close_webdriver=False,
-                compress=True
+                compress=True,
+            )
+            dump_profile_command.execute(
+                webdriver=None,
+                browser_params=self.browser_params,
+                manager_params=self.manager_params,
+                extension_socket=None,
             )
```
boolean5, line 429: As I understand this, we can't use `DumpProfileCommand` as part of a command sequence here and in `launch_browser_manager`. Our only choice is to call its `execute()` method directly. Passing `None` to  `webdriver` and `extension_socket` looks a bit weird though. I also had to change the type of `extension_socket` from `ClientSocket` to `Optional[ClientSocket]` in `DumpProfileCommand`'s `execute()` method for mypy to stop complaining.
Do you have a suggestion on how to do this in a better way?

boolean5:
`manual_test.py` [still sets](https://github.com/mozilla/OpenWPM/blob/cb95ecc05fb4618e4275faecbc62b84123ebf1b0/test/manual_test.py#L117) the browser profile via the `FirefoxProfile` class. Should we change it to use a custom profile instead?
My impression is that since we're not interested in being able to archive the profile in the context of `manual_test.py`, there's no need to change it, but I'd like a confirmation on this.

vringar:
Yes, we should use a custom profile Here.
Good job for remembering `manual_test.py`!

test/test_profile.py
```diff
@@ -1,83 +1,81 @@
...
-            manager_params.data_directory, "browser_profile"
-        )
-        return manager_params, browser_params
+def test_saving(default_params, task_manager_creator):
+    manager_params, browser_params = default_params
+    manager_params.num_browsers = 1
+    browser_params[0].profile_archive_dir = (
+        manager_params.data_directory / "browser_profile"
+    )
+    manager, _ = task_manager_creator((manager_params, browser_params[:1]))
```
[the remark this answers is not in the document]
vringar: I mostly did this to make sure our TaskManager tests weren't broken for more than one browser.
Maybe I should have set the default to 1.
I think running with 1 browser here is the right thing to do.

openwpm/browser_manager.py
vringar: Hmm, not really.
Maybe we can extract the actual implementation of the DumpProfileCommand.execute method into a free function and then call it from here and the Command.

englehardt: > Maybe we can extract the actual implementation of the DumpProfileCommand.execute method into a free function and then call it from here and the Command.

This sounds like the right solution to me.

openwpm/deploy_browsers/configure_firefox.py
```diff
@@ -1,7 +1,103 @@
 """ Set prefs and load extensions in Firefox """
 
+import json
+import re
+from pathlib import Path
+from typing import Any, Dict
+
+# TODO: Remove hardcoded geckodriver default preferences once
```
englehardt, line 8: We should add a step into our [release checklist](https://github.com/mozilla/OpenWPM/blob/master/docs/Release-Checklist.md) to check whether these prefs should be updated whenever a new version of geckodriver is released. We should also have an issue on file to remove the hack once the fix ships.

openwpm/deploy_browsers/deploy_firefox.py
```diff
@@ -110,16 +117,30 @@ def deploy_firefox(
         # TODO restore detailed logging
         # fo.set_preference("extensions.@openwpm.sdk.console.logLevel", "all")
 
+    # Geckodriver currently places the user.js file in the wrong profile
+    # directory, so we have to create it manually here.
+    # TODO: Remove this workaround once
```
englehardt, line 122: File a new issue for this removal and reference it here. (We don't want to track TODOs in the codebase).

openwpm/deploy_browsers/deploy_firefox.py
```diff
@@ -110,16 +117,30 @@ def deploy_firefox(
...
+    # Load existing preferences from the profile's user.js file
+    prefs = configure_firefox.load_existing_prefs(browser_profile_path)
+    # Load default geckodriver preferences
+    prefs.update(configure_firefox.DEFAULT_GECKODRIVER_PREFS)
+    # Pick an available port for Marionette (https://stackoverflow.com/a/2838309)
+    s = socket.socket()
+    s.bind(("", 0))
+    marionette_port = s.getsockname()[1]
+    s.close()
+    prefs["marionette.port"] = marionette_port
```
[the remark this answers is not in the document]
englehardt: > The way I've implemented picking a port and passing it to geckodriver has a race condition, as another process may get the port before Marionette does. I think this will rarely happen though. Is it something we can accept? Can we think of a better way to do this?

This seems acceptable. I agree with Stefan that a comment noting this is possible will be helpful.

What happens if this fails? I assume we'll automatically retry?

openwpm/deploy_browsers/deploy_firefox.py
```diff
@@ -129,16 +150,21 @@ def deploy_firefox(
...
     # Launch the webdriver
     status_queue.put(("STATUS", "Launch Attempted", None))
     fb = FirefoxBinary(firefox_path=firefox_binary_path)
     driver = webdriver.Firefox(
-        firefox_profile=fp,
         firefox_binary=fb,
-        firefox_options=fo,
+        options=fo,
         log_path=interceptor.fifo,
+        # TODO: Remove when https://github.com/mozilla/geckodriver/issues/1844
```
englehardt, line 165: Again, better to link to our own issue here.

test/test_crawl.py
```diff
@@ -1,18 +1,17 @@
-# type:ignore
-# As this file is no longer maintained, mypy shouldn't check this
-import os
+"""Runs a short test crawl.
```
englehardt, line 1: Check out https://github.com/mozilla/OpenWPM/issues/197. I disabled this test a while back because I don't love the idea of connecting to real websites during the tests.

Is there a way to restrict this test to only run on CI and not when a user runs the test suite locally?

test/test_crawl.py
```diff
@@ -42,117 +41,99 @@ def get_public_suffix(url):
...
+        ps = get_public_suffix(url)
+        hist_ps.add(ps)
+        statuses[ps] = command_status
+
+    # Grab urls from Firefox database
+    profile_ps = set()  # visited domains from firefox profile
+    rows = db_utils.query_db(ff_db, "SELECT url FROM moz_places")
+    for (host,) in rows:
+        try:
+            profile_ps.add(get_public_suffix(host))
```
englehardt, line 94: Is this actually adding the public suffix (e.g., `com`) or the public suffix + 1 (e.g., `example.com`)? If it's the former, it seems like we should update this to use public suffix + 1. To grab that we can directly use `du.get_ps_plus_1`.

test/test_profile.py
```diff
@@ -1,83 +1,81 @@
...
-        manager.get("example.com")  # Cause a selenium crash
-
-        # The browser will fail to launch due to the proxy crashes
-        try:
-            manager.get("http://example.com")
-        except CommandExecutionError:
-            pass
-        manager.close()
-        assert isfile(join(browser_params[0].profile_archive_dir, "profile.tar.gz"))
+@pytest.mark.skip(reason="proxy no longer supported, need to update")
```
englehardt, line 53: Just a reminder that we also want to figure out how to fix this.

englehardt (changes requested):
This is great! Thank you Georgia! I left a bunch of comments, but nothing major.

boolean5:
> Yes, we should use a custom profile Here.
> Good job for remembering `manual_test.py`!

`manual_test.py` happened to be broken for a different reason and also needed another update (check f5bacaed84ae624b0a9d602bd4f28770976b8164)

openwpm/deploy_browsers/configure_firefox.py
boolean5: Done! Here's the issue: https://github.com/mozilla/OpenWPM/issues/867

test/test_crawl.py
boolean5: Sure there is, I added this

```python
@pytest.mark.skipif(
    "CI" not in os.environ or os.environ["CI"] == "false",
    reason="Makes remote connections",
)
```

boolean5:
I assume that in these three cases [[1]](https://github.com/mozilla/OpenWPM/blob/b6849c71f53185d1202d862ed34ef71d9c8cc860/crawler.py#L223) [[2]](https://github.com/mozilla/OpenWPM/blob/b6849c71f53185d1202d862ed34ef71d9c8cc860/test/test_callback.py#L20) [[3]](https://github.com/mozilla/OpenWPM/blob/b6849c71f53185d1202d862ed34ef71d9c8cc860/test/test_profile.py#L92) `reset=True` was used because stateful crawling was disabled and we can now let it take the default `False` value. Is there a reason that I'm missing to keep `reset=True` for any of them?

englehardt:
> I assume that in these three cases [[1]](https://github.com/mozilla/OpenWPM/blob/b6849c71f53185d1202d862ed34ef71d9c8cc860/crawler.py#L223) [[2]](https://github.com/mozilla/OpenWPM/blob/b6849c71f53185d1202d862ed34ef71d9c8cc860/test/test_callback.py#L20) [[3]](https://github.com/mozilla/OpenWPM/blob/b6849c71f53185d1202d862ed34ef71d9c8cc860/test/test_profile.py#L92) `reset=True` was used because stateful crawling was disabled and we can now let it take the default `False` value. Is there a reason that I'm missing to keep `reset=True` for any of them?

For [1] we should keep reset=True. We want those crawls to be stateless (of course other users can change as they see fit). For [2] and [3] I'm not sure it matters, but I'll defer to @vringar just in case.

vringar:
[2] and [3] can be removed as well.

test/test_crawl.py
boolean5: Good catch! This was adding the publix suffix indeed so I changed it to use `du.get_ps_plus_1`.
Also, after looking at the test more closely, I noticed that one of the queries to the crawl database was not formed correctly (see 9e8298c).

openwpm/deploy_browsers/deploy_firefox.py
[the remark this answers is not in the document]
boolean5: Yes, the browser manager is automatically restarted

boolean5:
As we discussed, I checked and the original tar file is kept intact after extracting, so I removed the intermediate step of [copying](https://github.com/mozilla/OpenWPM/blob/ab01a2f6bdadcf75fe8c2878bc8d6610f5ed2a7a/openwpm/commands/profile_commands.py#L131L141) it from `load_profile()`.

test/test_profile.py
boolean5: Let's track this in https://github.com/mozilla/OpenWPM/issues/789, so that it doesn't delay this PR. I'll add a comment there.

boolean5:
@vringar, @englehardt, thanks for reviewing! This is ready, I've covered all the requested changes. You can take another look if you'd like.

test/test_profile.py
```diff
@@ -89,7 +87,7 @@ def test_seed_persistance(default_params, task_manager_creator):
 
     command_sequences = []
     for _ in range(2):
-        cs = CommandSequence(url="https://example.com", reset=True)
+        cs = CommandSequence(url="https://example.com")
```
vringar, line 90: Just realizing we're making an outbound connection here.
Could you change this to use our local server or should I do that in a different PR?

boolean5: Hm, this is also true for `test_saving` and `test_crash_profile`. I'll push a commit to fix that. Thanks for pointing it out!

openwpm/deploy_browsers/selenium_firefox.py
```diff
@@ -129,7 +122,7 @@ def __init__(
         self.service_args = service_args or []
 
     def command_line_args(self):
-        return ["--port", "%d" % self.port]
+        return ["--port", "%d" % self.port] + self.service_args
 
     def send_remote_shutdown_command(self):
```
vringar, line 127: Do you know why we pass here? What are we trying to supress/stop from happening?

vringar (approved):
Do we delete the profile folders on a normal shutdown? I don't think we need to but maybe we should still do it.

openwpm/deploy_browsers/selenium_firefox.py
boolean5: Do you mean why we pass in `send_remote_shutdown_command()`? class `PatchedGeckoDriverService` is a patched version of [Selenium's `Service` class](https://github.com/SeleniumHQ/selenium/blob/0849604f66a3e15e7ae650cfb0ac870328ce271f/py/selenium/webdriver/firefox/service.py) and we've copied this method from there.

vringar: Ah okay, thank you. 
Is there a reason we shouldn't inherit from the Firefox service and overwrite `__init__` instead of having to keep up with the changes in Selenium?

boolean5:
> Do we delete the profile folders on a normal shutdown? I don't think we need to but maybe we should still do it.

I think we do that [here](https://github.com/mozilla/OpenWPM/blob/49965db164e424f174dd749a87902ad95e10db63/openwpm/browser_manager.py#L419L421), in `shutdown_browser()`. Does that cover what you had in mind?

vringar:
> > Do we delete the profile folders on a normal shutdown? I don't think we need to but maybe we should still do it.
> 
> I think we do that [here](https://github.com/mozilla/OpenWPM/blob/49965db164e424f174dd749a87902ad95e10db63/openwpm/browser_manager.py#L419L421), in `shutdown_browser()`. Does that cover what you had in mind?

Yeah. Sorry for missing this and thanks for pointing it out.

openwpm/deploy_browsers/selenium_firefox.py
boolean5: I was wrong. `PatchedGeckoDriverService` is a subclass of the `Service` class in [`selenium.webdriver.common.service`](https://github.com/SeleniumHQ/selenium/blob/0849604f66a3e15e7ae650cfb0ac870328ce271f/py/selenium/webdriver/common/service.py) and not [`selenium.webdriver.firefox.service`](https://github.com/SeleniumHQ/selenium/blob/0849604f66a3e15e7ae650cfb0ac870328ce271f/py/selenium/webdriver/firefox/service.py) as I initially thought. That explains why it has to define the`command_line_args()` and `send_remote_shutdown_command()` methods.

I guess we could still do what you are suggesting though. Do you think we should?

vringar: Yes, because this way we only need to check what changed in the `__init__` Method of `selenium.webdriver.firefox.service` and not all changes to that class.

vringar (approved):

englehardt (approved):

birdsarah:
WOOOOTTTT!!!!

vringar:
Not officially released as of yet though. Gonna drop it with the next FF release.

PR #866, opened 2021-03-12
Release preparations

vringar:


CHANGELOG.md
```diff
@@ -1,12 +1,69 @@
...
+  It restructured a lot of internal code to bundle the definition of a command
+  with its implementation
+
+- Data Aggregator Rewrite (#753)  
+  This PR rewrote the way we do data saving from the ground up to enable a more
+  modular and scalable design.
+  Users are now empowered to provide their own storage backend if need be and can
+  easily pick and choose between provided implementations.
+
+Minor changes:
```
englehardt, line 30:
```suggestion
Additional changes:
```

openwpm/storage/cloud_storage/s3_storage.py
```diff
@@ -1,5 +1,103 @@
...
 from ..arrow_storage import ArrowProvider
+from ..storage_providers import TableName, UnstructuredStorageProvider
+
+
+class S3StructuredProvider(ArrowProvider):
+    """This class allows you to upload Parquet files to S3.
+
+    Inspired by the old S3Aggregator structure the S3StructuredProvider
+    will by default store into
+    base_path/visits/table_name in the given bucket.
```
englehardt, line 17:
```suggestion
    S3StructuredProvider will by default store into
    base_path/visits/table_name in the given bucket.
```

I don't think it's necessary to reference old code here.

openwpm/storage/cloud_storage/s3_storage.py
```diff
@@ -1,5 +1,103 @@
...
+
```
englehardt, line 18:
```suggestion
```

openwpm/storage/cloud_storage/s3_storage.py
```diff
@@ -1,5 +1,103 @@
...
+    async def write_table(self, table_name: TableName, table: Table) -> None:
+        self.file_system.start_transaction()
+        pq.write_to_dataset(
+            table,
+            self.base_path.format(table_name=table_name),
+            filesystem=self.file_system,
+        )
+        self.file_system.end_transaction()
+
+    async def shutdown(self) -> None:
```
englehardt, line 51: Are you sure we want to pass here? This overwrites ArrowProvider's default method which logs an error if there are any unwritten batches. That still seems useful.

I _think_ we can just drop this.

englehardt (changes requested):
Do we have any way of testing the S3 storage providers? We certainly don't want to add back in localstack and they are pretty simple, but I'm wondering what our options are.

openwpm/storage/cloud_storage/s3_storage.py
```diff
@@ -1,5 +1,103 @@
...
+        pass
+
```
vringar, line 53:
```suggestion
```

vringar:
One option I can think of is to write integration tests that use docker-compose.
We could build on top of what we have in https://github.com/mozilla/openwpm-crawler/tree/master/deployment/local-compose and our local test server to run crawls that write to localstack and then check that the files stored there are identical to the ones served (for content) and that the records contain everything we expect them to.

englehardt:
> One option I can think of is to write integration tests that use docker-compose.
> We could build on top of what we have in https://github.com/mozilla/openwpm-crawler/tree/master/deployment/local-compose and our local test server to run crawls that write to localstack and then check that the files stored there are identical to the ones served (for content) and that the records contain everything we expect them to.

Localstack has caused us so many issues that I'd prefer to avoid it, even if we're just using their docker container.

> However for now, I'd argue that we are providing users with enough flexibility that they can just use whatever works for them when using s3fs directly and copy paste these arguments into the S3Providers

What do you mean by `and copy paste these arguments into the S3Providers`?

vringar:
I
> Localstack has caused us so many issues that I'd prefer to avoid it, even if we're just using their docker container.

While I understand that feeling Localstack is the only way known to me that would allow us to test the S3StorageProviders locally

> > However for now, I'd argue that we are providing users with enough flexibility that they can just use whatever works for them when using s3fs directly and copy paste these arguments into the S3Providers
> 
> What do you mean by `and copy paste these arguments into the S3Providers`?

E.g. if a user got `S3FileSystem(key="bla", secret="blub")` to work then they could do `S3StructuredProvider("my_bucket_name", "base_path", key="bla", secret="blub")`, which would be copying and pasting the constructor arguments and since we just forward them, it should still work.

vringar:
@englehardt it's ready for merge now

englehardt (approved):

englehardt:
> @englehardt it's ready for merge now

LGTM, thanks!

PR #871, opened 2021-03-16
Restoring Docker build

### LINKED ISSUE #848: Add CI job for Docker builds
PR https://github.com/mozilla/OpenWPM/pull/838 moved our CI from Travis to Github Actions but did not include a job to build the Docker image and deploy it to DockerHub. We should add one that does this.

vringar:
Closes #848

### DIFF
--- a/.github/workflows/build-container.yaml
+++ b/.github/workflows/build-container.yaml
@@ -18,15 +18,16 @@ jobs:
           DOCKER_IMAGE=openwpm/openwpm
           TAGS="${DOCKER_IMAGE}:${GITHUB_SHA}"
           # If it's a tagged build it should get published as such
-          if [[ $GITHUB_REF == refs/tags/* ]]; then
-              VERSION=${GITHUB_REF#refs/tags/v}
-              TAGS="$TAGS,${DOCKER_IMAGE}:latest"
+          if [[ $GITHUB_REF == refs/tags/v* ]]; then
+            VERSION=${GITHUB_REF#refs/tags/v}
+            TAGS="$TAGS,${DOCKER_IMAGE}:$VERSION"
           fi
           # If it is on master it should get tagged as latest
           if [[ $GITHUB_REF == refs/heads/master ]]; then
               TAGS="$TAGS,${DOCKER_IMAGE}:latest"
           fi
           echo ::set-output name=tags::${TAGS}
+
       - 
         name: Checkout
         uses: actions/checkout@v2
--- a/.github/workflows/build-container.yaml
+++ b/.github/workflows/build-container.yaml
@@ -1,0 +1,47 @@
+name: Building and publishing container
+on:
+  push:
+    branches:
+      - master
+    tags:
+      - "v*.*.*"
+
+jobs:
+  build_and_publish:
+    runs-on: ubuntu-latest
+    steps:
+      - 
+        name: Checkout
+        uses: actions/checkout@v2
+      -       
+        name: Set up Docker Buildx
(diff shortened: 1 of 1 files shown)


.github/workflows/build-container.yaml
```diff
@@ -0,0 +1,47 @@
+name: Building and publishing container
+on:
+  push:
+    branches:
+      - master
+    tags:
+      - "v*.*.*"
```
vringar, line 7: Should the action be constrained like this or build on PRs as well.
Also I still need to figure out how to propagate the tag.

englehardt: > Should the action be constrained like this or build on PRs as well.

Is it possible to look for one of two tags? We can keep the `v*.*.*` tag, but add another one that's something like `force-docker-build` in the event we want some specific PR to update the image.

> Also I still need to figure out how to propagate the tag.

I'm not sure what you mean by this.

vringar: That the container we build off a tagged commit gets that same tag on dockerhub

vringar:
https://hub.docker.com/repository/docker/openwpm/openwpm
It's back :tada: 
However I'm unsure why it's not tagged as `force-docker-build` but instead as `latest`

englehardt:
> https://hub.docker.com/repository/docker/openwpm/openwpm
> It's back tada
> However I'm unsure why it's not tagged as `force-docker-build` but instead as `latest`

Hmm. I'm not sure we want it tagged as `force-docker-build` anyway. I see the value in having `latest` point to the latest versioned release. Maybe we just tag builds from `force-docker-build` with the commit hash and that's it?

vringar:
This PR is now ready for merging.

.github/workflows/build-container.yaml
```diff
@@ -0,0 +1,65 @@
...
+      - "force-docker-build"
```
boolean5, line 8: Using the `force-docker-build` tag to trigger the workflow for PRs seems a little bit impractical to me. We would have to remember to explicitly delete the tag in order to be able to reuse it. This could be rather annoying if we want to use the tag on multiple PRs that are open at the same time.

We can keep this as it is, but a more viable alternative would be to add the `workflow_dispatch` event to `build-container.yaml`. This enables us to manually trigger the workflow on any branch. See [here](https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/) for an example (note: this example provides inputs for the workflow but we don't need any)

.github/workflows/build-container.yaml
```diff
@@ -0,0 +1,65 @@
...
+        uses: docker/setup-buildx-action@v1
+      -     
+        name: Cache Docker layers
+        uses: actions/cache@v2
+        with:
+          path: /tmp/.buildx-cache
+          key: ${{ runner.os }}-buildx-${{ github.sha }}
+          restore-keys: |
+            ${{ runner.os }}-buildx-
+      -
```
boolean5, line 46:
```suggestion
      - name: Login to Docker Hub
```
We could add a name to the login action.
And perhaps we could have the hyphens on the same line as  `name:` throughout the file? That's an insignificant detail really, I just found it a little confusing.

boolean5 (approved):

vringar:
I tried using `workflow_dispatch` but don't see the option to trigger it manually where I'd expect it.
image

Does this option only become unlocked once it's merged to master?

vringar:
Turns out yes, it needed to be merged.

PR #872, opened 2021-03-18
Quick extension code cleanup

englehardt:
This removes some commented out code that was left behind from the upgrade.

 I also enable logging speculative connections, as there's no reason for us to skip logging. I suspect they're actually disabled anyway via [these prefs](https://github.com/mozilla/OpenWPM/blob/b6849c71f53185d1202d862ed34ef71d9c8cc860/openwpm/deploy_browsers/configure_firefox.py#L70), but if not they should definitely be logged.


openwpm/Extension/webext-instrumentation/src/lib/http-post-parser.ts
```diff
@@ -70,367 +41,6 @@ export class HttpPostParser {
         ),
       };
     }
-
-    // Return empty response until we have all instrumentation converted
```
vringar, line 74: Do you know what this refers to?

vringar (approved):
I wonder if that line should be an issue (or already is an issue) except for that r+

openwpm/Extension/webext-instrumentation/src/lib/http-post-parser.ts
englehardt: I think this was just precautionary since the previous file did a lot of parsing that's no longer necessary. We currently cover all properties of the requestBody. See [MDN](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/webRequest/onBeforeRequest#details).

PR #875, opened 2021-03-23
Hotfixed GetCommand

vringar:
`webdriver.switch_to.alert` unlike most other variants of the `switch_to` API is not a function but a property.
This led to `TypeError:'Alert' object is not callable` when there was actually an `Alert` to switch to.
This PR fixes that behaviour.


vringar:
I'm planning to cut a v0.14.1 release as soon as this is fixed, as it's a major breakage in one of our most central commands.

englehardt (approved):
Good catch! Thanks for the quick hotfix here.

PR #879, opened 2021-03-26
Remove python to js string

### LINKED ISSUE #857: Remove _python_to_js_string
https://github.com/mozilla/OpenWPM/blob/078395c4ad148a60e20a6dbac549a2287fde2cc3/openwpm/js_instrumentation.py#L23-L37
This function creates a string that looks like JSON while being invalid JSON due to the fact that all `object` fields are unquoted.
We decided to do this, so that once we had de-serialized the object in the WebExtension, the `object` fields directly reference the objects that need instrumenting.
However we have now realized, that it's very easy to get confused and think the string is actually valid JSON.
This has led to a number of unfortunate surprises.
The solution is to eval the `object` field in the WebExtension instead.

vringar:
We now pass along a Dict/JS object for as long as possible, only serialize it to valid JSON and eval at the last reasonable location.
Closes #857

### DIFF
--- a/openwpm/utilities/platform_utils.py
+++ b/openwpm/utilities/platform_utils.py
@@ -125,7 +125,7 @@ def get_configuration_string(manager_params, browser_params, versions):
         # Separate out long profile directory strings
         profile_dirs[browser_id] = str(item.pop("seed_tar"))
         archive_dirs[browser_id] = str(item.pop("profile_archive_dir"))
-        js_config[browser_id] = item.pop("js_instrument_settings")
+        js_config[browser_id] = item.pop("cleaned_js_instrument_settings")
 
         # Copy items in sorted order
         dct = OrderedDict()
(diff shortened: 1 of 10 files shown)


openwpm/utilities/platform_utils.py
```diff
@@ -125,7 +125,7 @@ def get_configuration_string(manager_params, browser_params, versions):
         # Separate out long profile directory strings
         profile_dirs[browser_id] = str(item.pop("seed_tar"))
         archive_dirs[browser_id] = str(item.pop("profile_archive_dir"))
-        js_config[browser_id] = item.pop("js_instrument_settings")
+        js_config[browser_id] = item.pop("cleaned_js_instrument_settings")
```
vringar, line 128: Should we pop off both?

englehardt: What do you mean by both?

englehardt (approved):
This LGTM, though I'm not sure what your question is asking?

openwpm/utilities/platform_utils.py
vringar: Now we only pop off the expanded version, however if a user were to supply their own expanded config, that would still ruin our printing.

englehardt: I see. If both are present than it makes sense to remove both from printing.

PR #881, opened 2021-04-06
Clean up some unused files

boolean5:
This PR removes some files that don't seem to be used anymore:

1. [`deploy-to-dockerhub.sh`](https://github.com/mozilla/OpenWPM/blob/8030c2c470015350f510af4e42f1389e1ffe1c1c/scripts/deploy-to-dockerhub.sh) was used on Travis but now we have a Github Actions workflow for this.
2. [`profile.tar.gz`](https://github.com/mozilla/OpenWPM/blob/8030c2c470015350f510af4e42f1389e1ffe1c1c/profile.tar.gz) was introduced in https://github.com/mozilla/OpenWPM/pull/736 (not really sure why) but is not used by any tests now. Do we have a reason for keeping it around?
3. [`openwpm/commands/utils/file_utils.py`](https://github.com/mozilla/OpenWPM/blob/8030c2c470015350f510af4e42f1389e1ffe1c1c/openwpm/commands/utils/file_utils.py) was introduced in https://github.com/mozilla/OpenWPM/commit/b6fe171d47b43c55305da1ba4ad9bbb9c7141957 and used in a flash related function that now doesn't exist.


vringar (approved):

vringar:
I'll wait for the tests to pass and merge it.

PR #882, opened 2021-04-06
Moved _issue_command to BrowserManagerHandle

vringar:
This piece of code conceptually belongs to the BrowserManagerHandle, so I moved it there.
However I couldn't think of an easy way to remove all references to the TaskManager so for now we still pass the TaskManager as a parameter.

### DIFF
--- a/openwpm/browser_manager.py
+++ b/openwpm/browser_manager.py
@@ -1,4 +1,5 @@
 import errno
+import json
 import logging
 import os
 import pickle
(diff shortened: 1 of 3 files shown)


openwpm/browser_manager.py
```diff
@@ -11,19 +12,28 @@
...
 from pathlib import Path
 from queue import Empty as EmptyQueue
-from typing import Optional, Union
+from typing import TYPE_CHECKING, Optional, Tuple, Union
 
 import psutil
+import tblib
 from multiprocess import Queue
 from selenium.common.exceptions import WebDriverException
 from tblib import pickling_support
```
boolean5, line 21:
```suggestion
```

openwpm/browser_manager.py
```diff
@@ -34,9 +44,9 @@
 pickling_support.install()
```
boolean5, line 44:
```suggestion
tblib.pickling_support.install()
```

boolean5 (changes requested):

boolean5:
We should also update the docs wherever there are references to the `Browser` class. With a quick look I just found [one](https://github.com/mozilla/OpenWPM/blame/master/docs/Platform-Architecture.md#L62) in `Platform-Architecture.md`.

vringar:
Thank you so much for looking into this and figuring out where my error was.

boolean5 (approved):

Issue #890, opened 2021-04-15
Readthedocs build failing 


vringar:
The RTD builder is failing on the following command `conda env create --quiet --name latest --file environment.yaml` with the following error message
```
Collecting package metadata: ...working... done
Solving environment: ...working... failed

ResolvePackageNotFound: 
  - geckodriver=0.29.0 -> __glibc[version='>=2.17,<3.0.a0']
```

I think I've seen this error before, but I don't remember the context.

vringar:
Okay, so the conda version read the docs uses is out of date and we need to downgrade our geckodriver to 0.26.0 for the build to work.

vringar:
After some searching I found [this issue](https://github.com/readthedocs/readthedocs-docker-images/issues/106) and then following from there [this ](https://github.com/Ouranosinc/raven/pull/167/files]
So we could also just define a `requirements-docs.yaml` that we try to keep in sync with the main `requirements.yaml` but remove all dependencies that we don't need for generating docs.

PR #893, opened 2021-04-19
Update profile tests

boolean5:
This PR  updates some of the existing profile tests.

### DIFF
--- a/test/test_profile.py
+++ b/test/test_profile.py
@@ -52,26 +52,23 @@ def test_profile_error(default_params, task_manager_creator):
         task_manager_creator((manager_params, browser_params[:1]))
 
 
-@pytest.mark.skip(reason="proxy no longer supported, need to update")
-def test_profile_saved_when_launch_crashes(default_params, task_manager_creator):
+def test_profile_saved_when_launch_crashes(
+    monkeypatch, default_params, task_manager_creator
+):
     manager_params, browser_params = default_params
     manager_params.num_browsers = 1
     browser_params[0].profile_archive_dir = (
         manager_params.data_directory / "browser_profile"
     )
-    browser_params[0].proxy = True
-    browser_params[0].save_content = "script"
     manager, _ = task_manager_creator((manager_params, browser_params[:1]))
     manager.get(BASE_TEST_URL)
 
-    # Kill the LevelDBAggregator
-    # This will cause the proxy launch to crash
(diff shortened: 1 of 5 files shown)


test/test_profile.py
```diff
@@ -52,26 +52,23 @@ def test_profile_error(default_params, task_manager_creator):
...
-    browser_params[0].proxy = True
-    browser_params[0].save_content = "script"
     manager, _ = task_manager_creator((manager_params, browser_params[:1]))
     manager.get(BASE_TEST_URL)
 
-    # Kill the LevelDBAggregator
-    # This will cause the proxy launch to crash
-    manager.ldb_status_queue.put("DIE")
+    # This will cause browser restarts to fail
+    monkeypatch.setenv("FIREFOX_BINARY", "/tmp/NOTREAL")
```
vringar, line 67: This is an absolutely genius way to introduce a crash at startup. Well done!

boolean5: Thank you! An alternative would be to unset one of [these two preferences](https://github.com/mozilla/OpenWPM/blob/625d81460b25d68846de36ddeeab23c9731bbda6/openwpm/deploy_browsers/configure_firefox.py#L251L252). This would result in the error observed in issue https://github.com/mozilla/OpenWPM/issues/606, because the extension won't be able to start.

boolean5:
Regarding the update of `test_browser_profile_coverage`:
As we discussed, because we haven't observed any missing urls so far, I removed the [buggy](https://github.com/mozilla/OpenWPM/blob/6efeade3597bdbd5c194039e708bb6145160aa49/test/test_crawl.py#L110) and [outdated](https://github.com/mozilla/OpenWPM/blob/6efeade3597bdbd5c194039e708bb6145160aa49/test/test_crawl.py#L121L127) for loop that determined whether a url is expected to be missing.

In case we observe failures of `test_browser_profile_coverage`, we should update and restore this part of the code. For this purpose, I've left a comment in the test pointing to this PR.

boolean5:
Since we decided to postpone working on the profile tests, I changed the scope of this PR to only include updates to tests so that it can be merged now.

Note that the failure in building the docs is not related to this PR, it was because of https://github.com/mozilla/OpenWPM/issues/890 and has now been fixed on the `master` branch.

vringar (approved):

PR #894, opened 2021-04-20
Created environment-rtd.yaml

### LINKED ISSUE #890: Readthedocs build failing

vringar:
Hopefully now read the docs will build.
Closes #890


vringar:
https://openwpm--894.org.readthedocs.build/en/894/

boolean5 (commented):
I thought `environment-rtd.yaml` would only contain the dependencies required to build the docs. This would also make readthedocs builds faster. Is there a reason to keep all our dependencies in this file?

Later we could also update our scripts to create this file automatically, so we don't have to keep it synced with `environment.yaml` manually. No need to do this in this PR however.

vringar:
> Is there a reason to keep all our dependencies in this file?

No, there isn't. I thought I still had a sphinx extension that did some `mypy` like type inference.
But since that is not the case, I'll remove the other dependencies.

boolean5 (approved):

Issue #892, opened 2021-04-17
Missing data when using browser restarts


bkrumnow:
Hey everybody,

I stumbled upon an inconsistency in the data collection produced by OpenWPM. In brief, I found that sometimes it takes a browser a bit longer to start all instruments, which can be long enough that a crawl starts before the recording has begun. 

**The setup:**
I conducted an experiment where multiple crawlers visit the same site in parallel (nearly synchronised). Each browser restarts after visiting a single site thanks to the reset option of the command sequence.

```
cs = CommandSequence(
  site,
  site_rank=rank,
  reset=True,
  callback=callback,
)
```
I used multiple instruments:  HTTP, JS cookies, navigations,  javascript, and the call stack instrument.
My tests include Ubuntu with OpenWPM v0.13.0 as well as OpenWPM v0.14.0 (Firefox 86.0.1 on both).

**The problem:**
I saw that multiple sites in my data set had a deviating number of requests. For example, the number of requests for a site can look like this:

<img width="589" alt="Screenshot 2021-04-17 at 14 18 32" src="https://user-images.githubusercontent.com/3897761/115112885-cf2a7680-9f87-11eb-9669-3f135311d7e6.png">

While this is usually not surprising, there is quite a number of sites with a varying number of first-party requests, which raised my attention.  I tracked down which are missing and found that requests at the beginning of a visit are often not recorded. For the example above, one crawler did not record the very first two requests.

![Screenshot 2021-04-15 at 14 32 54](https://user-images.githubusercontent.com/3897761/115112927-fd0fbb00-9f87-11eb-9fe8-c3ef5a7fd816.png)

I checked how many times the URL of the first recorded request differs for crawls of the same site. In an experiment where I scanned 100 sites, I found that the results on 88 sites are affected (8 crawlers per site). However, when this happens and which browser is affected is inconsistent.

**What is causing this?**
The current logging makes it a bit difficult to spot such kind of problems. I added logging to the extension to see when an instrument is done loading. 

https://github.com/mozilla/OpenWPM/blob/625d81460b25d68846de36ddeeab23c9731bbda6/openwpm/Extension/firefox/feature.js/index.js#L56-L93

**Side note:** OpenWPM-0.14.0 does not show any logging of the extension, but the logging in the extension is still the same. Was that done on purpose?

The log for a site looks like this: 
```
[...]

PR #897, opened 2021-04-22
Introduced delay in BrowserManager

vringar:
This PR is a quick-fix response to #892
By delaying the startup of a browser if we have the extension enabled, the extension should be able to fully start up before we start executing commands.

@bkrumnow could you please test this change and let me know if it resolves the issue?


bkrumnow:
I checked it on thirty sites. There were no missing requests at the beginning.

vringar:
Closing in favor of #902

PR #898, opened 2021-04-22
Expand socket capabilities

### LINKED ISSUE #892: Missing data when using browser restarts

vringar:
This was originally a PR aimed to address #892 but with help of nhnt11 I came up with a radically simpler solution (See #902)
 
This PR introduce the ability for our python classes ServerSocket and ClientSocket to both receive and send messages.
However for the server, we can't really send a message over just one connection, so we are broadcasting to all active connections.

Do we want this capability or is this not useful for our use case?

Please also note that this only works for Python to Python sockets, as I couldn't figure out how to implement this for WebExtension.


vringar:
I will also be adding tests for this shortly.

Issue #895, opened 2021-04-20
Generate Module index


vringar:
Somehow we aren't generating the module index right now.
This needs to be fixed ASAP

boolean5:
It seems that for the module index to be created we have to run `sphinx-apidoc -o source/ ../openwpm` (from within the `docs` directory) and then commit the autogenerated `rst` files ([[1]](https://sphinx-rtd-tutorial.readthedocs.io/en/latest/build-the-docs.html#generating-documentation-from-docstrings), [[2]](https://samnicholls.net/2016/06/15/how-to-sphinx-readthedocs/)).

This however, has the obvious drawback that we should remember to regenerate and commit these files whenever the structure of modules in our project changes, otherwise the documentation will be out of date. We could add a note about that in `CONTRIBUTING.md`, but it might be better if we could have readthedocs automatically run `sphinx-apidoc` before every build.

After looking into it a bit more, I came across this workaround, which does exactly that: https://github.com/readthedocs/readthedocs.org/issues/1139#issuecomment-312626491 

@vringar, I could open a PR if you haven't started working on something already.

vringar:
Thanks for doing this investigation

I haven't started working on this yet so I'd appreciate a PR

PR #900, opened 2021-04-26
Generate module index

boolean5:
This PR fixes https://github.com/mozilla/OpenWPM/issues/895 by setting up Sphinx to automatically run `sphinx-apidoc` for every build, using the workaround suggested in https://github.com/readthedocs/readthedocs.org/issues/1139#issuecomment-312626491. Also, it moves readthedocs dependencies under `docs/` and makes `prune-environment.py` automatically generate the `environment-rtd.yaml` file whenever we run `repin.sh`, so that we don't have to remember to keep it in sync with the main `environment.yaml` manually.

Note that we have to install all dependencies and not just the ones required to build the docs after all, as `autodoc` [imports the modules](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#module-sphinx.ext.autodoc) in order to extract the docstrings. Not installing all dependencies would result in import errors.

@vringar, could you please point me to where you found the conda version that readthedocs uses, that you mentioned in https://github.com/mozilla/OpenWPM/issues/890#issuecomment-820400447? I'd like to open an issue to revert our workaround of keeping `geckodriver` pinned to version `0.26.0` when readthedocs updates its conda version.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -13,7 +13,10 @@
 import os
 import sys
 
-sys.path.insert(0, os.path.abspath(".."))
+from sphinx.ext import apidoc
+
+project_root = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))
+sys.path.insert(0, project_root)
 
 
 # -- Project information -----------------------------------------------------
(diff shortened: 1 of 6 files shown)


docs/conf.py
```diff
@@ -67,3 +70,14 @@
 # relative to this directory. They are copied after the builtin static files,
 # so a file named "default.css" will overwrite the builtin "default.css".
 # html_static_path = ["_static"]
+
+
+# Auto-generate API documentation
+# Source: https://github.com/readthedocs/readthedocs.org/issues/1139
+def run_apidoc(_):
```
boolean5, line 77: @vringar, do you know how I should type annotate the `_` parameter? Or should we just tell `mypy` to ignore function definitions in this file?

vringar: You could annotate it with Any
```suggestion
def run_apidoc(_: Any) -> None:
```
so that mypy stops complaining.
But I think telling mypy to not enforce type annotations in the `docs` folder would be better.

vringar (approved):

PR #901, opened 2021-04-26
Browser manager as class

vringar:
This PR refactors the BrowserManager to be a class inheriting from multiprocess.Process.
This change allows the logic of BrowserManager to be broken up into multiple pieces as indicated by `_start_extension`

### DIFF
--- a/openwpm/browser_manager.py
+++ b/openwpm/browser_manager.py
@@ -12,10 +12,10 @@ import time
 import traceback
 from pathlib import Path
 from queue import Empty as EmptyQueue
-from typing import TYPE_CHECKING, Optional, Tuple, Union
+from typing import TYPE_CHECKING, Any, Optional, Tuple, Union
 
 import psutil
-from multiprocess import Queue
+from multiprocess import Process, Queue
 from selenium.common.exceptions import WebDriverException
 from tblib import Traceback, pickling_support
 
@@ -31,7 +31,6 @@ from .socket_interface import ClientSocket
 from .storage.storage_providers import TableName
 from .types import BrowserId, VisitId
 from .utilities.multiprocess_utils import (
-    Process,
     kill_process_and_children,
     parse_traceback_for_sentry,
 )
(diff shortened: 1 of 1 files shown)


openwpm/browser_manager.py
```diff
@@ -137,7 +135,8 @@ def launch_browser_manager(self):
         unsuccessful_spawns = 0
         success = False
 
-        def check_queue(launch_status):
+        def check_queue(launch_status: Dict[Any, Any]) -> Any:
```
boolean5, line 138: `launch_status` could be `Dict[str, bool]`, right?

openwpm/browser_manager.py
```diff
@@ -156,19 +155,19 @@ def check_queue(launch_status):
...
                 self.manager_params,
                 crash_recovery,
             )
-            self.browser_manager = Process(target=BrowserManager, args=args)
             self.browser_manager.daemon = True
             self.browser_manager.start()
 
             # Read success status of browser manager
-            launch_status = dict()
+            launch_status: Dict[Any, Any] = dict()
```
boolean5, line 170: Same here

openwpm/browser_manager.py
```diff
@@ -612,130 +611,158 @@ def shutdown_browser(self, during_init: bool, force: bool = False) -> None:
...
+            self.logger.error(
+                "BROWSER %i: Crash in driver, restarting browser manager"
+                % self.browser_params.browser_id,
+                exc_info=True,
+                extra=extra,
+            )
+            self.status_queue.put(("FAILED", pickle.dumps(sys.exc_info())))
+        finally:
+            if display is not None:
+                display.stop()
```
boolean5, line 767: The only difference I see from before is that now that we don't use the [`Process` wrapper](https://github.com/mozilla/OpenWPM/blob/cc10baebee4a0f3fbec38eeac17586448c903e60/openwpm/utilities/multiprocess_utils.py#L35L50) from `utilities.multiprocess_utils` if an exception is raised here it won't be logged. But I don't know if that's really a problem.

boolean5 (approved):
Looks good! Feel free to merge after taking a look at the comments.

openwpm/browser_manager.py
vringar: Yeah, that was my thinking as well. 
I think it's highly unlikely that the display stopping will fail and all other exceptions are caught in the except block above.

PR #908, opened 2021-04-29
Added npm script info under scripts

### LINKED ISSUE #870: Add `"info": "npm-scripts-info"` under `scripts`
We should probably add `"info": "npm-scripts-info"` under `scripts` as described [here](https://www.npmjs.com/package/npm-scripts-info) so that the descriptions under `scripts-info` can be displayed. But it's probably best if I open another PR for that as it's out of the scope of this one.

_Originally posted by @boolean5 in https://github.com/mozilla/OpenWPM/pull/865#discussion_r595074298_

Ayushsunny:
Added  "info": "npm-scripts-info" under scripts as described. so now the descriptions under scripts-info can be displayed.
Closes #870

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -94,6 +94,18 @@
         "sprintf-js": "~1.0.2"
       }
     },
+    "array-find-index": {
+      "version": "1.0.2",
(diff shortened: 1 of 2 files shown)


vringar (approved):
Thanks for implementing this!

PR #917, opened 2021-05-07
Save full Firefox profile

### LINKED ISSUE #62: Expand saved Firefox profile to full directory
Right now we restrict ourselves to just copying a few subcomponents of the profile. i.e. cookies and localStorage. Why not copy all files?

There may be some additional concerns such as extensions present in the profile directory that a user may not expect, but we can either constrain the browser configuration to be the same, or load and save the configuration alongside the profile.

boolean5:
This PR changes [`dump_profile`](https://github.com/mozilla/OpenWPM/blob/dafc26cb562a9d1a7444bb317a12ff69121441a7/openwpm/commands/profile_commands.py#L18L90) so that it saves the whole Firefox profile directory instead of only saving a few of its subcomponents. It also adds a test and expands an existing one in order to test this new feature.

Closes #62.

### DIFF
--- a/test/test_profile.py
+++ b/test/test_profile.py
@@ -1,3 +1,4 @@
+import tarfile
 from pathlib import Path
 from typing import Any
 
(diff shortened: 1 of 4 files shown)


test/test_profile.py
```diff
@@ -22,7 +23,36 @@ def test_saving(default_params, task_manager_creator):
...
+    tar_path = browser_params[0].profile_archive_dir / "profile.tar.gz"
+    assert tar_path.is_file()
+    # Test that the archived profile contains some basic items
+    profile_items = [
+        "cookies.sqlite",
+        "places.sqlite",
+        "webappsstore.sqlite",
+        "prefs.js",
+        "bookmarkbackups",
+        "cache2",
```
vringar, line 35: I don't see this file in my profile but since the tests are passing it must be some quirk in my setup I guess

vringar (approved):

test/test_profile.py
boolean5: Although I couldn't find this documented anywhere, after experimenting a little and reading the comments in https://stackoverflow.com/questions/25623705/selenium-retaining-firefox-cache-and-history-files, I concluded that when running Firefox with Selenium the browser cache is stored inside the profile, under the `cache2` directory, while when running Firefox normally it is stored under `~/.cache/mozilla/firefox/xxxxx/cache2`, where `xxxxx` is the same as the profile directory name.

Issue #789, opened 2020-11-11
Expand profile loading, saving, and recovery tests


englehardt:
We have relatively limited profile [test coverage](https://github.com/mozilla/OpenWPM/blob/master/test/test_profile.py). Once the tests are re-enabled (#383), we should define how we expect the `seed_tar` and `profile_tar` profiles to be handled for both stateful and stateless crawls and ensure our tests cover all conditions.

boolean5:
We should also update and reenable [`test_profile_saved_when_launch_crashes`](https://github.com/mozilla/OpenWPM/blob/ab01a2f6bdadcf75fe8c2878bc8d6610f5ed2a7a/test/test_profile.py#L59L80). This test uses the proxy to make the launch fail. Because the proxy is no longer supported, we need to find a different way to make the launch crash.

boolean5:
This is a list of things to test that I can think of. @vringar, @englehardt feel free to add more.

1. Test that for stateless crawls the profile specified by `seed_tar` is reloaded after each command sequence.
2. Test that the correct profile is loaded during crash recovery (a. the profile specified by `seed_tar` when running a stateless crawl, b. the profile specified by `recovery_tar`(i.e., the profile from the most recent page visit pre-crash) when running a stateful crawl)
3. Test dumping the profile as part of a command sequence (`DumpProfileCommand`)

vringar:
This condition is complicated and maybe worth testing or removing. https://github.com/mozilla/OpenWPM/blob/master/openwpm/browser_manager.py#L597

vringar:
This requires a test matrix.
1. operational mode: Stateful, Stateless
2. Seed_profile: Exists, Doesn't
3. Crash types: normal operation, crash during page visit, crash during launch, timeout during page visit

boolean5:
We should also test that [`load_profile`](https://github.com/mozilla/OpenWPM/blob/e20fc6a29e191e9bacc5922ff494a05a922c8587/openwpm/commands/profile_commands.py#L127):
 * leaves the tar file unmodified
 * can handle both compressed and uncompressed tar files

PR #924, opened 2021-05-12
Expand profile tests

boolean5:
This PR will add more profile tests and eventually close https://github.com/mozilla/OpenWPM/issues/789.

`test_profile_recovery` is a parametrized test that implements the test matrix outlined in https://github.com/mozilla/OpenWPM/issues/789#issuecomment-817924502.

I'll be adding a couple more small tests soon.

### DIFF
--- a/test/test_profile.py
+++ b/test/test_profile.py
@@ -7,7 +7,6 @@ from typing import Any
 import pytest
 
 from openwpm.command_sequence import CommandSequence
-from openwpm.commands.profile_commands import DumpProfileCommand
 from openwpm.commands.types import BaseCommand
 from openwpm.errors import CommandExecutionError, ProfileLoadError
 from openwpm.utilities import db_utils
@@ -164,6 +163,20 @@ class AssertConfigSetCommand(BaseCommand):
         assert result == self.expected_value
 
 
+def test_dump_profile_command(default_params, task_manager_creator):
+    """Test saving the browser profile using a command."""
+    manager_params, browser_params = default_params
+    manager_params.num_browsers = 1
+    manager, _ = task_manager_creator((manager_params, browser_params[:1]))
+    cs = CommandSequence(url=BASE_TEST_URL)
+    cs.get()
+    tar_path = manager_params.data_directory / "profile.tar.gz"
+    cs.dump_profile(tar_path, True)
+    manager.execute_command_sequence(cs)
+    manager.close()
+    assert tar_path.is_file()
+
+
 @pytest.mark.parametrize(
     "seed_tar",
     [None, Path("profile.tar.gz")],
@@ -211,7 +224,7 @@ def test_profile_recovery(
     cs.append_command(AssertConfigSetCommand("test_pref", expected_value))
     tar_directory = manager_params.data_directory / "browser_profile"
     tar_path = tar_directory / "profile.tar.gz"
-    cs.append_command(DumpProfileCommand(tar_path, True))
+    cs.dump_profile(tar_path, True)
     manager.execute_command_sequence(cs)
     manager.close()
 
--- a/test/test_profile.py
+++ b/test/test_profile.py
@@ -189,6 +189,21 @@ def test_load_tar_file(tmp_path):
     assert modified_time_before_load == tar_path.stat().st_mtime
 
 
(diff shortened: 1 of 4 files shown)


test/test_profile.py
```diff
@@ -159,3 +162,75 @@ def execute(
...
+@pytest.mark.parametrize(
+    "seed_tar",
+    [None, Path("profile.tar.gz")],
+    ids=["without_seed_tar", "with_seed_tar"],
+)
+@pytest.mark.parametrize("stateful", [True, False], ids=["stateful", "stateless"])
+@pytest.mark.parametrize(
+    "testcase",
+    ["on_normal_operation", "on_crash", "on_crash_during_launch", "on_timeout"],
+)
```
boolean5, line 176: I wonder if the `stateless-without_seed_tar` test cases are adding any value. @vringar, what do you think? If we decide that they are redundant, I can tweak the parametrization and we'll have 4 tests less.

test/test_profile.py
```diff
@@ -159,3 +163,114 @@ def execute(
...
+    manager_params.num_browsers = 1
+    browser_params[0].profile_archive_dir = (
+        manager_params.data_directory / "browser_profile"
+    )
+    # This will cause the browser launch to fail
+    browser_params[0].seed_tar = Path("/tmp/NOTREAL")
+    with pytest.raises(ProfileLoadError):
+        manager, _ = task_manager_creator((manager_params, browser_params[:1]))
+    tar_path = browser_params[0].profile_archive_dir / "profile.tar.gz"
+    assert not tar_path.is_file()
```
boolean5, line 204: @vringar, is this what you had in mind in https://github.com/mozilla/OpenWPM/issues/789#issuecomment-817917602 ?

boolean5:
This is ready now. @vringar, you can review whenever you have some time. Thanks!

Note that even though I moved the profile tests to their own Github Actions workflow job, they are pushing the total duration of the tests to 20 minutes. Are we ok with that or do we want to try and cut it down?
We could also consider splitting the profile tests to separate files, so that we can run them on different jobs.

test/test_profile.py
vringar: Sorry for not getting back to you in time.

I think what I wanted to express there, that if
1. We visit `http://a.com`
2. We crash visiting `b.com`
3. We visit `http://c.com`
all with a stateful crawl on the same browser, then `http://a.com` and `http://c.com` should both be in the `moz_places` DB

vringar:
I think we should have a script we run every release to redistribute the tests across the maximum amount of parallel instances based on their runtime.
The path based distribution is a hack and so I wouldn't spend too much time trying to fix it.

test/test_profile.py
vringar: I'd say they are redundant, so I'm going to adjust the parametrization and create a PR against your branch tomorrow.

test/test_profile.py
boolean5: That sounds like something that's tested by `test_profile_recovery`, no?
`during_init` is set to True [here](https://github.com/mozilla/OpenWPM/blob/b1dbee125a887b23724ec8ab0985cabcdf04b044/openwpm/task_manager.py#L186L188), if the task manager fails to launch the browsers the first time.
So from what I understand, when testing [this condition](https://github.com/mozilla/OpenWPM/blob/b1dbee125a887b23724ec8ab0985cabcdf04b044/openwpm/browser_manager.py#L606) we want to test that because the browsers didn't start successfully and did not execute any commands, no meaningful profile was created and thus no profile archive was saved.

vringar: You are of course correct!

test/test_profile.py
boolean5: Thanks for confirming! I could push this now since it's a pretty trivial change. Shall I?

boolean5: Done!

vringar: Thanks so much!

vringar (approved):

vringar:
Sorry for not reviewing this in time and thank you for still finishing this up!

boolean5:
No worries! Thanks for reviewing!

PR #929, opened 2021-05-18
Reenabling extension tests

### LINKED ISSUE #921: Fix extension tests and make them run on CI
Tests under `test/extension` are failing. We didn't notice this before because we forgot to include them in the Github Action [workflow jobs](https://github.com/mozilla/OpenWPM/blob/889b54d4295e1ad5c0da24c29a3b0ed746da33ba/.github/workflows/run-tests.yaml#L51).

`test_logging` is failing because on [this line](https://github.com/mozilla/OpenWPM/blob/889b54d4295e1ad5c0da24c29a3b0ed746da33ba/test/extension/test_logging.py#L22) it should be referencing the `log_list` variable instead of `log_line`.

`test_extension_startup_timeout` is probably failing because the JavaScript [code we use to introduce a delay](https://github.com/mozilla/OpenWPM/blob/889b54d4295e1ad5c0da24c29a3b0ed746da33ba/test/extension/test_startup_timeout.py#L13L20) before the instrumentation starts up seems to be running asynchronously, resulting in the extension starting on time. 

Finally, some of the existing test files (e.g., `test_extension.py`) might also belong semantically under the `extension` directory so we could look into moving them there. This way the duration of tests will be split more evenly over the workflow jobs.

vringar:
Closes #921

The second commit in this PR should be viewed separately.


vringar:
The doc builds are spuriously failing and I don't see a way to retrigger the build without pushing more changes.

PR #938, opened 2021-06-14
Update README.md

MaazBinMusa:
Since Dockerfile has CMD command, we dont need to specify in docker run


vringar:
Can you check that the default python in the docker container is python 3 and if it's not adjust the docker file as well.
I just want to make sure we aren't accidentally using python 2.

MaazBinMusa:
> Can you check that the default python in the docker container is python 3 and if it's not adjust the docker file as well.
> I just want to make sure we aren't accidentally using python 2.

Just checked. Its Python 3.9.4

PR #981, opened 2022-03-24
Refactor/extension unification

vringar:


vringar:
The unification is broken up into four pieces so far:
1. Move Extension/firefox into Extension
2. Move Extension/webext-instrumentation into Extension
3. Fix all linting errors in Extension/src
4. Applying linting and formatting to bundled/priviliged

PR #996, opened 2022-05-22
- update Docker command from `Running Measurements from inside the Container` in README

wolendranh:
**Explain problem:** 
it seems like command from description of OpenWPM was outdated. At leats it will not work with `demo.py` script. 

Old command:
```
    mkdir -p docker-volume && docker run -v $PWD/docker-volume:/opt/Desktop \
    -it openwpm
```
It mount host machine `$PWD/docker-volume` dir to `/opt/Desktop`  in container, which does not contain `demo.py` execution artifacts (logs, Sqlite db fiels etc.)

**Solution:**
I suggest to change it to 

```
$PWD/docker-volume:/opt/OpenWPM/datadir
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -274,7 +274,7 @@ X-server. You can do this by running: `xhost +local:docker`
 Then you can run the demo script using:
 
 ```bash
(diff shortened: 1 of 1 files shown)


vringar:
Thanks for this PR!

PR #1016, opened 2022-11-22
Add optional tranco list to demo script

englehardt:
This adds the `--tranco` option to the demo script, which will visit the top 5 sites as ranked by that list. I've decided to make it optional as the list may contain NSFW sites and I didn't want that to be the default path. I think they're unlikely to appear in the top 5, but wanted to be safe.

@vringar I'm curious for your thoughts on this. We're adding an additional dependency for an optional demo feature, but IMO a top-sites list is almost certainly needed by most users and folks may not be aware of the Tranco list / not know about the library.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+# Tranco list cache directory
+.tranco/
+
 # Docker volume
 docker-volume/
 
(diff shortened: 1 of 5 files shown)


vringar:
I'm not really that concerned with adding a dependecy and making people aware of tranco is definitely a nice benefit.
The only thing I'm wondering, if it makes sense to select the tranco top 100 and use 3 or 4 browsers. Just to demonstrate the parallel crawling capabilities of OpenWPM.

englehardt:
@vringar I was conservative, but bumped both.

PR #1019, opened 2023-01-11
Quaterly Mega PR

vringar:
This PR does the following:

- Switch from conda to [mamba](https://github.com/mamba-org/mamba)
- Update dependencies and Firefox to 108.0.2
- Refactor audio_fingerprinting so that local tests pass


englehardt:
@vringar Are you still planning to break this up a bit?

I installed mamba without any issues, but started receiving the following error when running `install.sh` after doing so:

```
❯ ./install.sh
Creating / Overwriting openwpm conda environment.
Collecting package metadata (repodata.json): ...working... done
Solving environment: ...working... failed

ResolvePackageNotFound: 
  - pyvirtualdisplay=3.0
  - geckodriver=0.32.0
```

I was able to track down the issue by running the following:

```
❯ conda search -c conda-forge geckodriver
Loading channels: done
No match found for: geckodriver. Search: *geckodriver*

PackagesNotFoundError: The following packages are not available from current channels:

  - geckodriver

Current channels:

  - https://conda.anaconda.org/conda-forge/osx-arm64
  - https://conda.anaconda.org/conda-forge/noarch
```

It looks like [geckodriver](https://anaconda.org/conda-forge/geckodriver) and [pyvirtualdisplay v3](https://anaconda.org/conda-forge/pyvirtualdisplay) are `osx-64` only and not `osx-arm64` (I'm on an M1 Macbook Pro).  I'm not sure why this issue popped up after installing mamba?

I was able to solve that by running `conda config --env --set subdir osx-64`, but I also don't think this is the right fix since afaik this is a global change if run in the base env before install. I pushed a more targeted fix in https://github.com/openwpm/OpenWPM/pull/1019/commits/6dc62bd431d0c0328556a3b4d280cffe089470b3

vringar:
I've split this PR into #1022 #1025  and #1026 .
However, I'd like to leave this PR open and after the other ones are merged, check that there is nothing left in this PR that isn't covered by the other ones.
@englehardt Does that make sense?

englehardt:
SGTM! r+ on the other three!

PR #1021, opened 2023-01-12
[WIP] Move dns_instrument from onCompleted to onHeadersReceived

### LINKED ISSUE #1020: DNS Instrument only records the final request in a redirect chain
Originally reported in #989, but I noticed it myself while playing around with some crawl data recently. The problem stems from the listener we're using to record DNS entries.

https://github.com/openwpm/OpenWPM/blob/ab5fac2d7b81f9b05704bd7460b3b20aa8f23f7b/Extension/src/background/dns-instrument.ts#L29-L44

The `onCompleted` event is only fired [after redirects are resolved](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/webRequest). Instead, we want to use an event that is part of the redirect loop. e.g., `onHeadersReceived`. I tested this and it works well, but I still need to add some tests.

englehardt:
Fixes #1020 WIP. I haven't tested this much.

### DIFF
--- a/Extension/src/background/dns-instrument.ts
+++ b/Extension/src/background/dns-instrument.ts
@@ -2,10 +2,11 @@ import { PendingResponse } from "../lib/pending-response";
 import { DnsResolved } from "../schema";
 import { allTypes } from "./http-instrument";
 import RequestFilter = browser.webRequest.RequestFilter;
+import { WebRequestOnHeadersReceivedDetails } from "../types/browser-web-request-event-details";
 
 export class DnsInstrument {
   private readonly dataReceiver;
-  private onCompleteListener;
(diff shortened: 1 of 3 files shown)


vringar:
Good Catch! Does this also resolve the issue where DNS requests for domains that don't resolved aren't captured?

englehardt:
I'm not sure. I'll need to write a test page for both of these conditions.

vringar:
Superseded by #1158 (merged in v0.34.0), which moves dns_instrument from onCompleted to onHeadersReceived. See also #1164 (failed DNS via onErrorOccurred) and #1156 (DNS errors excluded from consecutive failure limit).

PR #1025, opened 2023-01-18
fix(audio_fingerprinting): trigger audio fingerprint with click

vringar:
This is required as AudioContext doesn't work without user interaction

### DIFF
--- a/test/test_extension.py
+++ b/test/test_extension.py
@@ -1,12 +1,19 @@
 import os
+import time
 from datetime import datetime
 from pathlib import Path
 from sqlite3 import Row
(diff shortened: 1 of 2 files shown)


englehardt:
Non-blocking: do you know if this is an AudioContext spec change where the API actually is gated behind user interaction, or whether there is simply a race condition here?

englehardt (approved):

vringar:
There was an error message in the console saying smt like Audio context is unavailable without user interaction. And with some googling I found people also reported the same for Chrome. So maybe audio fingerprinting is just dead now?

PR #1028, opened 2023-01-18
Unify tool configuration in pyproject.toml

### LINKED ISSUE #1027: Use pyproject.toml and fix coverage
As described [in the 4.0.0 release notes](https://pytest-cov.readthedocs.io/en/latest/changelog.html#id1) pytest-cov removed support for multiprocessing.
Instead of adding an additional config file, it seems seems better to migrate from the old setup.cfg to the new pyproject.toml and configure coverage there according to [this documentation](https://coverage.readthedocs.io/en/latest/config.html).

By migrating from the setup.cfg to pyproject.toml for all the other configuration we keep the amount of config files constant.

Both [mypy](https://mypy.readthedocs.io/en/stable/config_file.html) and [isort](https://pycqa.github.io/isort/docs/configuration/config_files) support and recommend pyproject.toml.

vringar:
Closes #1027


englehardt (approved):
Thanks! I wasn't aware of this file type before. Glad to see it unified.

PR #1034, opened 2023-04-12
Update to Firefox 112

englehardt:
Bump Firefox version to 112


vringar:
Updated directly to 114

PR #1039, opened 2023-06-18
OpenWPM StorageWatchdog complete

gridl0ck:
Introducing the StorageWatchdog module for OpenWPM, a powerful tool designed to enhance your web scraping experience. This module offers a range of features that enable you to efficiently manage temporary files, monitor their size, and effectively handle browser profiles, all without compromising performance.

With the StorageWatchdog module, you gain the ability to easily redirect temporary files to a directory of your choice. This functionality ensures that any temporary files generated during your web scraping activities are conveniently stored in a location that suits your needs. Whether you prefer a specific directory for organization or have limited storage space concerns, this feature allows you to maintain control over the storage location of these files effortlessly.

In addition to redirecting the temporary files, the StorageWatchdog module offers a monitoring system for their size. This capability enables you to keep track of the growth and consumption of disk space by the temporary files and browser profiles. The watchdog performs checks at 5 minute intervals and checks the size of the current browser profile to determine whether or not that specific browser needs to be reset to clear space.

This addition was created as a result of my capstone team running into issues with longer crawls running out of storage on our smaller infrastructure. Because we only needed the data from the generated database (more than most people need in the first place), these other random artifacts were simply taking up space on our system.

The usage is fairly simple and a *demo_watchdog.py* has been supplied to demonstrate its functionality and usage. Simply enable the watchdog as you would any other watchdog, configure the size in bytes for the check to monitor (this is located in the periodic_check function in storage_watchdog.py) and optionally specify a tmp file location (this defaults to /tmp as it has before) using the new Browser Paramater, *tmp_profile_dir*.

I hope you accept these changes and agree that this simple addition could be helpful to others!


vringar (commented):
Hey,
thank you for this PR. I see that this is a useful feature especially when running natively on machines that are not under your control.
And while this looks good on first pass, I'd need to spend more time on a thorough review which I don't have atm.
I'll try to get back to you as soon as my schedule allows

openwpm/config.py
```diff
@@ -99,6 +99,9 @@ class BrowserParams(DataClassJsonMixin):
     profile_archive_dir: Optional[Path] = field(
         default=None, metadata=DCJConfig(encoder=path_to_str, decoder=str_to_path)
     )
+    
+    tmp_profile_dir: str = "/tmp"
```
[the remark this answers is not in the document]
gridl0ck: Yes sir,

I can make that change and push in a separate pull request. Additionally, all the temporary files that gets created when a browser is spun up (the profile, the configuration files etc.) are all stored in the tmp_profile_dir which was formerly "/tmp", which is why it has that as the default value in the BrowserParams class.

gridl0ck:
If there are any other changes requested, feel free to comment! I really enjoyed working on this project and would love to help continue to make it better!

openwpm/utilities/storage_watchdog.py
vringar, line 1: As a general rule, I try to only get new code with type annotations, so if you wouldn't mind adding them, that would be great.
However, I can also do that later.

openwpm/config.py
```diff
@@ -99,6 +100,10 @@ class BrowserParams(DataClassJsonMixin):
     profile_archive_dir: Optional[Path] = field(
         default=None, metadata=DCJConfig(encoder=path_to_str, decoder=str_to_path)
     )
+    
+    # tmp_profile_dir: str = "/tmp"
```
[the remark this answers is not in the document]
vringar: Also maybe naming the option just `profile_dir` and a docstring explaining that the `profile_archive_dir` gets unpacked into the `profile_dir` and then Firefox writes into the `profile_dir`?

openwpm/config.py
```diff
@@ -133,6 +138,12 @@ class ManagerParams(DataClassJsonMixin):
     """A watchdog that tries to ensure that no Firefox instance takes up too much memory.
     It is mostly useful for long running cloud crawls"""
     process_watchdog: bool = False
+
+    """A watchdog that serves as a happy medium between killing a browser after each 
+    crawl and allowing the application to still perform quickly. Used as a way to save space 
+    in a limited environment with minimal detriment to speed."""
+    storage_watchdog_enable: bool = False
```
vringar, line 145: What if this was an `Optional[int]`? So if you set a max size in byte, then the watchdog gets activated?

openwpm/utilities/storage_watchdog.py
```diff
@@ -0,0 +1,97 @@
...
+def periodic_check(profile_path, obj):
+    logger = logging.getLogger("openwpm")
+    # 1073741824: # 1GB
+    # 20971520: # 20MB - for testing purposes
+    # 52428800: # 50MB
+    # 73400320: # 70MB
+    # 104857600: 100MB - IDEAL for 10+ browsers
+    
+    # Max Size before a restart expressed in bytes
+    MAX_DIRSIZE = 104857600
```
vringar, line 80: That way you wouldn't have to hardcode this

vringar (changes requested):
Thanks for following up on this and sorry for letting it linger.

environment.yaml
```diff
@@ -43,7 +43,9 @@ dependencies:
   - jsonschema==4.18.6
   - plyvel==1.5.0
   - tranco==0.6
+
   - types-pyyaml==6.0.12.11
   - types-redis==4.6.0.3
   - types-tabulate==0.9.0.3
+  - watchdog==3.0.0
```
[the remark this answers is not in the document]
gridl0ck: Resolved. I'll comment on each one as I resolve them but I will make a final push with all of the corrections and note that as well.

openwpm/utilities/storage_watchdog.py
gridl0ck: I will add the annotations to the primitive types as type annotations are new to me, but seem to make sense.

openwpm/config.py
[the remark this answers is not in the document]
gridl0ck: I simply added the docstring and made it use the tempfile module's "gettempdir" function. I did not rename the variable for simplicity's sake.

openwpm/config.py
gridl0ck: I simply modeled it after the other watchdogs. Made the change now and added some error checking in the object's "run" function.

openwpm/utilities/storage_watchdog.py
gridl0ck: I put this in the docstring for the "storage_watchdog_enable" variable in the ManagerParam class in config.py

gridl0ck (commented):
See current implementation but marking this as resolved.

PR #1075, opened 2023-12-06
fix #1068 : experimented changes

### LINKED ISSUE #1068: Freezing During Shutdown: Exception Handling in finalize_visit_id Causes Controller to Halt
Hi,

When the storage controller receives a shutdown signal, it waits for all tasks associated with each visit ID. However, during the shutdown process, if an exception is thrown while processing records in finalize_visit_id, the controller freezes and doesn't continue. 
Why does it freeze? Are there any ways to overcome this issue? In the case of an exception, shouldn't it log the error and continue?

Exception (KeyError) is thrown at

https://github.com/openwpm/OpenWPM/blob/553b1f3fb678bf8e11b437e73fa34cd26faed492/openwpm/storage/storage_controller.py#L219

which is being called from

https://github.com/openwpm/OpenWPM/blob/553b1f3fb678bf8e11b437e73fa34cd26faed492/openwpm/storage/storage_controller.py#L258

and at the end causes an infinite loop at

https://github.com/openwpm/OpenWPM/blob/553b1f3fb678bf8e11b437e73fa34cd26faed492/openwpm/task_manager.py#L394-L402

because it is waiting for the visit_id to appear in the completion queue.

in my logs i have already that the visit_id is awaited
```storage_controller   - INFO     - Awaited all tasks for visit_id 5095731113063583 while finalizing```
which indicates, it has already been finalized

https://github.com/openwpm/OpenWPM/blob/553b1f3fb678bf8e11b437e73fa34cd26faed492/openwpm/storage/storage_controller.py#L220-L222

but 3 lines below i have the above exception which says that it didn't find the visit id which makes sense since it is finalized 3 lines up and deleted afterwards

### PR DESCRIPTION
**Pull request by MohammadMahdiJavid**
Changes Applied for experimentations

### DIFF
--- a/openwpm/storage/storage_controller.py
+++ b/openwpm/storage/storage_controller.py
@@ -214,9 +214,12 @@ class StorageController:
             return None
 
         self.logger.info("Awaiting all tasks for visit_id %d", visit_id)
-        for task in self.store_record_tasks[visit_id]:
-            await task
+
+        store_record_tasks = self.store_record_tasks[visit_id]
         del self.store_record_tasks[visit_id]
+        for task in store_record_tasks:
+            await task
+
         self.logger.debug(
             "Awaited all tasks for visit_id %d while finalizing", visit_id
         )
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: openwpm/storage/storage_controller.py [hunk 1]
```diff
@@ -535,6 +538,7 @@ def get_most_recent_status(self) -> int:
 
         # Check last status signal
         if (time.time() - self._last_status_received) > STATUS_TIMEOUT:
+            self._last_status = None
```
vringar, line 541: I don't understand the motivation for resetting this. What was your reasoning?

PR #1089, opened 2024-05-20
mamba --force argument removed of  `install.sh` script

AlbertoFDR:
Remove the argument `--force` from `install.sh`.


vringar:
I have to admit that I tried to change my local mamba install to use micromamba and that broke everything.
I currently do not have the time or motivation to fix this, but seeing that this works with the latest mamba in CI, I'm just going to merge it.

Issue #1090, opened 2024-06-07
XPI files not deleted from /tmp


ndanner-wesleyancs:
Preface:  I don't think this is an OpenWPM bug, but I'm wondering if anybody has suggestions on how to work around it.

Platform:  OpenWPM v0.28, Selenium 4.21.0, Geckodriver 0.34.0.

When the extension is added to a deployed Firefox instance in [`deploy_firefox.py`, line 161](https://github.com/openwpm/OpenWPM/blob/f74e79fca69f8824ffd1c69f1b349890e97b4217/openwpm/deploy_browsers/deploy_firefox.py#L161), a copy of the XPI file is made in `/tmp`, presumably by either Selenium or `geckodriver`.  The problem is that it is not deleted.

That's a problem, since when I was crawling a large number of sites, I ended up with enough copies of the XPI file in `/tmp` that it filled up.

As I said, I don't think this is the fault of OpenWPM.  My understanding is that Selenium's `WebDriver.quit()` is supposed to delete temporary files.  It does appear to delete profile directories, just not XPI files.

Has anyone else seen this?  And if so, any suggested workarounds?

vringar:
Hey, this sounds like smt that should be done in the shutdown code of the browser manager 
Please be aware however, that the Firefox version used by OpenWPM is severely out of date and before running any large scale crawls you should probably take the steps described under docs/Release-Checklist to look closer to normal web traffic

ndanner-wesleyancs:
I haven't been able to find anything in the Selenium docs that would let me get the name of the temporary file that is created by `install_addon` in order to delete it, though I won't say that I've done a truly exhaustive search.

This issue seems likely related to https://github.com/SeleniumHQ/selenium/issues/10841, and in the comments therein, one of the posters relates it to https://github.com/mozilla/geckodriver/issues/299.

ndanner-wesleyancs:
Thanks for pointing out the issue with the FF version; I'll be sure to take that into account.

ndanner-wesleyancs:
Just to be sure that this isn't caused by something I'm doing in my crawl script, here is what I think is a pretty minimal crawl script that demonstrates the undesired behavior:
```[python]
"""
MWE to demonstrate that a copy of the OpenWPM extension XPI file is left behind
in /tmp each time a fresh browser is invoked.

N. Danner
"""

import pathlib

from openwpm.command_sequence import CommandSequence
[...]

PR #1093, opened 2024-07-03
Set custom temporary directory used by FF and delete it

ndanner-wesleyancs:
This pull request modifies OpenWPM as follows:
- When a browser is launched, a custom temporary directory is created.
- Before `geckodriver` is started, the TMPDIR environment variable is set in the environment used by `geckodriver`.
- When the browser terminates, the custom temporary directory is deleted.

The proximate rationale for doing this is that when `geckodriver` adds the OpenWPM extension to FF, it makes a copy of the XPI file in its temporary directory.  Unfortunately, it doesn't delete it.  That means one copy of the XPI file for each browser that is launched is left behind in `/tmp`.  On a big stateless crawl, that's a lot of copies.  We can't find out the name of the temporary file that `geckodriver` creates to delete it ourselves, and we don't want to just delete all XPI files in `/tmp` because we don't actually know to whom the all belong.  This approach seems to resolve these problems.

This patch is based on v.0.28.0.

See #1090.

### DIFF
--- a/openwpm/browser_manager.py
+++ b/openwpm/browser_manager.py
@@ -358,22 +358,6 @@ class BrowserManagerHandle:
                     )
                     return
 
-                # Delete the temporary directory used by geckodriver.
-                try:
-                    self.logger.debug(
-                        "BROWSER %i: deleting temp dir %s" %
-                        (self.browser_params.browser_id, 
-                            self.browser_params.tmpdir)
-                    )
-                    shutil.rmtree(self.browser_params.tmpdir)
-                    self.browser_params.tmpdir = None
-                except Exception as e:
-                    self.logger.warn(
-                        "BROWSER %i: failed to delete temp dir %s: %s" %
-                        (self.browser_params.browser_id, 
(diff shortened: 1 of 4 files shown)


vringar (changes requested):
Hey, 
thanks for you contribution.
Could you create a test where you assert that the original TMPDIR doesn't contain any .xpi and also address the mypy issues?
Or is this contribution done from your side and if I want those things I need to do them myself?

ndanner-wesleyancs:
I'll try to work on these in the next couple of weeks.

ndanner-wesleyancs:
I have fixed the `mypy` errors.

ndanner-wesleyancs:
I'm a little perplexed about your suggested test.  The original TMPDIR (typically `/tmp`) could very well have XPI files in it, for example from some other independent process.  I'm not sure why we would want to guard against that.  If you meant the new TMPDIR, then since it is created by `mkdttmp`, it is necessarily empty, isn't it?

PR #1095, opened 2024-07-12
Update to Firefox 128

palmeida:
Firefox was updated to [version 128](https://hg.mozilla.org/releases/mozilla-release/rev/FIREFOX_128_0_RELEASE).

As a side note, `repin.sh` was still using mamba's `--force` switch that has been removed, so that was fixed.


vringar:
Thank you so much for this PR!
I hope there is no breakage in the WebExtension but I think most of the tricky things are already broken for a while, so tests should be passing

vringar (approved):
Looks exactly what I expected it to look like!
Good job!

PR #1105, opened 2024-08-09
Get demo.py running on MacOS

englehardt:
A few notes:
* I was getting errors from mamba about not wanting to overwrite an existing environment. From the docs, I think we intend to, so I've added `--force` which will overwrite. I'm not sure if this is a platform difference?
* The firefox binary has changed names on MacOS
* I've migrated `storage_watchdog.py` over to python native code. Let's avoid subprocess calls out to system utilities as much as possible. MacOS has independent implementations of the OS utils, often with different options. I was tempted to install `gnu-coreutils` to fix this, but I'd rather not add another dependency when we can easily do the same calculation in python.

### DIFF
--- a/environment.yaml
+++ b/environment.yaml
@@ -9,32 +9,32 @@ dependencies:
 - dill=0.3.8
 - easyprocess=1.1
 - gcsfs=2024.6.1
-- geckodriver=0.34.0
+- geckodriver=0.35.0
 - ipython=8.26.0
 - isort=5.13.2
 - leveldb=1.23
 - multiprocess=0.70.16
-- mypy=1.10.1
-- nodejs=22.4.0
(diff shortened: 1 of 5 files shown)


englehardt:
The tests were failing because of the `--force` option I added. I see in https://github.com/openwpm/OpenWPM/issues/1088 we removed it intentionally. I probably just have an outdated version of mamba.

But it looking it up, it seems the mamba solver is now used by conda: https://github.com/conda-forge/miniforge?tab=readme-ov-file#whats-the-difference-between-mambaforge-and-miniforge. So we should update https://github.com/openwpm/OpenWPM?tab=readme-ov-file#pre-requisites and the install files to revert back to the latest miniforge.

PR #1106, opened 2024-09-30
Update to Firefox 130

palmeida:
Firefox was updated to [version 130.0.1](https://hg.mozilla.org/releases/mozilla-release/rev/FIREFOX_130_0_1_RELEASE).


vringar:
Thank you for this! Going to wait for the tests on master to complete and then tag a release.

palmeida:
No problem! Like I said before, when the process goes smoothly (the first time there were still issues with mamba parameters, etc), it's no trouble at all. I'll do it anyway in my local repository when I run OpenWPM audits, but I can keep making pull requests regularly. Maybe once a month?

vringar:
Yeah, i think the release cycle is every four weeks, so if you could keep this up that would be phenomenal!

PR #1109, opened 2025-01-16
Update to Firefox 134 and migrate back to conda

englehardt:
As mentioned in https://github.com/openwpm/OpenWPM/pull/1105, mamba has been merged upstream into conda. Looks like as of 2025 they've fully deprecated it and stopped building the mamba binaries.

I replaced all references to mamba with conda and things "just work" both locally and on the CI tests, so I think we're all good.


vringar:
Added back the ability to rerun our conda modifying scripts (repin and install) without having to first delete the openwpm environment
Everything else looks great

vringar (approved):
Added back the ability to rerun our conda modifying scripts (repin and install) without having to first delete the openwpm environment
Everything else looks great

PR #1112, opened 2025-04-20
Bump to Firefox 137

englehardt:


vringar:
Hey :)
Seems like smt broke in the profile handling 
Might have some time to look into those tomorrow, but idk how broken my current conda install is

PR #1115, opened 2025-12-23
Feature/add frontier link fallback logic to crawler

parthav1:
Refactored and removed redundant checks as well.


parthav1:
Opened accidentally against upstream instead of fork.

PR #1138, opened 2026-02-23
fix(cleanup): remove XPI temp files from /tmp on shutdown

### LINKED ISSUE #1090: XPI files not deleted from /tmp

vringar:
## Summary
- Track temporary XPI files created by geckodriver during `install_addon()` and clean them up on browser shutdown/restart
- Move tempdir cleanup outside the success block so it runs even on failed browser launches
- Use `extend()` instead of `=` when collecting tmp_files to avoid leaking files from retried spawn attempts

Closes #1090

## Changes
- `openwpm/deploy_browsers/deploy_firefox.py`: Detect new XPI files via glob before/after `install_addon()`, report them via status queue
- `openwpm/browser_manager.py`: Track `tmp_files` on `BrowserManagerHandle`, clean up in `shutdown_browser()` and `restart_browser_manager()`

## Test plan
- [ ] Verify XPI files are cleaned up after browser shutdown
- [ ] Verify temp profile directories are cleaned up even on failed launches
- [ ] Run full test suite to check for regressions

### DIFF
--- a/openwpm/browser_manager.py
+++ b/openwpm/browser_manager.py
@@ -12,7 +12,7 @@ import time
 import traceback
 from pathlib import Path
 from queue import Empty as EmptyQueue
-from typing import TYPE_CHECKING, Any, Dict, Optional, Tuple, Type, Union
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Type, Union
 
 import psutil
 from multiprocess import Queue
@@ -91,6 +91,8 @@ class BrowserManagerHandle:
         """indicates if the BrowserManager is new (to optimize restarts)"""
         self.restart_required: bool = False
         """indicates if the browser should be restarted"""
+        self.tmp_files: List[Path] = []
+        """temporary files (e.g. XPI copies) that should be cleaned up on shutdown"""
 
         self.current_timeout: Optional[int] = None
         """timeout of the current command"""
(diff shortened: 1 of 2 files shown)


vringar:
Closing in favor of #1104 — the racy tracking approach doesn't align with OpenWPM's robustness-first philosophy.

PR #1151, opened 2026-03-02
feat(ci): dynamic test duration tracking for pytest-split

vringar:
## Summary

- Add pytest hooks in `test/conftest.py` to record per-test wall-clock durations, gated on `GROUP` env var (CI only)
- Update `scripts/ci.sh` to prefer cached durations over the committed `.test_durations` baseline
- Add cache restore/save steps and a `consolidate-durations` job to `.github/workflows/run-tests.yaml` that merges per-group artifacts into a single cached file

## How it works

**Per-run data flow:**
```
Restore cache → ci.sh picks cached or committed file → pytest-split distributes tests
    → conftest hook records durations → upload per-group artifacts
    → consolidation job merges all groups → save to cache
```

**Fallback chain** (priority order):
1. Cached durations from the most recent run on the same branch
2. Cached durations from the most recent run on `master`
3. Committed `.test_durations` file (unchanged, serves as baseline)

## Test plan

- [x] Run `pytest test/test_dataclass_validations.py -v` with `GROUP=1` — verified `.test_durations_group_1` is created with correct node IDs matching `.test_durations` format
- [ ] Verify CI workflow: cache restore (miss on first run), artifact upload, consolidation, cache save
- [ ] Push again to verify cache hit and that splitting uses cached durations

### DIFF
--- a/.github/workflows/run-tests.yaml
+++ b/.github/workflows/run-tests.yaml
@@ -94,6 +94,7 @@ jobs:
           name: test-durations-${{ matrix.group }}
           path: .test_durations_group_${{ matrix.group }}
           if-no-files-found: ignore
+          include-hidden-files: true
       - name: Upload coverage artifact
         uses: actions/upload-artifact@v4
         if: always()
--- a/.github/workflows/run-tests.yaml
+++ b/.github/workflows/run-tests.yaml
@@ -34,6 +34,14 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: ./.github/actions/setup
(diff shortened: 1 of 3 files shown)


vringar:
Superseded by #1155 (feat/dynamic-ci-test-durations-v2), a strict superset: identical change plus a `CI`-env guard on the durations-file write (`if os.environ.get("CI") and group and _test_durations`) so it only writes in CI, not on local runs, plus the `.gitignore` entries. Nothing unique here. Closing in favor of #1155.

PR #1153, opened 2026-04-04
refactor(test): dynamic ports with fixture injection

vringar:
## Summary
- Use OS-assigned ports (`port=0`) instead of hardcoded `localhost:8000` to enable parallel test execution
- Inject `ServerUrls` fixture via pytest, removing global state from test infrastructure
- Fix `test_get_with_popup_blocking` NameError (missing `server` fixture)
- Replace `os.chdir()` with `SimpleHTTPRequestHandler(directory=)` parameter
- Fix `FileNotFoundError` in `dump_page_source` tests by using absolute paths

## Drive-by fixes folded in
- `test_xvfb_browser.py` and `test_callback.py` previously constructed URLs as `BASE_TEST_URL + "/test_pages/simple_a.html"`, which resolved to `http://localhost:8000/test_pages/test_pages/simple_a.html` (404). Neither test asserts on fetch correctness, so this went unnoticed. Now corrected to `server.base + "/simple_a.html"`.

## Known follow-up
- `test_cache_hits_recorded` still has ~1/8 flake rate where resources expected to be cached come back with `is_cached=0`. This predates this PR — an earlier draft of this branch added an explicit `Cache-Control: max-age=3600` header to the test server thinking it would help, but measurement showed it actually made the test *more* flaky (3/8 pass vs 7/8 without). The header was removed in fdb9c0a. The residual flakiness is inherent to Firefox's heuristic cache behavior on localhost and is out of scope here.

Supersedes #1136.

### DIFF
--- a/openwpm/utilities/db_utils.py
+++ b/openwpm/utilities/db_utils.py
@@ -1,14 +1,36 @@
 import sqlite3
 from collections.abc import Iterable
 from pathlib import Path
-from typing import Any, AnyStr, Iterator, List, Optional, Tuple, Union
+from typing import (
+    Any,
+    AnyStr,
+    Iterator,
+    List,
+    Literal,
+    Optional,
+    Tuple,
+    Union,
+    overload,
+)
 
 import plyvel
 
 
+@overload
+def query_db(
+    db: Path, query: str, params: Any = ..., as_tuple: Literal[False] = False
+) -> List[sqlite3.Row]: ...
+@overload
+def query_db(
+    db: Path, query: str, params: Any = ..., *, as_tuple: Literal[True]
+) -> List[Tuple[Any, ...]]: ...
+@overload
+def query_db(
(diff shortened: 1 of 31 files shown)


vringar:
Closing as redundant now that #1136 has merged to master.

- The dynamic-ports + `ServerUrls` fixture-injection work here landed via #1136.
- The type-checker migration and the `query_db` / `get_javascript_entries` overloads are tracked separately in #1170.
- All the drive-by fixes folded in here are already present in #1136: the `test_xvfb_browser.py` / `test_callback.py` double-`/test_pages/` 404 fix, the `test_get_with_popup_blocking` NameError, the `dump_page_source` path handling, and the Cache-Control header removal.

The only item not carried over is the `os.chdir()` → `SimpleHTTPRequestHandler(directory=)` cleanup; #1136 keeps `os.chdir()` (which works fine — test parallelism is process-level). That tidy-up can be a small standalone follow-up if wanted.

PR #1165, opened 2026-04-20
feat(scripts): sync pre-commit linter versions with conda in update.py

vringar:
## Summary

- Adds `sync_precommit_linter_versions()` to `scripts/update.py` that reads pinned versions of black, isort, and mypy from `environment.yaml` and updates the corresponding `rev:` entries in `.pre-commit-config.yaml`
- This runs automatically as part of `python scripts/update.py`, right after repinning the conda environment
- Updates `docs/Release-Checklist.md` to document the new linter drift verification step and fixes step numbering

## Motivation

black, isort, and mypy versions can drift between the conda environment (where they run day-to-day) and `.pre-commit-config.yaml` (where they run in CI/pre-commit hooks). Version drift causes formatting disagreements and false positives/negatives. This change ensures the two stay in sync automatically during the release update process.

## Test plan

- [x] `pre-commit run --all-files` passes
- [ ] Run `python scripts/update.py` in a full conda environment and verify linter versions are synced

### DIFF
--- a/docs/Release-Checklist.md
+++ b/docs/Release-Checklist.md
@@ -4,15 +4,17 @@ We aim to release a new version of OpenWPM with each new Firefox release (~1 rel
 
 1. Run `python scripts/update.py` — this will:
     - Repin the conda environment
+    - Sync linter versions in `.pre-commit-config.yaml` to match `environment.yaml` (black, isort, mypy)
     - Bump all npm dependencies to their latest compatible versions
     - Rebuild the extension
     - Automatically detect and apply any new Firefox release to `scripts/install-firefox.sh`
-2. Check if we can unpin our python version (this requires Linux to test) See <https://github.com/openwpm/OpenWPM/issues/1111>
-3. Run `./scripts/install-firefox.sh` to install the updated Firefox locally and verify the test suite passes.
-4. Increment the version number in [VERSION](../VERSION)
-5. Add a summary of changes since the last version to [CHANGELOG](../CHANGELOG.md)
-6. Squash and merge the release PR to master.
-7. Publish a new release from <https://github.com/openwpm/OpenWPM/releases>:
+2. Verify no linter drift warnings were printed by `update.py`. If any `WARNING` lines mention a linter that could not be synced, resolve the mismatch manually.
+3. Check if we can unpin our python version (this requires Linux to test) See <https://github.com/openwpm/OpenWPM/issues/1111>
+4. Run `./scripts/install-firefox.sh` to install the updated Firefox locally and verify the test suite passes.
+5. Increment the version number in [VERSION](../VERSION)
+6. Add a summary of changes since the last version to [CHANGELOG](../CHANGELOG.md)
+7. Squash and merge the release PR to master.
+8. Publish a new release from <https://github.com/openwpm/OpenWPM/releases>:
     1. Click "Draft a new release".
     2. Enter the "Tag version" and "Release title" as `vX.X.X`.
     3. In the description:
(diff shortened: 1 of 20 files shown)


scripts/update.py
```diff
@@ -33,6 +41,80 @@ def conda_run(*cmd: str, cwd: Path = ROOT) -> None:
...
+    for line in env_yaml.read_text().splitlines():
+        stripped = line.strip()
+        if not stripped.startswith("- ") or stripped.startswith("- pip"):
+            continue
+        spec = stripped[2:]
+        if "==" in spec or "=" not in spec:
+            continue
+        name, _, version = spec.partition("=")
+        versions[name] = version
+    return versions
```
vringar: It feels like we should be able to call out to conda and use it to get this information?

scripts/update.py
```diff
@@ -33,6 +41,80 @@ def conda_run(*cmd: str, cwd: Path = ROOT) -> None:
...
+    updated = False
+    for pkg, (repo_url, prefix) in _LINTER_MAP.items():
+        if pkg not in conda_versions:
+            print(
+                f"WARNING: {pkg} not found in environment.yaml, skipping",
+                file=sys.stderr,
+            )
+            continue
+
+        target_rev = f"{prefix}{conda_versions[pkg]}"
```
vringar, line 103: Same here tbh. Why are we parsing the file ourselves.

scripts/update.py
vringar: Addressed in 27e22a3 — `_query_conda_versions` now shells out to `conda list -n <env> --json` and parses the result. The regex-over-environment.yaml approach is gone.

scripts/update.py
vringar: Looked into this — pre-commit doesn't expose a public API for "set the rev for repo X to version Y":

- `pre-commit autoupdate` only updates to the latest tag (or `HEAD` with `--bleeding-edge`). No way to target a specific version.
- Pre-commit's own internal implementation (`pre_commit/commands/autoupdate.py`) uses regex (`REV_LINE_RE = r'^(\s+)rev:(\s*)([\'"]?)([^\s#]+)(.*)(\r?\n)$'`) to surgically rewrite `rev:` lines while preserving comments/formatting — exactly the same approach we're using. So they push the YAML burden onto themselves too.
- There is a private `_write_new_config(path, rev_infos)` we could call, but it's underscore-prefixed and depends on `pre_commit.clientlib` + `pre_commit.yaml` internals. Reaching into private API for ~10 lines of regex feels like the wrong trade-off.

Decided to keep the regex. Conda side now uses `conda list --json` instead of regex-parsing environment.yaml (27e22a3), and the function fails loud (`RuntimeError`) on missing linters or missing hooks. Unit tests added in 03d8c1b (`test/test_update_script.py`).

docs/Release-Checklist.md
```diff
@@ -4,10 +4,11 @@ We aim to release a new version of OpenWPM with each new Firefox release (~1 rel
 
 1. Run `python scripts/update.py` — this will:
     - Repin the conda environment
+    - Sync linter versions in `.pre-commit-config.yaml` to match `environment.yaml` (black, isort, mypy)
     - Bump all npm dependencies to their latest compatible versions
     - Rebuild the extension
     - Automatically detect and apply any new Firefox release to `scripts/install-firefox.sh`
-2. Check if we can unpin our python version (this requires Linux to test) See <https://github.com/openwpm/OpenWPM/issues/1111>
+2. Verify no linter drift warnings were printed by `update.py`. If any `WARNING` lines mention a linter that could not be synced, resolve the mismatch manually.
 3. Run `./scripts/install-firefox.sh` to install the updated Firefox locally and verify the test suite passes.
```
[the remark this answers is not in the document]
vringar: Good catch — step 2 was a leftover from when the parser used `WARNING:` for non-exact pins. Since `sync_precommit_linter_versions()` now raises `RuntimeError` on any drift, there's nothing for the release engineer to verify post-hoc; if `update.py` exits successfully, the sync is correct. Removed step 2 and renumbered (folded into the refactor commit so the doc + behavior change land together).

PR #1172, opened 2026-05-13
fix the aborting issue of js instrument

### LINKED ISSUE #1171: [BUG FIX] the extension doesn't instrument all the APIs in js_instrumentaion_collections
OpenWPM doesn't instrument all the APIs mentioned in my json file in json_instrumentation_collections. Although I wrote the JSON file based on the documentation [here](https://github.com/openwpm/OpenWPM/blob/master/docs/schemas/js_instrument_settings-settings-objects-properties-log-settings.md) and used the default settings. Also the [fingerprinting.json](https://github.com/openwpm/OpenWPM/blob/master/openwpm/js_instrumentation_collections/fingerprinting.json) file was my other reference for writing my own APIs. **My JSON file is attached to this message.** 
This is how I encountered the bug:
  1. installed openwpm in an Ubuntu server
  2. added my own instrumentation JSON file
  3. edited [demo.py](https://github.com/openwpm/OpenWPM/blob/master/demo.py) file to use xvfb mode and assigned True to `browser_param.js_instrument` 
  4. used top 50 Tranco websitEs for crawling
  5. run demo.py
  6. checked the javascript.csv table, the `symbol` column

There are hundreds of APIs in my json file but the crawler didn't instrument most of them, only a few.  For instance, it didn't instrument `Navigator.userAgent` although it's the kind of API that is called by almost every website. 

**My fix**
The code in [js_instument.ts](https://github.com/openwpm/OpenWPM/blob/master/Extension/src/lib/js-instruments.ts) file caused the problem. Some browser-exposed objects or properties throw exceptions during either:

`eval(item.object)` or  `Object.getPropertyNames(object)`

Since these exceptions are uncaught in the original implementation, the entire instrumentation process aborts when a single target fails. As a result, many remaining APIs from the instrumentation collection are never processed.

Original code:
```ts
propertiesToInstrument = Object.getPropertyNames(object);
```

My fix:
```ts
      try {
        propertiesToInstrument = Object.getPropertyNames(object);
      } catch (e) {
        console.warn(
          `OpenWPM: Cannot enumerate properties of ${instrumentedName}:`,
          e,
        );
        propertiesToInstrument = [];
      }
```

Original code:
```ts
instrumentObject(
        eval(item.object),
        item.instrumentedName,
        item.logSettings,
      );
```

My fix:
```ts
      let targetObject;
      try {
        targetObject = eval(item.object);
      } catch (e) {
        console.warn("OpenWPM: eval failed for instrument target:", item.object, e);
        targetObject = undefined;
      }
      instrumentObject(targetObject, item.instrumentedName, item.logSettings);
```
OpenWPM version: 0.31.0
[fingerprinting_keywords2.json](https://github.com/user-attachments/files/27719855/fingerprinting_keywords2.json)

RozhanMK:
Fixes #1171

This PR prevents the JS instrumentation process from aborting when:

- Object.getPropertyNames() throws during property enumeration
- eval(item.object) fails for an instrumentation target

The change adds defensive try/catch handling so instrumentation can continue for remaining APIs even if individual targets fail.

### DIFF
--- a/Extension/src/lib/js-instruments.ts
+++ b/Extension/src/lib/js-instruments.ts
@@ -668,7 +668,15 @@ export function getInstrumentJS(eventId: string, sendMessagesToLogger) {
     if (logSettings.propertiesToInstrument === null) {
       propertiesToInstrument = [];
     } else if (logSettings.propertiesToInstrument.length === 0) {
-      propertiesToInstrument = Object.getPropertyNames(object);
+      try {
+        propertiesToInstrument = Object.getPropertyNames(object);
+      } catch (e) {
+        console.warn(
+          `OpenWPM: Cannot enumerate properties of ${instrumentedName}:`,
+          e,
+        );
+        propertiesToInstrument = [];
+      }
     } else {
       propertiesToInstrument = logSettings.propertiesToInstrument;
     }
@@ -757,11 +765,14 @@ export function getInstrumentJS(eventId: string, sendMessagesToLogger) {
     // More details about how this function is invoked are in
     // content/javascript-instrument-content-scope.ts
     JSInstrumentRequests.forEach(function (item) {
-      instrumentObject(
-        eval(item.object),
-        item.instrumentedName,
-        item.logSettings,
-      );
+      let targetObject;
+      try {
+        targetObject = eval(item.object);
+      } catch (e) {
+        console.warn("OpenWPM: eval failed for instrument target:", item.object, e);
+        targetObject = undefined;
+      }
+      instrumentObject(targetObject, item.instrumentedName, item.logSettings);
     });
   }
 


vringar:
Hey,
I'm unsure if this is what we want here or whether we want to try to fully crash the browser on invalid APIs in the observation specification. 
Obviously silently swallowing this and moving on it very bad, however you'd also only see this warning when looking at the console in the website you are crawling correct? So a typo in one of the APIs you care about would make it seem like that API was never called.
Do you agree? Is there a way a particular website could delete a WebAPI before we run and prevent us from making progress? Afaik we try to run before anything else, so we capture everything.
I haven't used the tool for years, so I welcome your user input

RozhanMK:
Hey, I was working on building a fingerprinting dataset, so I needed OpenWPM to capture all APIs listed in my  JSON file in order to gather a complete dataset.

The main issue for me was that the failure was silent. The crawler continued running normally and the browser never crashed, so everything appeared to work correctly. However, the instrumentation in `javascript.csv` was clearly incomplete. I spent nearly two weeks trying to identify the root cause because there was no explicit sign that the instrumentation process had partially aborted.

Regarding websites deleting APIs before instrumentation: in my experiments, the failures seemed more related to  instrumentation targets or exceptions during property enumeration rather than site interference. After applying the fix, I reran the crawler on the same set of websites (I had a list of around 1000 sites), and the instrumentation results became way more complete. 

I do agree that completely swallowing these errors would be problematic, especially if the instrumentation file contains typos.

One possible compromise could be:

* continue instrumentation for the remaining APIs,
* but also report failed instrumentation targets more explicitly (for example through logging, or a instrumentation summary).

vringar:
Hey,
in #1173 I decided to go for the explicit failure approach. Please lmk if this 1. catches it for you and 2. if that is useful to you

RozhanMK:
Hey,
Even just marking those visits as failed in `crawl_history` is a big help. Thank you.

Issue #31, opened 2015-11-07
Exception in Proxy should notify browser manager


englehardt:
Currently the proxy simply shuts itself down, causing the following javascript command to time out. This means that there is likely proxy data missing for the site that causes the exception, and the following site needlessly times out.

dreisman:
The proxy now talks to the TaskManager through the status_queue, to report when it has hit an exception.

Issue #16, opened 2015-06-25
Firefox binary path hard coded


randomwalker:
The path to the binary is 

./firefox/firefox

(in deploy_firefox.py:170)

Is the assumption that the user has a custom firefox install inside the OpenWPM directory? We don't seem to mention this in the tutorial. The demo fails out of the box for this reason (probably what's causing the error in #15 ).

englehardt:
This should be created by `install.sh`. We now use a specific version of Firefox and selenium that I've found to work well together. I've updated the wiki instructions. Let me know if you still have issues after running that script.

Issue #15, opened 2015-06-23
Where is run_simple_crawl.py?


Antilumo:

englehardt:
`run_simple_crawl.py` was replaced by the much simpler `demo.py` a few releases ago. I've updated the documentation to reflect that.

Antilumo:
Thank you for responding and pointing this out. I have installed OpenWPM using the manual installation instructions given on your "Setting up OpenWPM" page in your wiki. I followed all instructions line by line and installed it on a clean install of ubuntu 14.04. When I run the file demo.py, I get this every time:

deploy_mitm_proxy    - INFO     - BROWSER 4: Intercepting Proxy listening on 36360
Process Process-3:
Traceback (most recent call last):
  File "/usr/lib/python2.7/multiprocessing/process.py", line 258, in _bootstrap
    self.run()
  File "/usr/lib/python2.7/multiprocessing/process.py", line 114, in run
    self._target(*self._args, **self._kwargs)
  File "/home/lm/Desktop/OpenWPM/automation/BrowserManager.py", line 191, in BrowserManager
    (driver, prof_folder, browser_settings) = deploy_browser.deploy_browser(status_queue, browser_params, crash_recovery)
  File "/home/lm/Desktop/OpenWPM/automation/DeployBrowsers/deploy_browser.py", line 9, in deploy_browser
    return deploy_firefox.deploy_firefox(status_queue, browser_params, crash_recovery)
  File "/home/lm/Desktop/OpenWPM/automation/DeployBrowsers/deploy_firefox.py", line 172, in deploy_firefox
    driver = webdriver.Firefox(firefox_profile=fp, firefox_binary=fb)
  File "/usr/local/lib/python2.7/dist-packages/selenium/webdriver/firefox/webdriver.py", line 64, in __init__
    self.binary, timeout),
  File "/usr/local/lib/python2.7/dist-packages/selenium/webdriver/firefox/extension_connection.py", line 51, in __init__
    self.binary.launch_browser(self.profile)
  File "/usr/local/lib/python2.7/dist-packages/selenium/webdriver/firefox/firefox_binary.py", line 69, in launch_browser
    self._start_from_profile_path(self.profile.path)
  File "/usr/local/lib/python2.7/dist-packages/selenium/webdriver/firefox/firefox_binary.py", line 92, in _start_from_profile_path
    env=self._firefox_env)
  File "/usr/lib/python2.7/subprocess.py", line 710, in __init__
    errread, errwrite)
[...]

Issue #1163, opened 2026-04-08
Flaky test: test_browser_profile_coverage finds unexpected baidu.com domain


vringar:
## Summary

`test_browser_profile_coverage` intermittently fails when Firefox contacts `baidu.com` during the test. The test asserts no unexpected domains appear in the browser profile, but `baidu.com` shows up on some CI runs.

## Example failure

```
FAILED test/test_crawl.py::test_browser_profile_coverage
AssertionError: assert {'baidu.com'} == set()
```

## Impact

- Intermittently fails across ALL open PRs (tests group 1)
- Passes on master scheduled runs
- Not caused by any code change — appears and disappears randomly

## Investigation areas

- Firefox default search/suggestion behavior: is Firefox contacting Baidu as part of search engine initialization?
- Firefox preferences: are search-related prefs (`browser.search.*`, `keyword.enabled`) properly disabled?
- Network requests during startup: does Firefox make connections to search partners before instrumentation starts?
- DNS prefetching: is `network.dns.disablePrefetch` set to true?
- GeoIP behavior: does the CI runner's geographic location affect which search partners Firefox contacts?

vringar:
## Revised approach

Instead of removing live sites or marking `xfail`, restructure the test to leverage DNS instrumentation (merged in #1158, extended by #1159):

For each site in `TEST_SITES`:
1. Assert a `dns_responses` row exists (proves the browser attempted resolution)
2. If `addresses IS NOT NULL` (DNS succeeded) → assert `http_requests` and `places.sqlite` entries exist
3. If DNS failed (`error IS NOT NULL`) → skip HTTP/places assertions for that domain

This makes the test **deterministic regardless of network conditions**. `baidu.com` timing out becomes a valid test path, not a failure.

Depends on #1159 (`onErrorOccurred` for DNS failure capture).

PR #1182, opened 2026-06-14
ci: auto-retry known-flaky tests via pytest-rerunfailures

vringar:
Closes crosslink #31.

## Problem

`pytest-split` shards mix flaky browser tests with reliable ones, so a single transient flake reds an entire test group and blocks otherwise-green PRs, forcing manual CI re-runs.

## Change

Decorate the known-flaky tests with `@pytest.mark.flaky(reruns=4, reruns_delay=5)` (via `pytest-rerunfailures`) so transient flakes self-heal in place rather than needing a human to re-trigger the job. A test that fails all 5 runs (1 + 4 reruns) is a real failure and still blocks the PR.

- Add `pytest-rerunfailures` to `environment.yaml` (pinned) and `scripts/environment-unpinned-dev.yaml` (unpinned), matching the other `pytest-*` test deps.
- Register the `flaky` marker in `pyproject.toml` (required because `addopts` uses `--strict-markers`).
- Decorate the three known flakies:
  - `test_cache_hits_recorded` (crosslink #15)
  - `test_browser_profile_coverage` (crosslink #16)
  - `test_audio_fingerprinting`

## Verification

- A throwaway test that fails on first run and passes on retry showed `R.` → `1 passed, 1 rerun`, confirming the rerun mechanism works (test then removed).
- `pytest --markers` lists the registered `flaky` marker with no `--strict-markers` error.
- All three decorated tests still collect cleanly.

### DIFF
--- a/environment.yaml
+++ b/environment.yaml
@@ -23,6 +23,7 @@ dependencies:
 - pyarrow=24.0.0
 - pytest-asyncio=1.3.0
 - pytest-cov=7.1.0
+- pytest-rerunfailures=16.3
 - pytest=9.0.3
 - python=3.14.4
 - pyvirtualdisplay=3.0
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,6 +6,7 @@ addopts = "--strict-markers"
 markers = [
     "pyonly: marks a test as being python only and so server and xpi not needed",
     "slow: marks a test as slow (omit slow tests with '-m \"not slow\"')",
+    "flaky: known-flaky test auto-retried via pytest-rerunfailures; cite its tracking issue",
 ]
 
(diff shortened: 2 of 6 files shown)


vringar:
Closing in favor of root-cause fixes rather than retry-quarantine. An adversarial review showed @flaky(reruns) would mask issues here. Replacement plan: test_cache_hits_recorded -> deterministic restructure with explicitly-encoded tolerated ranges (#1162); test_browser_profile_coverage -> DNS-aided deterministic fix (#1163); test_audio_fingerprinting's marker was a no-op (the test is skipif-skipped on CI). The finalize/visit_id reliability is separately handled by the extension-finalize-ack work.

PR #1193, opened 2026-06-16
fix(js-instrument): don't abort instrumentation when a target fails to resolve (#1171)

### LINKED ISSUE #1171: [BUG FIX] the extension doesn't instrument all the APIs in js_instrumentaion_collections

vringar:
## Summary

Fixes #1171.

The legacy JS instrument iterates over all configured instrumentation requests and calls `eval(item.object)` for each one. When a configured object does not exist in the current scope — for example a Worker-only global such as `Client` or `DedicatedWorkerGlobalScope` configured for a window scope — `eval` throws. Because the iteration used a single `forEach` with no per-request error handling, that exception aborted the **entire** loop, so every request after the failing one (including very common targets like `Navigator`) was silently never instrumented.

This is why, as the reporter observed, configuring a large collection (e.g. a fingerprinting collection containing Worker-only globals) resulted in only a handful of APIs actually appearing in the `javascript` table, even though the configured objects were valid per the schema.

## Fix

In `Extension/src/lib/js-instruments.ts`, resolve each request's target object in its own `try`/`catch` and skip only the failing request (logging a warning), and likewise guard the `instrumentObject` call. A single unresolvable or failing target can no longer prevent the rest of the collection from being instrumented.

## Test

Adds `TestJSInstrumentRequestAfterFailingTarget` plus a test page that passes an instrumentation list whose first request's `object` string fails to evaluate, followed by a valid object. The test asserts the valid object is still instrumented.

- Fails on `master` (valid object never instrumented — loop aborted).
- Passes with this change.

Verified locally: FAIL-before / PASS-after, existing JS instrument tests still pass, extension lint/prettier clean, and Python pre-commit (isort/black/mypy) clean.

### DIFF
--- a/Extension/src/lib/js-instruments.ts
+++ b/Extension/src/lib/js-instruments.ts
@@ -757,11 +757,28 @@ export function getInstrumentJS(eventId: string, sendMessagesToLogger) {
     // More details about how this function is invoked are in
     // content/javascript-instrument-content-scope.ts
     JSInstrumentRequests.forEach(function (item) {
-      instrumentObject(
-        eval(item.object),
-        item.instrumentedName,
-        item.logSettings,
-      );
+      // Resolve the target object. Some configured objects don't exist in
+      // every scope (e.g. Worker-only globals on a window), in which case
+      // eval throws. Catch per-request so one bad target doesn't abort
+      // instrumentation of all subsequent requests (issue #1171).
+      let targetObject;
+      try {
+        targetObject = eval(item.object);
+      } catch (error) {
+        console.warn(
+          `OpenWPM: Could not resolve instrument target '${item.object}', skipping:`,
+          error,
+        );
+        return;
+      }
+      try {
+        instrumentObject(targetObject, item.instrumentedName, item.logSettings);
+      } catch (error) {
+        logErrorToConsole(error, {
+          object: item.object,
+          instrumentedName: item.instrumentedName,
+        });
+      }
     });
   }
 
--- a/test/test_js_instrument.py
+++ b/test/test_js_instrument.py
@@ -394,3 +394,32 @@ class TestJSInstrumentRecursiveProperties(OpenWPMJSTest):
             expected_method_calls=self.METHOD_CALLS,
             expected_gets_and_sets=self.GETS_AND_SETS,
         )
+
+
+class TestJSInstrumentRequestAfterFailingTarget(OpenWPMJSTest):
+    # Regression test for issue #1171: a request whose `object` string fails
+    # to evaluate (e.g. a Worker-only global configured for a window scope)
+    # must not abort instrumentation of subsequent requests. Here the valid
+    # `window.test` object follows a failing target and must still be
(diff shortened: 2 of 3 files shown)


vringar:
Closing as a duplicate of #1173 (and #1172), which I overlooked when this was opened. #1173 takes the preferred approach — propagating the instrumentation failure so the browser never becomes healthy and never measures (fatal-on-failure), rather than skip-and-continue. The regression test from this PR has been lifted into #1173 and adjusted to assert that behavior. Apologies for the noise.

Issue #46, opened 2015-12-23
Consider switching tests to localtest.me


englehardt:
Gunes points out that hosting some tests on `localhost` may cause issues since `localhost` is handled differently than most origins. `localtest.me` always points to 127.0.0.1, so it may be any easy way to test locally with normal security policies.

Another option is to host the tests on an OpenWPM github page, but this fragments the code (since the test pages live on a separate branch), and makes it more difficult to test the tests before pushing them public.

englehardt:
Closed in c9e1ed8ab0a80964af63082c8341cc17af38227d

PR #1203, opened 2026-06-20
feat(storage): data-failure policy — fail visits on constraint violations (crosslink #46)

vringar:
## Data-failure policy (crosslink #46)

Implements the data-failure policy in the core `StorageController` / provider
path. This is the keystone that resolves the adversarial **G1 / G1b / G2**
robustness gaps and aligns the in-flight storage work.

> **Stacks on #1197** (`fix/storage-forward-progress`, the data-integrity
> foundation) — this PR targets that branch, not `master`. Review/merge #1197
> first. The forward-progress fix handles the *transient* half of the policy
> (retry a write/flush blip); this PR adds the *data-fault* half.

### The policy

A failed `store_record` / finalize is classified into two failure classes:

| Class | What it is | Handling |
|---|---|---|
| **Constraint violation** (data fault) | A record tripped the storage schema: DB `IntegrityError`, type mismatch, NOT-NULL / PK / UNIQUE, unknown table/column, parquet `ArrowInvalid` | **Catch, log a clear investigable error** (visit_id, table, which constraint, record key context — PII/size-aware), **mark the VISIT FAILED** (recorded in `incomplete_visits`). Never silently dropped, never retried (a malformed record can't succeed on retry). The site surfaces by repeatedly failing with a good error message. |
| **Transient error** (infra fault) | A write/flush blip (e.g. an S3 `PutObject` failure) | **Retried**, per the forward-progress fix already on the base branch. |

### What changed

- **`openwpm/errors.py`** — new `ConstraintViolation(message, *, table, visit_id, reason)`: the typed data-fault signal carrying enough context to investigate.
- **`openwpm/storage/sql_provider.py`** — `store_record` now *raises* `ConstraintViolation` on schema-tripping errors (`IntegrityError` / `ProgrammingError` / `InterfaceError`, and the schema-mismatch subset of `OperationalError` — "no such table/column"), instead of logging "Unsupported record" and silently dropping. Genuinely transient `OperationalError`s ("database is locked", "disk I/O error") are re-raised for the retry path.
- **`openwpm/storage/arrow_storage.py`** — `_create_batch` raises `ConstraintViolation` on `pa.lib.ArrowInvalid` instead of silently dropping the batch.
- **`openwpm/storage/storage_controller.py`**:
  - `_guarded_store_record` isolates every per-record store at the task boundary: a `ConstraintViolation` (or any unexpected exception) is logged with a clear, investigable message and recorded in `failed_visits`; it **never re-raises** into the connection handler or the finalize path.
  - `finalize_visit_id` forces the visit's effective success to `False` if any of its records tripped the schema (even if the browser sent `success=True`), and handles a `ConstraintViolation` surfaced lazily by the Arrow provider at finalize time — re-finalizing the visit as interrupted so it still reaches a FAILED terminal state.
  - A permanent flush failure on shutdown **surfaces the real error** (not a masked "timeout") and still drains every pending visit as FAILED so shutdown neither hangs nor strands a visit.

### How the invariants now hold

The four non-negotiables (and the adversarial invariants) hold:

1. **Never strand** — every started visit reaches a terminal state. A raising store no longer aborts `finalize_visit_id`, so the visit is always enqueued (as failed). *(G1: `test_raising_store_record_visit_still_finalizes`, `..._unfinalized_visit_enqueued_on_shutdown`)*
2. **Never hang** — a callback `CommandSequence` blocked on a failed record always gets a terminal result. A permanent write fault no longer leaves a finalize token unresolved and blocking shutdown. *(G2: `test_permanent_write_table_fault_visit_still_completes`)*
3. **Never break the shared `DataSocket`** — the per-record failure is isolated so the connection (and the rest of the crawl) survives one bad record. *(G1b: `test_raising_store_record_does_not_break_shared_connection`)*
4. **Never silently drop** — a schema-tripping record fails the visit and is recorded in `incomplete_visits`. *(`test_constraint_violation_fails_visit_and_records_incomplete`)*

### Tests

Brings the relevant adversarial scenarios in as **passing** tests (they were `xfail(strict=True)` in the adversarial suite — under this policy they now pass):

- `test/storage/test_adversarial_storage_controller.py` — G1, G1b, G2, a dedicated constraint-violation test, plus the surviving S4/S6 scenarios.
- `test/storage/test_adversarial_socket.py` — wire-protocol hostility (S3).
- `test/storage/conftest.py` — `adversarial_mp_logger` fixture (errors are expected).

The existing forward-progress liveness tests (`test_storage_controller_liveness.py`) stay green, as do `test_storage_controller.py`, `test_storage_providers.py`, and `test_arrow_cache.py`. `mypy` and `pre-commit` pass.

### Cross-PR coordination

- **#1198** (`test/adversarial-robustness`): the four `xfail(strict=True)` G1/G1b/G2 tests **flip to PASS** under this policy. That suite should rebase onto / be superseded by this PR — its xfail markers must be removed (they would XPASS-fail otherwise).
- **#1161** (SQLAlchemy/Postgres provider): its `store_record` needs the **same** constraint→fail-visit handling (raise `ConstraintViolation` on DB integrity/schema errors). Its provider isn't on this base, so that's a **follow-up on its own branch**.
- **#1174**: the error-surfacing concern (don't mask a real finalize error as a timeout) **aligns** — the shutdown drain here surfaces the genuine failure instead of a timeout.

### Open sub-decision for a maintainer

On the **Arrow** path, a constraint violation that hits one of several tables in a visit keeps the already-validated batches from the other tables (partial-keep) and still records the visit as incomplete. This is the non-destructive default (don't roll back valid data). If a failed visit should instead be **all-or-nothing** (roll back every batch for that visit), that's a deliberate change — flagging it rather than guessing.

### DIFF
--- a/openwpm/errors.py
+++ b/openwpm/errors.py
@@ -40,3 +40,45 @@ class BrowserCrashError(Exception):
     def __init__(self, message, *args):
         self.message = message
         super(BrowserCrashError, self).__init__(message, *args)
+
+
+class ConstraintViolation(Exception):
+    """A record could not be stored because it tripped the storage schema.
+
+    This is a *data* fault, not an infrastructure fault: a NOT-NULL / primary
+    key / unique / type / unknown-column / schema mismatch on a specific
+    record. Per the data-failure policy (crosslink #46) a constraint violation
+    means a page found a way to trip our schema and the affected visit must be
+    failed and surfaced for manual investigation -- NOT silently dropped and
+    NOT retried (retrying a malformed record can never succeed).
+
+    Storage providers raise this from ``store_record`` (or while turning a
+    record into a batch) so the ``StorageController`` can fail the owning visit
+    with an investigable error. Distinguish from a *transient* storage error (a
+    write/flush blip), which is any other exception and is retried.
+
+    Attributes
+    ----------
+    table
+        The table whose schema was tripped.
+    visit_id
+        The visit that owns the offending record.
+    reason
+        A short, human-readable description of which constraint was tripped
+        (e.g. the underlying DB error). Kept concise and PII/size-aware -- it is
+        logged and must be safe to surface.
+    """
+
+    def __init__(
+        self,
+        message: str,
+        *,
+        table: str,
+        visit_id: int,
+        reason: str,
+    ) -> None:
+        self.message = message
+        self.table = table
+        self.visit_id = visit_id
+        self.reason = reason
+        super().__init__(message)
--- a/openwpm/storage/arrow_storage.py
+++ b/openwpm/storage/arrow_storage.py
@@ -10,6 +10,7 @@ import pandas as pd
 import pyarrow as pa
 from pyarrow import Table
 
+from openwpm.errors import ConstraintViolation
 from openwpm.types import VisitId
 
 from .parquet_schema import PQ_SCHEMAS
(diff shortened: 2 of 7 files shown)


vringar:
Sub-decision resolved (maintainer): **partial-keep confirmed** for the Arrow path. When one table's record trips the schema mid-visit, keep the already-validated batches from the other tables and mark the visit incomplete — no all-or-nothing rollback.

Rationale: the first postprocessing step routinely drops every record associated with a failed visit_id, so the partial data never pollutes analysis — but it stays useful for debugging *why* the visit failed.

Issue #53, opened 2016-02-07
Some SSL connections fail when MITMProxy is enabled


englehardt:
Specifically the error is: `ssl_error_rx_record_too_long`.

This seems to be an issues for other users as well, see: [this issue](https://github.com/mitmproxy/mitmproxy/issues/427) and [this issue](https://github.com/mitmproxy/mitmproxy/issues/264) on the mitmproxy bug tracker.

Since we're still on version 0.13 (and using a MITM cert even older than that) upgrading and refreshing the cert store (#36) will hopefully fix this.

englehardt:
This is part of the reason we are removing the proxy. Deprecated in: 5d86590149a6eae9d805076ab0543c959dac7df2.

No use investigating a fix.

PR #1207, opened 2026-06-20
fix(extension): strip moz-extension frames from legacy instrument error stacks

vringar:
## Problem

The legacy JS instrument runs in the **page world** (its instrument body is
stringified into a page `<script>` via `${getInstrumentJS}`). Until now it
re-threw exceptions raised by instrumented APIs **bare**, with no sanitization
(`Extension/src/lib/js-instruments.ts`, `instrumentFunction`'s
`return func.apply(this, arguments)` and the getter/setter call sites).

Because the rethrow propagates the original error back out through the
instrument's own wrapper frames, any page that does
`try { instrumentedApi() } catch (e) { e.stack }` could read the extension's
frames off `e.stack` — `moz-extension://<uuid>/...` — and trivially detect the
instrument. This is a classic anti-measurement detection vector.

## Fix

Back-port the stealth instrument's stack-cleaning predicate to legacy, adapted
for the page world:

- Add two small **inline** page-world helpers inside `getInstrumentJS` (no ES6
  imports / no webpack runtime refs, so they survive the stringification into
  the page `<script>`):
  - `cleanErrorStack(stack)` — drops every stack line containing the literal
    `moz-extension://` scheme (same predicate as
    `Extension/src/stealth/error.ts`).
  - `rethrowWithCleanStack(err)` — overwrites the caught error's own `.stack`
    with the cleaned value and re-throws the **same** error object.
- Wrap the three page-observable call sites (`instrumentFunction`'s
  `func.apply`, and the instrumented getter's / setter's `original*.call`) in
  `try/catch` that re-throws via `rethrowWithCleanStack`.

Unlike the stealth instrument (which runs in the isolated content world behind
an Xray wrapper and must reconstruct the error via `wrappedJSObject`), legacy
already runs in the page world. The caught error is therefore a page-world
object built by the page's own constructors — so we simply overwrite its own
`.stack` (which shadows the prototype accessor) and re-throw it, preserving the
error's original type and identity. Thrown primitives (no `.stack`) are
re-thrown unchanged.

## Test

`test/test_js_instrument_error_stack.py` instruments `window.atob`, makes it
throw from page code (`atob("…invalid base64…")` → `InvalidCharacterError`),
catches the error in the page, records its `.stack`, and asserts the
page-observable stack contains **no** `moz-extension://` frame. Passes on
Firefox 152.

Closes crosslink #53 (review item R16).

### DIFF
--- a/test/test_js_instrument_error_stack.py
+++ b/test/test_js_instrument_error_stack.py
@@ -1,0 +1,101 @@
+"""Invariant guard for crosslink #53 / review item R16.
+
+The legacy JS instrument is injected as an inline ``<script>`` element (see
+``Extension/src/content/javascript-instrument-content-scope.ts``), so Firefox
+attributes ALL of its frames -- including the wrapper closures -- to the page
+document URL, never to ``moz-extension://``. Consequently a page that catches
+an error thrown by an instrumented API sees only page-URL frames on
+``error.stack``; no ``moz-extension://`` frame appears on this path.
+
+This test asserts that end-to-end invariant. It is NOT a reproduction of a
+fixed leak: with the current page-world injection the assertion holds by
+construction -- there is nothing in ``js-instruments.ts`` that strips frames,
+because none reach page code to strip. It DOES fail -- and earns its keep -- if
+the injection model ever regresses to an extension-world script (or a
+``//# sourceURL=moz-extension://...`` directive) that leaks extension frames.
+Should that happen, this test flags it so the leak can be addressed.
+
+Note: the genuine, inherent legacy tell -- the page-URL wrapper frame itself --
+is not covered here and cannot be stripped without breaking script attribution.
+"""
+
+from pathlib import Path
+
+from selenium.webdriver import Firefox
+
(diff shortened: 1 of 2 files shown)


vringar:
**Local verification status (draft):**

- `npm ci && npm run build` — green xpi; the new `cleanErrorStack` / `rethrowWithCleanStack` helpers are confirmed present in the compiled `content.js` (so they survive stringification into the page `<script>`).
- `npm run lint` — green.
- Python pre-commit (isort/black/mypy) — green.

The browser regression test (`test/test_js_instrument_error_stack.py`) could **not** be run to completion locally: in this environment the OpenWPM browser-spawn path fails for *all* browser tests, including unmodified pre-existing ones (e.g. `TestJSInstrumentByPython::test_instrument_object` fails identically with a spawn failure before profile creation). A direct headless Firefox 152 + geckodriver launch works, so the binaries are fine — the failure is in OpenWPM's fork-based multiprocess spawn under Python 3.14.6 (`os.fork()` in a multi-threaded process), unrelated to this change. Leaving as draft so CI validates the browser test.

PR #1236, opened 2026-09-10
build(mypy): type-check against the real environment

vringar:
## The symptom

`mypy openwpm` died in the conda env on numpy's own `__init__.pyi` — *"Type statement is only supported in Python 3.12 and greater"* — before checking a single OpenWPM file, while CI stayed green. Both halves of that were bugs.

## `python_version` was stale

`[tool.mypy] python_version` said `3.10` while `environment.yaml` pins Python 3.14.7.

mypy applies that setting to the third-party stubs it follows, not just to our own code, so a stale value breaks on any dependency using newer syntax. It also meant the project was being type-checked at an older language level than it actually ships.

Set to `3.14`, and `scripts/update.py` gains `sync_mypy_python_version` so a repin can't leave it behind again — alongside the existing pre-commit-rev and `engines.node` syncs. It rewrites only the `[tool.mypy]` table and fails loud if the section or the setting is missing, matching the convention of the other sync helpers.

## CI missed it because the hook didn't run in this environment

`mirrors-mypy` installs mypy into its own venv, so it saw only the four packages named in `additional_dependencies` and never resolved numpy at all. That list was a second, hand-maintained copy of the dependency set — and the config comment already conceded the treadmill (*"We may need to add more and more dependencies here"*). Everything not on it degraded to `Any`, so the hook was passing largely by not looking.

mypy now runs as a `language: system` hook. The `pre-commit` CI job already sets up the conda env, so no workflow change was needed, and it now type-checks against the packages the project actually installs and their bundled type information.

- **The version is unchanged**: `environment.yaml` pins `mypy=2.3.1`, the same release the removed `rev: v2.3.1` pointed at. It's now pinned in one place instead of two.
- That's also why mypy comes out of `update.py`'s `_LINTER_MAP`: a system hook has no rev to sync, and leaving it there would make the next repin raise. A regression test pins that.
- `--scripts-are-modules` is carried over from the mirrors-mypy hook definition and is load-bearing: without it a file with a shebang is named `__main__`, which the `module = "test.*"` overrides would not match. `test/test_http_instrumentation.py` has one.
- `--ignore-missing-imports`, also inherited, is dropped because `pyproject.toml` already sets it.

## Verification

No new errors surface: mypy in the conda env over all tracked Python files is clean. `test/test_update_script.py` grows from 9 to 15 tests, covering the new sync helper (update, no-op-without-rewriting, missing python, missing section, missing setting) plus a regression test asserting mypy stays out of `_LINTER_MAP`.

Split out of #1230, where it was unrelated noise.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,14 +8,29 @@ repos:
     hooks:
       - id: black
         language_version: python3
-  - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v2.3.1
+  - repo: local
     hooks:
+      # mypy runs from the conda env rather than a pre-commit-managed venv, so
+      # it type-checks against the packages the project actually installs and
+      # their bundled type information. Under mirrors-mypy it saw only a
+      # hand-listed subset of dependencies, so every other third-party import
+      # degraded to `Any` -- and that list was a second copy of the dependency
+      # set, free to drift from environment.yaml. Its version now comes from
+      # environment.yaml like every other tool, and `python_version` in
+      # pyproject.toml is kept in step by scripts/update.py.
+      #
+      # `--scripts-are-modules` is inherited from the mirrors-mypy hook
+      # definition and is load-bearing: without it a file with a shebang is
+      # named `__main__`, which the `module = "test.*"` overrides would not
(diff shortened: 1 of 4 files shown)


vringar (commented):
I feel like update.py should use actual parsers and not just string together regexes

